PR #4, opened 2024-08-15
Refactor maintenance structure, add prompt library

paul-paliychuk:


core/graphiti.py
```diff
@@ -1,47 +1,124 @@
...
+                self.llm_client, episode, new_nodes, relevant_schema, previous_episodes
+            )
+            edges.extend(new_edges)
+            episodic_edges = build_episodic_edges(nodes, episode, datetime.now())
+            edges.extend(episodic_edges)
+
+            # invalidated_edges = await self.invalidate_edges(
+            #     episode, new_nodes, new_edges, relevant_schema, previous_episodes
+            # )
+
```
paul-paliychuk, line 106: @prasmussen15 temp commented out not runnable bits

core/graphiti.py
```diff
@@ -76,18 +153,9 @@ async def build_indices(self):
...
-        """Retrieve the last n episodic nodes from the graph"""
-        ...
-
-    # Utility function, to be removed from this class
-    async def clear_data(self): ...
-
     async def search(
         self, query: str, config
-    ) -> (list)[Tuple[EntityNode, list[EntityEdge]]]:
+    ) -> (list)[tuple[EntityNode, list[EntityEdge]]]:
```
paul-paliychuk, line 158: @prasmussen15 I think we should prefer to use the default python types (so tuple vs Tuple from typing, any vs Any)

PR #180, opened 2024-10-07
Add mmr reranking

prasmussen15:
> [!IMPORTANT]
> Add MMR reranking to search, normalize embeddings, and update community building with group ID support.
> 
>   - **Search Enhancements**:
>     - Add MMR reranking method to `search.py` for edges, nodes, and communities.
>     - Update `search_config.py` and `search_config_recipes.py` to include MMR reranking.
>     - Add `maximal_marginal_relevance()` function to `search_utils.py`.
>   - **Embedding Normalization**:
>     - Use `normalize_l2()` in `openai.py` and `voyage.py` to normalize embeddings.
>   - **Community Building**:
>     - Update `build_communities()` in `graphiti.py` and `community_operations.py` to accept `group_ids`.
>   - **Testing**:
>     - Update `test_graphiti_int.py` to test new search configurations and community building.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=getzep%2Fgraphiti&utm_source=github&utm_medium=referral)<sup> for 6f69a613efee88e3e2fc379933c2958ac88d8cf8. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/graphiti_core/embedder/openai.py
+++ b/graphiti_core/embedder/openai.py
@@ -20,6 +20,7 @@ from openai import AsyncOpenAI
 from openai.types import EmbeddingModel
 
 from .client import EmbedderClient, EmbedderConfig
+from ..helpers import normalize_l2
 
 DEFAULT_EMBEDDING_MODEL = 'text-embedding-3-small'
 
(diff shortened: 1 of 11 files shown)


danielchalef:
@prasmussen15 it looks like we're normalizing vectors. Do we need to run a migration on existing installs? Would it now make sense to use dot product rather then cosine as distance metric?

paul-paliychuk (approved):

prasmussen15:
> @prasmussen15 it looks like we're normalizing vectors. Do we need to run a migration on existing installs? Would it now make sense to use dot product rather then cosine as distance metric?

Neo4j only offers cosine similarity and not dot product so a migration on existing installs isn't necessary for the search. We could run a migration for MMR since I use dot product there.

Issue #331, opened 2025-04-07
running example failed


litaolaile:
2025-04-07 16:25:08 - graphiti_core.llm_client.openai_client - ERROR - Error in generating LLM response: Error code: 400 - {'object': 'error', 'message': "[{'type': 'string_type', 'loc': ('body', 'model'), 'msg': 'Input should be a valid string', 'input': ['QwQ-32B']}, {'type': 'float_type', 'loc': ('body', 'temperature'), 'msg': 'Input should be a valid number', 'input': [0.2]}]", 'type': 'BadRequestError', 'param': None, 'code': 400}

danielchalef:
It looks like you're using a Qwen model with Graphiti and the model is not generating JSON output in the format Graphiti requires. ~Please would you try with a frontier model from OpenAI or Google and let us know if this issue persists.~

Update: I just noticed you're using the OpenAIClient. Please use the `OpenAIGenericClient` when not using the OpenAI API. The `OpenAIGenericClient` adds the expected schema to the prompt, while this isn't needed when using OpenAI Structured Output.

mingfang:
`OpenAIGenericClient` is not exported
https://github.com/getzep/graphiti/blob/main/graphiti_core/llm_client/__init__.py#L6

I'm getting this error
```
ImportError: cannot import name 'OpenAIGenericClient' from 'graphiti_core.llm_client' (/home/coder/graphiti/graphiti_core/llm_client/__init__.py)
```

PR #438, opened 2025-05-04
Simplifying imports for llm_client and cross_encoder

t41372:
**Why?**

* It didn’t make sense to expose only `OpenAIClient` while omitting `OpenAIGenericClient`, `AnthropicClient`, `GeminiClient`, and `GroqClient`. This led to confusion for users.
* A similar issue existed in `cross_encoder`, so I addressed that as well.
* I noticed related confusion mentioned in #331.

**What’s fixed:**

* Added the missing clients to `__init__.py` to ensure they’re properly exposed.

**Note:**

I didn’t open a separate issue to discuss this because I ran into the problem myself while using Graphiti, and it was quicker to submit a PR. Please feel free to close this if there are other considerations I missed.

### DIFF
--- a/graphiti_core/cross_encoder/__init__.py
+++ b/graphiti_core/cross_encoder/__init__.py
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+from .bge_reranker_client import BGERerankerClient
 from .client import CrossEncoderClient
 from .openai_reranker_client import OpenAIRerankerClient
 
-__all__ = ['CrossEncoderClient', 'OpenAIRerankerClient']
+__all__ = ['CrossEncoderClient', 'OpenAIRerankerClient', 'BGERerankerClient']
--- a/graphiti_core/llm_client/__init__.py
+++ b/graphiti_core/llm_client/__init__.py
@@ -1,6 +1,19 @@
+from .anthropic_client import AnthropicClient
 from .client import LLMClient
 from .config import LLMConfig
 from .errors import RateLimitError
+from .gemini_client import GeminiClient
+from .groq_client import GroqClient
 from .openai_client import OpenAIClient
+from .openai_generic_client import OpenAIGenericClient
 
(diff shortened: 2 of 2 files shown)


danielchalef:

t41372:

graphiti_core/cross_encoder/__init__.py
```diff
@@ -14,7 +14,8 @@
 limitations under the License.
 """
 
+from .bge_reranker_client import BGERerankerClient
 from .client import CrossEncoderClient
 from .openai_reranker_client import OpenAIRerankerClient
 
-__all__ = ['CrossEncoderClient', 'OpenAIRerankerClient']
+__all__ = ['CrossEncoderClient', 'OpenAIRerankerClient', 'BGERerankerClient']
```
prasmussen15: It was intentional the we don't expose the BGEReranker client here as it adds additional dependencies. BGERerankerClient is a sample client and should only be used locally for dev testing. In prod the client should be hosted on its own server rather than being downloaded like in the BGEReranker example.

The dependency added is SentenceTransformers, which itself has PyTorch as a dependency. This leads to hundred of MB of additional dependencies that won't ever actually be used in real deployments of graphiti.

prasmussen15 (changes requested):

graphiti_core/cross_encoder/__init__.py
t41372: Thanks for the heads up! That makes total sense. I’ll go ahead and remove it.

t41372: Actually I just noticed that this problem is not unique to `BGERerankerClient`. Same problem also applies to gemini, anthropic, and groq because they are all optional dependencies.

Rather than removing them, I’ve updated the code to conditionally expose each client only if its dependencies are available. Just pushed a commit with this change — let me know if it looks good to you!

danielchalef:
This has been superseded by #675

PR #584, opened 2025-06-15
Feat/driver factory

karn09:
Context and Purpose:

This PR introduces a driver factory to the codebase, enabling automatic selection of the appropriate graph database driver (e.g., Neo4j or FalkorDB) based on the URI scheme provided to the Graphiti class.

Description:
Previously, the backend driver was hardcoded to Neo4j unless a driver instance was manually provided. However, due to import limitations, it was not possible to instantiate and inject a custom driver from outside the project, making it difficult to use alternative backends like FalkorDB. The new factory approach streamlines backend selection and improves extensibility.

Solution Implemented:
* Added a GraphDriverFactory with a static get_driver method that selects and instantiates the correct driver based on the URI scheme.
* Refactored the Graphiti class to use the factory for driver instantiation, removing direct imports of specific drivers.

Impact:
* Users can now specify a neo4j:// or falkor:// URI, and the correct driver will be used automatically.
* Manual driver injection from outside the project is still not possible due to import restrictions, but this change lays the groundwork for future improvements in that area.

### DIFF
--- a/graphiti_core/driver/factory.py
+++ b/graphiti_core/driver/factory.py
@@ -1,0 +1,27 @@
+"""
+Copyright 2024, Zep Software, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


danielchalef:

karn09:

prasmussen15:
Checking URI schemas is not a good way to determine options. Neo4j is the default for backwards compatibility. To use other GraphDBs, one must pass in a `graph_driver` object. This pattern aligns with other options when initializing Graphiti such as the LLM client, embedder, and reranker.

PR #660, opened 2025-07-02
Fix: Robust handling for dynamic labels during bulk ingestion

### LINKED ISSUE #659: Bug: Ingestion fails with Cypher Error when using dynamic labels due to unsafe apoc.create.addLabels call
#### Description

When ingesting data using `graphiti.add_episode()`, the process can fail with a `neo4j.exceptions.CypherError` if an entity contains dynamic `labels`.

The root cause is the Cypher query in `node_db_queries.py`, which unconditionally calls the `apoc.create.addLabels` procedure. This call is not resilient and fails if the `labels` parameter passed from the application is `null`, an empty list, or not a list of strings, which can easily happen with dynamic data.

This makes the ingestion process brittle and requires the calling application to perform extensive data cleaning beforehand. A more robust query should handle these cases gracefully within Cypher.

#### Steps to Reproduce

1.  Set up Graphiti with a Neo4j backend that has the APOC library installed.
2.  Call `graphiti.add_episode()` with data that generates an entity where the `labels` attribute is `None`, an empty list (`[]`), or contains non-string values.
3.  The ingestion query fails, and a `CypherError` is thrown by the Neo4j driver.

#### Expected Behavior

The node should be created or merged successfully. If the `labels` list is valid and contains strings, the labels should be applied. If the `labels` list is null or empty, the query should complete without error and without applying any new labels.

#### Actual Behavior

A `neo4j.exceptions.CypherError` is raised, preventing the data from being ingested. The error message typically relates to a type or syntax error within the `apoc.create.addLabels` procedure call.

#### Proposed Solution

The Cypher query for bulk node ingestion should be modified to make the call to `apoc.create.addLabels` conditional. By wrapping the `CALL` subquery in a `FOREACH` hack, we can ensure it only executes when `node.labels` is a non-empty list.

This makes the query resilient to bad data, removes the burden of pre-processing from the client, and ensures the entire batch transaction does not fail over a single malformed entity.

**Suggested Code Change in `graphiti_core/models/nodes/node_db_queries.py`:**

The current query is conceptually similar to this:

```cypher
# Current, brittle query
UNWIND $nodes AS node
MERGE (n:Entity {uuid: node.uuid})
ON CREATE SET n = node.properties, n.created_at = $timestamp
ON MATCH SET n += node.properties
WITH n, node
// This fails if node.labels is null or empty
CALL apoc.create.addLabels(n, node.labels) YIELD node AS ignored
RETURN count(*)
```

It should be replaced with a more robust version:

```cypher
# Proposed, robust query
UNWIND $nodes AS node
MERGE (n:Entity {uuid: node.uuid})
ON CREATE SET n = node.properties, n.created_at = $timestamp
ON MATCH SET n += node.properties
WITH n, node
// Use FOREACH as a conditional wrapper to safely call the subquery
FOREACH(i IN (CASE WHEN node.labels IS NOT NULL AND size(node.labels) > 0 THEN [1] ELSE [] END) | 
    CALL apoc.create.addLabels(n, node.labels) YIELD node as ignored_node
)
RETURN count(*)
```

This approach is idiomatic, efficient, and makes the Graphiti ingestion process far more robust.

iberi22:
This PR addresses issue #659 by making the bulk ingestion of `EntityNode` more robust.

**Problem:**
The previous implementation could fail if a node's `labels` property was `None` or an empty list, causing a `CypherError` during the query.

**Solution:**
This change uses a conditional `FOREACH` clause in the Cypher query. This ensures that labels are only applied if the `labels` list is not null and not empty, preventing errors during bulk ingestion.

**Validation:**
An integration test (`test_bulk_save_with_problematic_labels`) has been added to verify that nodes with `None` or empty `labels` can be saved successfully using the bulk ingestion mechanism.

Closes #659

----

> [!IMPORTANT]
> Fixes handling of `None` or empty `labels` in `EntityNode` bulk ingestion with a conditional Cypher query and adds an integration test.
> 
>   - **Behavior**:
>     - Updates `ENTITY_NODE_SAVE` and `ENTITY_NODE_SAVE_BULK` in `node_db_queries.py` to use a conditional `FOREACH` clause for applying labels only if `labels` is not `None` or empty.
>   - **Testing**:
>     - Adds `test_bulk_save_with_problematic_labels` in `test_node_labels_bug_int.py` to verify successful saving of nodes with `None` or empty `labels`.
>     - Uses `pytest` and `AsyncMock` to simulate database interactions and validate the solution.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=getzep%2Fgraphiti&utm_source=github&utm_medium=referral)<sup> for c6f2ca94d9f2a24e4005e9b0fa5fd32e1aca45d1. You can [customize](https://app.ellipsis.dev/getzep/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/graphiti_core/models/nodes/node_db_queries.py
+++ b/graphiti_core/models/nodes/node_db_queries.py
@@ -31,7 +31,10 @@ EPISODIC_NODE_SAVE_BULK = """
 
 ENTITY_NODE_SAVE = """
         MERGE (n:Entity {uuid: $entity_data.uuid})
-        SET n:$($labels)
+        WITH n, $labels
+        WITH n, $labels AS labels_list
+        // Workaround: Si APOC/addLabels no está disponible, usar Cypher puro para etiquetas dinámicas
+        FOREACH (label IN CASE WHEN labels_list IS NOT NULL AND size(labels_list) > 0 THEN labels_list ELSE [] END | SET n:label)
         SET n = $entity_data
         WITH n CALL db.create.setNodeVectorProperty(n, "name_embedding", $entity_data.name_embedding)
         RETURN n.uuid AS uuid"""
@@ -39,7 +42,10 @@ ENTITY_NODE_SAVE = """
 ENTITY_NODE_SAVE_BULK = """
     UNWIND $nodes AS node
     MERGE (n:Entity {uuid: node.uuid})
-    SET n:$(node.labels)
+    WITH n, node
+    WITH n, node, node.labels AS labels_list
+    // Workaround: Si APOC/addLabels no está disponible, usar Cypher puro para etiquetas dinámicas
+    FOREACH (label IN CASE WHEN labels_list IS NOT NULL AND size(labels_list) > 0 THEN labels_list ELSE [] END | SET n:label)
     SET n = node
     WITH n, node CALL db.create.setNodeVectorProperty(n, "name_embedding", node.name_embedding)
     RETURN n.uuid AS uuid
--- a/tests/test_node_labels_bug_int.py
+++ b/tests/test_node_labels_bug_int.py
@@ -1,0 +1,99 @@
+"""
+Copyright 2024, Zep Software, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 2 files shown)


danielchalef:
   - - -
   - - -

iberi22:
recheck

PR #716, opened 2025-07-14
Support Multi-Graph Usage with a Single Database Connection

galshubeli:
## Summary
This change improves the way multi-graph setups are handled using a single graph connection with FalkorDB driver.
After initializing a database connection, you can pass it to the Graphiti client and specify different graph names, ensuring a separate graph for each user/agent.

Example:
```
driver = FalkorDriver(...)
graphiti_a = Graphiti(graph_driver=driver.clone('team_a'))
graphiti_b = Graphiti(graph_driver=driver.clone('team_b'))
```

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Performance improvement
- [ ] Documentation/Tests

## Objective
**For new features and performance improvements:** Clearly describe the objective and rationale for this change.

## Testing
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] All existing tests pass

## Breaking Changes
- [ ] This PR contains breaking changes

If this is a breaking change, describe:
- What functionality is affected
- Migration path for existing users

## Checklist
- [ ] Code follows project style guidelines (`make lint` passes)
- [ ] Self-review completed
- [ ] Documentation updated where necessary
- [ ] No secrets or sensitive information committed

## Related Issues
Closes #[issue number]

----

> [!IMPORTANT]
> Adds support for specifying a graph database name in Graphiti initialization and operations, with a new `with_database` method for driver database switching.
> 
>   - **New Feature**:
>     - Adds `graph_name` parameter to `Graphiti.__init__()` in `graphiti.py` to specify the graph database name.
>     - Introduces `with_database()` method in `GraphDriver` in `driver.py` to create a shallow copy of the driver with a new database name.
>   - **Driver Changes**:
>     - `FalkorDriver.__init__()` in `falkordb_driver.py` now sets `_database` during initialization.
>     - `FalkorDriver.execute_query()` in `falkordb_driver.py` uses `_database` for graph selection.
>   - **Behavior**:
>     - `Graphiti` uses `with_database()` to switch databases if `graph_name` is provided during initialization.
>     - Default database name is used if `graph_name` is not specified.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=getzep%2Fgraphiti&utm_source=github&utm_medium=referral)<sup> for ffee14cf00d2e581af90ce1f56dcffb757432c30. You can [customize](https://app.ellipsis.dev/getzep/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #741, opened 2025-07-17
fix missing attribute for self._database

soraxas:
## Summary
Brief description of the changes in this PR.

## Type of Change
- [x] Bug fix

## Objective
A Trivial change. The `self._database` is only set when db is not given (within a conditional branch), but it's used later in other member methods, e.g.

https://github.com/getzep/graphiti/blob/099dc4002532e3cc531064f0131fd228519afc6f/graphiti_core/driver/falkordb_driver.py#L106-L108

...resulting in attribute error

----

> [!IMPORTANT]
> Fixes missing `_database` attribute initialization in `FalkorDriver` to prevent attribute errors.
> 
>   - **Bug Fix**:
>     - Ensure `self._database` is always set in `FalkorDriver.__init__()` in `falkordb_driver.py`, preventing attribute errors in methods like `_get_graph()` and `execute_query()`.
>     - Previously, `self._database` was only set when `falkor_db` was not provided, causing issues when accessed later.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=getzep%2Fgraphiti&utm_source=github&utm_medium=referral)<sup> for 099dc4002532e3cc531064f0131fd228519afc6f. You can [customize](https://app.ellipsis.dev/getzep/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -95,7 +95,7 @@ class FalkorDriver(GraphDriver):
             self.client = falkor_db
         else:
             self.client = FalkorDB(host=host, port=port, username=username, password=password)
-            self._database = database
+        self._database = database
 
     def _get_graph(self, graph_name: str | None) -> FalkorGraph:
         # FalkorDB requires a non-None database name for multi-tenant graphs; the default is "default_db"


danielchalef:

soraxas:

danielchalef:
I think the intention here is that `falkor_db` would have a database set. Your change would potentially overwrite it with the default. 

@galshubeli Would you mind taking a look here and telling us how you'd like to address the bug?

soraxas:
@danielchalef I don't understand what you meant. 

The `_database` missing attribute is set on `graphiti`'s wrapper `FalkorDriver`

The issue is that when an already constructed `falkor_db` is passed into `graphiti`'s wrapper
```py
... = FalkorDriver(falkor_db=...)
```
the object is missing that attribute (which is used later in its own methods).

Are you saying we should respect `falkor_db`'s own database name when its explicitly passed? I'm not sure if it has a default db name, and if it does, then it should be that `graphiti`'s wrapper should be adjusted to somehow use that attribute?

galshubeli:
Thanks @soraxas, @danielchalef this fixes the bug that occurred when passing a FalkorDB connection instead of credentials.

FalkorDB supports using different graphs on the same connection with full isolation.
For example (after the change):
```
from falkordb.asyncio import FalkorDB

# The client holds the connection, client.select_graph(graph_name) chooses the specific graph
falkordb_client = FalkorDB(host=..., port=..., username=..., password=...)
driver_a = FalkorDriver(falkor_db=falkordb_client, database='team_a')
driver_b = FalkorDriver(falkor_db=falkordb_client, database='team_b')
graphiti_a = Graphiti(graph_driver=driver_a)
graphiti_b = Graphiti(graph_driver=driver_b)
```

This change is required and it's already handled in the PR: https://github.com/getzep/graphiti/pull/716

danielchalef:
Duplicate of #716 and #761

PR #968, opened 2025-10-02
fix: Improve edge extraction entity ID validation

danielchalef:
## Summary
Fixes invalid entity ID references in edge extraction that caused warnings like:
```
WARNING: source or target node not filled WILL_FIND. source_node_uuid: 23 and target_node_uuid: 3
```

The LLM was hallucinating entity IDs outside the valid range, causing edges to be skipped during extraction.

## Root Cause
- ENTITIES list was formatted as Python repr instead of proper JSON, making it harder for LLMs to parse
- Field descriptions didn't clearly indicate these are indices from the provided ENTITIES list
- Extraction rules didn't explicitly validate entity ID ranges
- Error messages included PII (fact content)

## Changes
- Format ENTITIES list as proper JSON in prompt using `to_prompt_json()` for better LLM parsing
- Clarify field descriptions to reference entity `id` from ENTITIES list
- Add explicit entity ID validation as #1 extraction rule with concrete examples
- Improve error logging: removed PII, added entity count and valid range

## Testing
- All existing unit tests pass (7/7 in `test_edge_operations.py`)
- Changes follow patterns from `extract_nodes.py` and `dedupe_nodes.py` where entity referencing works reliably

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -783,7 +783,7 @@ wheels = [
 
 [[package]]
 name = "graphiti-core"
-version = "0.21.0rc6"
+version = "0.21.0rc11"
 source = { editable = "." }
 dependencies = [
     { name = "diskcache" },


prasmussen15 (approved):

uv.lock
```diff
@@ -783,7 +783,7 @@ wheels = [
 
 [[package]]
 name = "graphiti-core"
-version = "0.21.0rc11"
+version = "0.21.0rc6"
```
danielchalef:
```suggestion
version = "0.21.0rc11"
```

PR #1053, opened 2025-11-06
feat(openai): add JSON-schema fallback for structured outputs in openai_client

Galleons2029:
## Summary
This PR improves compatibility with OpenAI-compatible providers that don’t yet support the Responses API’s structured parsing. We keep the primary path using `client.responses.parse(...)` for first-class structured outputs and add an automatic fallback to `chat.completions.create(...)` with JSON Schema response_format when the Responses API appears unsupported (404/NotFound/AttributeError).

Motivation
Many third-party “OpenAI-compatible” platforms only implement the classic `chat.completions` endpoint but do support JSON-mode structured outputs. Previously, Graphiti’s structured output only worked via `responses.parse`, limiting compatibility across providers. This PR expands compatibility without impacting the OpenAI-native path.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests
### What changed
- `OpenAIClient._create_structured_completion`:
  - Try Response API (`responses.parse`) first.
  - On clear non-support (404/NotFound/AttributeError or error messages mentioning “responses” / “not found”), fall back to `chat.completions.create` with `response_format: {type: "json_schema"}`.
  - Build JSON Schema from the Pydantic model, supporting both Pydantic v2 (`model_json_schema`) and v1 (`schema`).
  - Normalize provider output to a JSON string and wrap it in a simple response with `.output_text` so downstream code can continue to parse consistently.
  - Preserve reasoning-model temperature behavior (suppress temperature for gpt‑5/o1/o3 families) in both primary and fallback paths.
- Minor import additions (`json`, `openai`).

## Objective
- OpenAI-native structured output remains unchanged and preferred.
- When falling back, the returned object exposes `.output_text`, preserving consumer expectations to parse a JSON string. This should be non-breaking for existing code that expects textual structured output from the Responses API path.

## Testing
- [x] All existing tests pass
- Existing tests pass on my end with both the primary path and the fallback path validated.
- I exercised:
  - Successful `responses.parse` path with a Pydantic response model
  - Simulated “unsupported” provider conditions triggering fallback
  - Reasoning-model behavior to confirm temperature suppression in both paths
- If you’d like a quick local run:
  - For OpenAI-native: set TEST_OPENAI_* env vars and run `make test`.
  - For fallback: point the client at a provider that lacks `/v1/responses` but supports `chat.completions` with `json_schema`.

## Breaking Changes
- Providers exposing only `chat.completions` can now return structured outputs by implementing `response_format = {"type": "json_schema", ...}`.
- The schema name defaults to the Pydantic model class name.
- The client supplies `max_tokens` and suppresses temperature for reasoning-model families.

Implementation notes
- Fallback is intentionally conservative; we re-raise non-404/unknown-endpoint errors to avoid masking real failures.
- The fallback uses a tiny wrapper class that exposes `.output_text` to align with the Responses API’s textual output expectations.

Maintenance
- Import ordering/formatting is left to the project formatter. Run:
  - make format
  - make lint
  - make test

## Checklist
- [x] Self-review completed
- [x] Primary structured parsing path unchanged for OpenAI
- [x] JSON Schema fallback for third-party `chat.completions`
- [x] Reasoning-model temperature suppression applied consistently
- [x] Pydantic v1/v2 schema support
- [x] Defensive JSON normalization and simple response wrapper
- [x] Tests pass locally

## Completion summary

- Implemented the structured-output fallback in `openai_client.py` with careful error checks and provider-compatibility details.
- Preserved reasoning-model temperature handling in both primary and fallback paths.
- Quick lint check flagged import ordering; it’s minor and should be resolved by `make format` to align with your repo’s rules.

Notes: No breaking changes; existing tests pass.

### DIFF
--- a/graphiti_core/llm_client/openai_client.py
+++ b/graphiti_core/llm_client/openai_client.py
@@ -116,7 +116,7 @@ class OpenAIClient(BaseOpenAIClient):
             # Some providers require a schema name; use model class name by default
             schema_name = getattr(response_model, '__name__', 'structured_response')
 
-            print(f'Falling back to chat.completions with JSON schema for model {model}...')
+            # Falling back to chat.completions with JSON schema for model
             completion = await self.client.chat.completions.create(
                 model=model,
                 messages=messages,


danielchalef:

Galleons2029:

danielchalef:
Hi @Galleons2029 , thanks for this contribution. Graphiti already has a generic OpenAI client that uses "json" mode output. Are there any improvements we should make to the existing client?  

https://github.com/getzep/graphiti/blob/674619cc896b9433ae7c01ef1ffebb5d61978ea6/graphiti_core/llm_client/openai_generic_client.py#L107

Galleons2029:
> Hi @Galleons2029 , thanks for this contribution. Graphiti already has a generic OpenAI client that uses "json" mode output. Are there any improvements we should make to the existing client?
> 
> https://github.com/getzep/graphiti/blob/674619cc896b9433ae7c01ef1ffebb5d61978ea6/graphiti_core/llm_client/openai_generic_client.py#L107

Thank you for your reminder, which made me realize that the json schema in the generic OpenAI client might also need to be replaced. Simply adding json_format as the type parameter cannot better enable the model to accurately output json type. Moreover, there is no provided accurate schema for the output json here, meaning that if the format of the model is not accurately specified in the prompt, the model will not know what kind of json to output. 

image


At the same time, I added it to the OpenAI client considering that the current generic OpenAI client does not mention it in the import path. Usually, developers directly use it, but it will result in a 404 error. Adding this patch is also to ensure correct operation by reverting to the json schema. I compared the same LLM that I used with the results of adding and not adding the scheme. Currently, the accuracy rate of the added json schema is 100%, while the generic OpenAI client has frequently reported errors.
image

danielchalef:
Hi @Galleons2029 -- it appears adding the schema is how [OpenAI structured output](https://platform.openai.com/docs/guides/structured-outputs#structured-outputs-vs-json-mode) works in the completion API. I'm not sure how widespread this is supported by other OpenAI-compatible inference services.

We can make this change to the generic client at the risk of breaking these other servers. I'll cut a PR.

danielchalef:
@Galleons2029 would you like to test out this PR? https://github.com/getzep/graphiti/pull/1061

Galleons2029:
> @Galleons2029 would you like to test out this PR? #1061

Sure, after conducting the test, I found that the original json problem had been resolved, but at the same time, a new minor issue emerged：
The LLM response is truncated by the output token limit, as confirmed by finish_reason=length in the logs. The raw tail shows an array of objects cut mid-entry (e.g., after "name": "A10938W095"), so the JSON is incomplete， producing the same error：
image
image


Because the payload is cut off, json.loads fails with “Unterminated string…”. Earlier, the client tried to parse this partial JSON; after our change, it correctly flags truncation before parsing.
I think the cause is that the requested structured output (a large list of entities) exceeds the allotted max_tokens (and may be further inflated by schema/prompt overhead). The provider stops output before closing braces/quotes, producing invalid JSON. And the repeated retries don’t help because the context gets longer while max_tokens stays the same, leading to the same cutoff point and the same failure.

Galleons2029:
> > @Galleons2029 would you like to test out this PR? #1061
> 
> Sure, after conducting the test, I found that the original json problem had been resolved, but at the same time, a new minor issue emerged： The LLM response is truncated by the output token limit, as confirmed by finish_reason=length in the logs. The raw tail shows an array of objects cut mid-entry (e.g., after "name": "A10938W095"), so the JSON is incomplete， producing the same error： image

image

> 
> Because the payload is cut off, json.loads fails with “Unterminated string…”. Earlier, the client tried to parse this partial JSON; after our change, it correctly flags truncation before parsing. I think the cause is that the requested structured output (a large list of entities) exceeds the allotted max_tokens (and may be further inflated by schema/prompt overhead). The provider stops output before closing braces/quotes, producing invalid JSON. And the repeated retries don’t help because the context gets longer while max_tokens stays the same, leading to the same cutoff point and the same failure.

If necessary, I can spend some time on it to solve it, or it can hand over to Claude.

danielchalef:
@Galleons2029 this should be fixed.

PR #1086, opened 2025-11-28
Fix: Prevent oscillation in label propagation algorithm

ZLBillShaw:
## Summary
Fixed infinite loop bug in label propagation algorithm caused by synchronous updates leading to oscillation in certain graph structures.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
The previous implementation of the label propagation algorithm in `community_operations.py` used synchronous updates where all nodes update their community labels simultaneously. This approach can cause oscillation in certain graph topologies (e.g., bipartite graphs, regular graphs), where nodes continuously swap between two or more community assignments without ever converging, resulting in infinite loops.

**Root Cause:**
- Synchronous updates create systematic patterns that can cause nodes to oscillate
- No termination guarantee for non-convergent graphs
- No mechanism to detect or break out of oscillation cycles

**Solution:**
1. **Asynchronous Updates**: Process nodes in randomized order each iteration, breaking systematic oscillation patterns
2. **Maximum Iteration Limit**: Set `MAX_ITERATIONS = 100` to guarantee termination
3. **Oscillation Detection**: Track recent state history and detect when the algorithm enters a cycle
4. **Deterministic Tie-Breaking**: Improved sorting for consistent results across runs

### DIFF
--- a/graphiti_core/utils/maintenance/community_operations.py
+++ b/graphiti_core/utils/maintenance/community_operations.py
@@ -85,99 +85,106 @@ async def get_community_clusters(
 
 def label_propagation(projection: dict[str, list[Neighbor]]) -> list[list[str]]:
     """
-    Implement the label propagation community detection algorithm with oscillation prevention.
-    
+    Implement the label propagation community detection algorithm.
+
     Algorithm:
     1. Start with each node being assigned its own community
     2. Each node will take on the community of the plurality of its neighbors
     3. Ties are broken by going to the largest community
     4. Continue until no communities change during propagation
-    
+
     Oscillation prevention:
     - Uses asynchronous updates (randomized node order)
     - Maximum iteration limit to prevent infinite loops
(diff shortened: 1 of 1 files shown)


prasmussen15 (changes requested):
Are you able to fix the failing test case?

ZLBillShaw:
> Are you able to fix the failing test case?

Yes, I've fixed the failing test case.

danielchalef:
Closing in favor of #1456 (consolidated fix). Good instinct here — async LPA with oscillation handling — but this patches only `community_operations.py` and misses the duplicate `label_propagation` in `driver/operations/graph_utils.py`. #1456 applies the same async approach to both copies and adds a hard `max_iterations` termination guarantee plus parametrized tests. Thanks for the contribution!

PR #1118, opened 2025-12-20
Fix FalkorDB sanitize to handle forward slash character

david-morales:
## Summary
- Fixes RediSearch query syntax errors when entity names contain forward slashes
- Adds `/` to the separator map in FalkorDB's `sanitize()` method

## Problem
When entity names contained forward slashes (e.g., "Cruise / En route"), RediSearch queries were failing with syntax errors because the forward slash was not being sanitized. This resulted in malformed queries like:
```
(@group_id:\_) (Cruise | / | En | route)
```

## Solution
Added `/` to the `separator_map` in the `sanitize()` method (line 324 in `falkordb_driver.py`), treating it like other special characters by replacing it with whitespace.

Now "Cruise / En route" is sanitized to "Cruise En route", which produces valid RediSearch queries.

## Testing
Verified that:
- "Cruise / En route" → "Cruise En route" 
- "Landing/Taxi" → "Landing Taxi"
- Multiple slashes are handled correctly

Fixes data insertion issues when using FalkorDB backend with forward slash characters in entity names.

### DIFF
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -289,7 +289,7 @@ class FalkorDriver(GraphDriver):
     def sanitize(self, query: str) -> str:
         """
         Replace FalkorDB special characters with whitespace.
-        Based on FalkorDB tokenization rules: ,.<>{}[]"':;!@#$%^&*()-+=~
+        Based on FalkorDB tokenization rules: ,.<>{}[]"':;!@#$%^&*()-+=~/?
         """
         # FalkorDB separator characters that break text into tokens
         separator_map = str.maketrans(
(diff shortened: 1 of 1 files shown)


danielchalef:
Triage note: keeping this open as a **distinct** concern — it adds forward-slash `/` handling to the FalkorDB query-text `sanitize()` layer, which is separate from #1549's group_id escaping and #1475's backtick/empty-query handling. We'll review it on its own. Thanks!

danielchalef:
Thanks again for surfacing the FalkorDB special-char handling. The forward-slash case is already resolved on main by #1183, which added `/`, `|`, and `\` to the `separator_map` (a superset of this PR's `/`). Closing as superseded by #1183.

PR #1151, opened 2026-01-14
fix(falkordb): add missing special characters to sanitize method

### LINKED ISSUE #1144: FalkorDriver sanitize should escape forward/back slashes for RediSearch
## Problem
We are seeing RediSearch query syntax errors when entity names or file paths include forward/back slashes (e.g., `C:\\repo\\src\\main.py` or `some/path/file`). The current `FalkorDriver.sanitize()` does not escape `/` or `\\`, so these strings can break full-text queries.

## Expected
Paths with `/` or `\\` should be sanitized into a safe form before building RediSearch queries, similar to how other special chars are handled.

## Proposed fix
Patch `FalkorDriver.sanitize` (and `build_fulltext_query` output) to replace `/` and `\\` with spaces before/after the existing sanitization logic. Example patch we use downstream:

```python
_original_sanitize = FalkorDriver.sanitize
_original_build_fulltext = FalkorDriver.build_fulltext_query

@functools.wraps(_original_sanitize)
def _patched_sanitize(self, query, *args, **kwargs):
    if isinstance(query, str):
        query = query.replace('/', ' ').replace('\\', ' ')
    return _original_sanitize(self, query, *args, **kwargs)

@functools.wraps(_original_build_fulltext)
def _patched_build_fulltext(self, query, *args, **kwargs):
    if isinstance(query, str):
        query = query.replace('/', ' ').replace('\\', ' ')
    result = _original_build_fulltext(self, query, *args, **kwargs)
    if isinstance(result, str):
        result = result.replace('/', ' ').replace('\\', ' ')
    return result
```

## Context
We currently patch this in downstream apps to avoid RediSearch errors. Upstreaming would let us remove the workaround.

Internal tracking: mem-unify-004

sgaluza:
## Summary

Add forward slash (`/`), backslash (`\`), and pipe (`|`) to the `sanitize()` method to prevent RediSearch query syntax errors.

## Problem

When entity names contain certain special characters, RediSearch queries fail with syntax errors. The current `sanitize()` method handles most special characters but misses three important ones:

| Character | RediSearch Meaning | Issue |
|-----------|-------------------|-------|
| `\|` | OR operator | Breaks query logic |
| `\` | Escape character | Causes parsing errors |
| `/` | Path separator | Causes syntax errors (see #1144) |

For example, an entity name like `"Cruise / En route"` would produce malformed queries.

## Solution

Added the three missing characters to `separator_map` in `sanitize()`:

```python
'/': ' ',   # Forward slash - causes query syntax errors
'\\': ' ',  # Backslash - escape character in RediSearch
'|': ' ',   # Pipe - OR operator in RediSearch
```

Also updated the docstring with:
- Reference to [RediSearch escaping documentation](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/escaping/)
- Note clarifying this method is for fulltext search, not exact path matching

## Research

RediSearch reserved characters that require escaping (from [official docs](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/escaping/)):
- Tokenization separators: `,.<>{}[]"':;!@#$%^&*()-+=~?`
- Query operators: `|` (OR), `-` (negation), `*` and `?` (wildcards)
- Escape character: `\`

The existing code already handles most of these, this PR adds the missing ones.

## Related

- Fixes #1144
- Related to #1118 (adds only `/`, this PR is more complete)

## Test Plan

- [x] Verify `"Cruise / En route"` → `"Cruise En route"`
- [x] Verify `"path\\to\\file"` → `"path to file"`
- [x] Verify `"option1|option2"` → `"option1 option2"`

---

Built with [OnSteroids](https://onsteroids.ai)

### DIFF
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -288,10 +288,21 @@ class FalkorDriver(GraphDriver):
 
     def sanitize(self, query: str) -> str:
         """
-        Replace FalkorDB special characters with whitespace.
-        Based on FalkorDB tokenization rules: ,.<>{}[]"':;!@#$%^&*()-+=~
+        Replace FalkorDB/RediSearch special characters with whitespace.
+
+        This method is designed for full-text search queries where text is split
+        into searchable words. Not suitable for exact path matching (e.g., file
+        paths like '/home/user/file.txt' will become 'home user file txt').
+
+        RediSearch (used by FalkorDB for full-text search) has reserved characters
+        that must be sanitized to prevent query syntax errors:
+        - Tokenization separators: ,.<>{}[]"':;!@#$%^&*()-+=~?
+        - Query operators: | (OR), / and \\ (path separators that break queries)
+
+        See: https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/escaping/
+        Related: https://github.com/getzep/graphiti/issues/1144
         """
-        # FalkorDB separator characters that break text into tokens
+        # RediSearch separator and operator characters
         separator_map = str.maketrans(
             {
                 ',': ' ',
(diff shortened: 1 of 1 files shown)


danielchalef:

sgaluza:

danielchalef:
Thanks for this! Closing — the `sanitize()` separator additions here already landed via #1183, so this is now a no-op against `main`. Appreciate the contribution.

PR #1159, opened 2026-01-17
fix(mcp): pass base_url to Anthropic LLM and OpenAI Embedder clients

Milofax:
## Summary

The `LLMClientFactory` and `EmbedderFactory` in `factories.py` read the `api_url` from config but don't pass it to the client constructors. This causes:

- `AnthropicClient` to always use `https://api.anthropic.com` instead of configured URL
- `OpenAIEmbedder` to always use `https://api.openai.com/v1` instead of configured URL

This breaks self-hosted/proxy setups where users want to route requests through custom endpoints.

## Changes

- Pass `base_url=config.providers.anthropic.api_url` to `GraphitiLLMConfig` for Anthropic
- Pass `base_url=config.providers.openai.api_url` to `OpenAIEmbedderConfig` for OpenAI embedder

## Test Plan

- [x] Tested with CLIProxyAPI (Anthropic-compatible proxy) - requests now correctly route to custom endpoint
- [x] Tested with Ollama (OpenAI-compatible embeddings) - embeddings now correctly use custom endpoint

### DIFF
--- a/mcp_server/src/services/factories.py
+++ b/mcp_server/src/services/factories.py
@@ -205,6 +205,7 @@ class LLMClientFactory:
 
                 llm_config = GraphitiLLMConfig(
                     api_key=api_key,
+                    base_url=config.providers.anthropic.api_url,
                     model=config.model,
                     temperature=config.temperature,
                     max_tokens=config.max_tokens,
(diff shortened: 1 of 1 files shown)


Milofax:
Just checking in — this ensures base_url is properly passed to both the Anthropic LLM and OpenAI Embedder clients, which is needed for self-hosted setups. Would appreciate a look when you have time!

danielchalef:
Thanks for this. The OpenAI embedder `base_url` support is now on main (`EmbedderFactory` passes `base_url=config.providers.openai.api_url`), and the LLM base_url routing landed via the factory refactor. Closing as superseded; if there's a remaining Anthropic-specific base_url need, let's track it as a focused issue.

PR #1170, opened 2026-01-22
fix(falkordb): Clone driver for single group_id search

### LINKED ISSUE #1161: [BUG] FalkorDB search fails with single group_id - driver not cloned to correct graph
## Bug Description

When using FalkorDB as the database backend, searching with a **single group_id** returns no results even though data exists. This is because the `handle_multiple_group_ids` decorator only clones the driver to the correct graph when `len(group_ids) > 1`.

In FalkorDB, each `group_id` is stored as a **separate graph (database)**, not as a property within a single graph. The current implementation assumes group_id is just a filter property, which works for Neo4j but fails for FalkorDB's multi-graph architecture.

## Steps to Reproduce

```python

# Setup
driver = FalkorDriver(host="localhost", port=6379, database="default_db")
graphiti = Graphiti(graph_driver=driver)

# Add episode with specific group_id (creates "my-project" graph in FalkorDB)
await graphiti.add_episode(
    name="test",
    episode_body="React 19 is used in this project",
    group_id="my-project",
    source_description="test"
)

# Search with single group_id - returns empty!
results = await graphiti.search(
    query="React",
    group_ids=["my-project"],  # Single group_id
    num_results=10
)
print(len(results))  # 0 - but data exists in "my-project" graph
```

## Expected Behavior

Search should return results from the `my-project` graph when `group_ids=["my-project"]` is specified.

## Actual Behavior

Search returns 0 results because:
1. The decorator condition `len(group_ids) > 1` is `False` for single group_id
2. Driver is not cloned to the correct graph
3. Query executes against `default_db` instead of `my-project` graph
4. Data exists in `my-project` graph but is never searched

## Environment
- **Graphiti Version**: 0.25.0 - 0.26.0
- **Python Version**: 3.13
- **Operating System**: macOS 15.2 (also reproducible on Linux)
- **Database Backend**: FalkorDB latest (Docker)
- **LLM Provider & Model**: OpenAI gpt-4o

## Installation Method
- [x] pip install

## Error Messages/Traceback

No error is thrown - the search simply returns empty results silently.

Verification that data exists in the correct graph:
```bash
# Direct FalkorDB query shows data exists
redis-cli GRAPH.QUERY my-project "MATCH (n) RETURN count(n)"
# Returns: count > 0

# But default_db is empty
redis-cli GRAPH.QUERY default_db "MATCH (n) RETURN count(n)"  
# Returns: 0
```

## Configuration

```python
# graphiti_core/decorators.py (current code)
def handle_multiple_group_ids(func: F) -> F:
    async def wrapper(self, *args, **kwargs):
        ...
        # Only handle FalkorDB with multiple group_ids
        if (
            hasattr(self, 'clients')
            and hasattr(self.clients, 'driver')
            and self.clients.driver.provider == GraphProvider.FALKORDB
            and group_ids
            and len(group_ids) > 1  # ← Bug: should be >= 1
        ):
            # Clone driver for each group_id
            ...
```

## Additional Context

- **Consistency**: This happens 100% of the time with single group_id on FalkorDB
- **Component**: Core library (`graphiti_core`)
- **Related**: FalkorDB uses separate graphs per group_id (by design), unlike Neo4j which uses a single graph with group_id as a property

The issue can be verified by checking the graph list in FalkorDB:
```bash
redis-cli GRAPH.LIST
# Shows: my-project, default_db, etc. (each group_id becomes a graph)
```

## Possible Solution

Change the condition in `graphiti_core/decorators.py` from `> 1` to `>= 1`:

```python
# Before
and len(group_ids) > 1

# After  
and len(group_ids) >= 1
```

This ensures the driver is cloned to the correct graph even when a single group_id is specified. I can submit a PR if this approach is acceptable.

himorishige:
## Summary
- Fix FalkorDB search failing when using a single group_id
- Change decorator condition from `len(group_ids) > 1` to `len(group_ids) >= 1`
- Update comments to clarify FalkorDB's multi-graph architecture

## Type of Change
- [x] Bug fix

## Objective
FalkorDB stores each `group_id` as a separate graph (database), not as a property within a single graph. The `handle_multiple_group_ids` decorator was only cloning the driver when multiple group_ids were specified, causing searches with a single group_id to execute against `default_db` instead of the target graph.

## Testing
- [x] Verified fix resolves the issue locally with FalkorDB
- [x] All existing tests pass

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1161

### DIFF
--- a/graphiti_core/decorators.py
+++ b/graphiti_core/decorators.py
@@ -44,13 +44,14 @@ def handle_multiple_group_ids(func: F) -> F:
         if group_ids is None and group_ids_pos is not None and len(args) > group_ids_pos:
             group_ids = args[group_ids_pos]
 
-        # Only handle FalkorDB with multiple group_ids
+        # Handle FalkorDB where each group_id maps to a separate graph (database)
+        # For FalkorDB, we need to clone the driver to use the correct graph even for a single group_id
         if (
             hasattr(self, 'clients')
             and hasattr(self.clients, 'driver')
             and self.clients.driver.provider == GraphProvider.FALKORDB
             and group_ids
-            and len(group_ids) > 1
+            and len(group_ids) >= 1
         ):
             # Execute for each group_id concurrently
             driver = self.clients.driver


danielchalef:

himorishige:

Naseem77:
@himorishige Tested it: single-group reads still return 0 with this patch. #1326 gets further on the same bug, suggest consolidating there.

himorishige:
@Naseem77 Thanks for taking the time to test this and for the heads-up. You're right — the decorator condition was only part of the problem, and this patch doesn't cover the `get_by_group_ids` classmethods that bypass the decorator entirely, so single-group reads can still hit the default graph.

Since #1326 includes the same decorator change plus routing for those paths, consolidating there makes sense to me. Closing this PR in favor of #1326. Hope it lands soon!

PR #1176, opened 2026-01-25
fix(mcp): prevent queue worker task garbage collection

Milofax:
## Summary

- Fix asyncio task reference bug in QueueService where worker tasks could be garbage collected mid-execution
- Add documentation about in-memory queue limitations

## Problem

Worker tasks created with `asyncio.create_task()` were not stored in a strong reference, allowing Python's garbage collector to collect them mid-execution. This is explicitly warned against in the Python documentation:

> "Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution."
> https://docs.python.org/3/library/asyncio-task.html#creating-tasks

**Symptom:** The queue worker task could silently disappear without running its `finally` block, causing `_queue_workers[group_id]` to remain `True` indefinitely while the actual worker was gone. This resulted in queued episodes never being processed since the check `if not self._queue_workers.get(group_id, False)` would return `False` (worker "running") but no task was actually processing.

## Solution

- Add `_worker_tasks` dict to store task references per group_id
- Store task reference after `asyncio.create_task()`
- Add done callback to clean up references when tasks complete
- Document in-memory queue limitation in class docstring

## Test plan

- [x] Code review of asyncio task handling patterns
- [x] Verified worker tasks are now stored in `_worker_tasks` dict
- [x] Confirmed done callback properly cleans up references

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/mcp_server/src/services/queue_service.py
+++ b/mcp_server/src/services/queue_service.py
@@ -10,7 +10,13 @@ logger = logging.getLogger(__name__)
 
 
 class QueueService:
-    """Service for managing sequential episode processing queues by group_id."""
+    """Service for managing sequential episode processing queues by group_id.
+
+    Note: This implementation uses an in-memory queue (asyncio.Queue) which does not
+    persist across service restarts. Messages queued but not yet processed will be lost
+    if the service crashes or restarts. For production deployments requiring durability,
+    consider using a persistent message broker (e.g., Redis Streams, RabbitMQ).
+    """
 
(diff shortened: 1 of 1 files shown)


Milofax:
Hi team! Just checking in — this PR prevents queue worker tasks from being garbage collected, which we've been hitting in production. CI approval is still pending. Would love to get this reviewed when you have a moment. Thanks!

PR #1197, opened 2026-02-03
fix(edges): preserve all signatures when edge type is reused across node pairs

prasmussen15:
## Summary
- Fix bug where edge types used across multiple node type pairs would lose all but the last signature
- Changed `edge_type_signature_map` (singular) to `edge_type_signatures_map` (plural) which maps each edge type to a **list** of valid signatures
- Updated `edge_types_context` to use `fact_type_signatures` containing all valid node type combinations

## Problem
When defining an `edge_type_map` with the same edge type for multiple node type pairs:

```python
edge_type_map = {
    ("User", "Person"): ["InterpersonalRelationship"],
    ("Person", "Person"): ["InterpersonalRelationship"],  # Same type, different signature
}
```

The dict comprehension was overwriting previous entries:
```python
edge_type_signature_map: dict[str, tuple[str, str]] = {
    edge_type: signature
    for signature, edge_types in edge_type_map.items()
    for edge_type in edge_types
}
```

This caused `InterpersonalRelationship` to only retain `("Person", "Person")`, losing `("User", "Person")`.

## Solution
Build a mapping from edge type name to a list of valid signatures:
```python
edge_type_signatures_map: dict[str, list[tuple[str, str]]] = {}
for signature, edge_type_names in edge_type_map.items():
    for edge_type in edge_type_names:
        if edge_type not in edge_type_signatures_map:
            edge_type_signatures_map[edge_type] = []
        edge_type_signatures_map[edge_type].append(signature)
```

## Test plan
- [ ] Verify edge types with multiple signatures are all preserved in `edge_types_context`
- [ ] Verify LLM receives complete signature information for all valid edge type combinations
- [ ] Verify existing single-signature edge types continue to work correctly

### DIFF
--- a/tests/utils/maintenance/test_edge_operations.py
+++ b/tests/utils/maintenance/test_edge_operations.py
@@ -445,3 +445,117 @@ async def test_resolve_extracted_edges_fast_path_deduplication(monkeypatch):
     assert resolve_call_count == 1
     assert len(resolved_edges) == 1
     assert invalidated_edges == []
+
+
+class InterpersonalRelationship(BaseModel):
+    """A relationship between two people."""
+
+
+class LocatedIn(BaseModel):
+    """A relationship indicating something is located in a place."""
+
+
+def test_edge_type_signatures_map_preserves_multiple_signatures():
+    """Test that edge types used across multiple node type pairs preserve all signatures.
+
+    This tests the fix for the bug where dict comprehension would overwrite
+    previous signatures when the same edge type appeared in multiple node pairs.
+    """
+    # Edge type map where the same edge type is used for multiple node pair signatures
+    # This is the scenario that was broken before the fix
+    edge_type_map: dict[tuple[str, str], list[str]] = {
+        ('Person', 'Person'): ['InterpersonalRelationship'],
+        ('Person', 'Entity'): ['InterpersonalRelationship'],  # Same type, different signature
+        ('Person', 'City'): ['LocatedIn'],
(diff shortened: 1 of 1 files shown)


prasmussen15:
## Addressed Review Comments

### 1. Test Coverage Added

Added two unit tests to `tests/utils/maintenance/test_edge_operations.py`:

- **`test_edge_type_signatures_map_preserves_multiple_signatures`**: Verifies that when the same edge type (e.g., `InterpersonalRelationship`) is used for multiple node pair signatures (e.g., `(Person, Person)` and `(Person, Entity)`), ALL signatures are preserved in `edge_types_context`.

- **`test_edge_type_signatures_map_single_signature_still_works`**: Verifies backward compatibility when edge types have only a single signature.

### 2. Prompt Schema Compatibility

The change from `fact_type_signature` (singular tuple) to `fact_type_signatures` (list of tuples) is **backward compatible** because:

1. The prompt simply converts the `edge_types` context to JSON via `to_prompt_json()` and includes it in `<FACT_TYPES>` tags
2. The LLM instruction says: "If FACT_TYPES are provided and the relationship matches one of the types **(considering the entity type signature)**, use that fact_type_name"
3. The LLM receives the JSON structure directly - it will now see a list of valid signatures instead of a single one, which gives it MORE information, not less
4. There's no strict schema validation on the LLM side - it's free-form JSON in the prompt

The change is purely additive from the LLM's perspective: instead of seeing one valid signature, it now sees all valid signatures for each edge type.

PR #1220, opened 2026-02-11
fix(search): use OR syntax for FalkorDB multi-label node filters

david-morales:
## Summary

- FalkorDB rejects Neo4j pipe syntax (`n:Label1|Label2`) when filtering by multiple node labels. This adds FalkorDB-specific branches in both `node_search_filter_query_constructor` and `edge_search_filter_query_constructor` that generate `(n:Label1 OR n:Label2)` instead.

## Problem

`search_filters.py` has an `else` branch that uses `'|'.join(filters.node_labels)` to build label filters. This works for Neo4j but FalkorDB doesn't support the pipe multi-label syntax:

```
Invalid input '|': expected an expression (line 1, column 66 (offset: 65))
"... WHERE n:Occurrence|Operator ..."
```

The bug only manifests with 2+ labels — single-label filters work because `'|'.join(["Label"])` produces `"Label"` which is valid on all providers.

## Fix

Added `elif provider == GraphProvider.FALKORDB` branches that generate OR-based syntax:
- Node search: `(n:Label1 OR n:Label2)`
- Edge search: `(n:Label1 OR n:Label2) AND (m:Label1 OR m:Label2)`

## Test plan

- [x] 17 new unit tests covering single-label, multi-label (2 and 3 labels), and empty label cases across all providers (Neo4j, FalkorDB, KUZU, Neptune)
- [x] Existing search_utils tests still pass
- [x] Verified fix against live FalkorDB instance

### DIFF
--- a/graphiti_core/search/search_filters.py
+++ b/graphiti_core/search/search_filters.py
@@ -87,6 +87,9 @@ def node_search_filter_query_constructor(
         if provider == GraphProvider.KUZU:
             node_label_filter = 'list_has_all(n.labels, $labels)'
             filter_params['labels'] = filters.node_labels
+        elif provider == GraphProvider.FALKORDB:
+            parts = [f'n:{label}' for label in filters.node_labels]
+            node_label_filter = '(' + ' OR '.join(parts) + ')'
         else:
             node_labels = '|'.join(filters.node_labels)
             node_label_filter = 'n:' + node_labels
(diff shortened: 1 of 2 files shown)


Naseem77:
@david-morales Confirmed live: FalkorDB throws Invalid input '|' on the old syntax and your OR rewrite returns the right nodes. Tests pass. LGTM!

prasmussen15:
Users need to sign the CLA

prasmussen15 (changes requested):
Creator needs to sign the CLA

Issue #1177, opened 2026-01-25
RFC: Replace in-memory queue with Redis Streams for crash-resilient episode processing


Milofax:
## Summary

Replace the current in-memory `asyncio.Queue` in the MCP server's `QueueService` with Redis Streams for persistent, crash-resilient message processing.

## Motivation

The current implementation uses an in-memory queue that loses all pending episodes on service restart:

```python
# Current implementation
self._episode_queues: dict[str, asyncio.Queue] = {}
```

**Problems:**
1. **Data Loss**: Episodes queued but not yet processed are lost on crash/restart
2. **No Recovery**: After restart, there's no way to know what was lost
3. **Silent Failure**: Users receive "queued for processing" but episodes disappear

**Real-world impact**: In our production deployment, we lost episodes during a routine container restart. The response said "queued" but they never appeared in the graph.

## Proposed Solution

Use Redis Streams with Consumer Groups for guaranteed delivery:

```python
# Proposed implementation
class QueueService:
    async def add_episode_task(self, ...) -> str:
        # XADD - immediately persisted by Redis
        message_id = await self._redis.xadd(stream_key, message.to_dict())
        return message_id
    
    async def _process_stream(self, group_id: str):
        # XREADGROUP with Consumer Groups
        # XACK on success
        # xautoclaim for abandoned messages after crash
```

### Key Features

1. **Persistence**: Messages survive service restarts (Redis RDB/AOF)
2. **Guaranteed Delivery**: Consumer Groups with XACK
3. **Crash Recovery**: `xautoclaim` reclaims abandoned messages
4. **Backward Compatible**: Same API, just persistent

### Required Changes

1. **queue_service.py** (~400 LOC change):
   - Replace asyncio.Queue with Redis Streams
   - Add Consumer Group management
   - Add xautoclaim for crash recovery
   - **Important**: EpisodeType enum must be converted with `EpisodeType.from_str()` after deserialization (Redis serializes enums as strings)

2. **config/schema.py** (~20 LOC):
   - Add Redis connection configuration
   - Add consumer group settings

3. **Dependencies**:
   - Add `redis[hiredis]` to pyproject.toml

### Message Flow

```
add_memory() 
  → QueueService.add_episode_task()
    → XADD to Redis Stream
    → Return message_id immediately

Worker (background):
  → XREADGROUP (blocking read)
  → Process episode with graphiti_client.add_episode()
  → XACK on success (removes from pending)
[...]

Issue #1171, opened 2026-01-22
[BUG] Entity extraction fails to include pronoun-referenced entities from previous messages, causing incorrect edge creation


JelionNg:
---

### Title

**Entity extraction fails to include pronoun-referenced entities from previous messages, causing incorrect edge creation**

---

### Description

When a pronoun in the current message references an entity from previous messages, that entity is not included in the extracted entities list. This causes the subsequent edge creation to link relationships incorrectly.

### Reproduction Steps

1. Start a conversation with the following turns:

**Turn 1:**
```
User: This is my friend John, he's a software engineer at Google.
Assistant: Got it, I've noted that John is your friend and works at Google as a software engineer.
```

**Turn 2:**
```
User: He also has a friend named Mike, they often play basketball together.
```

2. Check the extracted entities for Turn 2.

### Expected Behavior

**Extracted entities:** `[User, John, Mike]`

Since "He" in Turn 2 refers to "John" from Turn 1, John should be extracted as an implicitly mentioned entity.

**Expected edge:** `John -[FRIEND_OF]-> Mike`

### Actual Behavior

**Extracted entities:** `[User, Mike]`

John is excluded because the prompt states to "Exclude entities mentioned only in the PREVIOUS MESSAGES."

**Actual edge:** `User -[FRIEND_OF]-> Mike` ❌

The edge is incorrectly created between User and Mike, because John is not available in the entity list during edge creation.

### Root Cause Analysis

The issue is in `graphiti_core/prompts/extract_nodes.py`. There are **conflicting instructions**:

1. **Line 105-107** says pronouns should be disambiguated:
   > "Pronoun references such as he/she/they should be disambiguated to the names of the reference entities."

2. **Line 114-115** says to exclude entities from previous messages:
   > "Exclude entities mentioned only in the PREVIOUS MESSAGES (they are for context only)."

The LLM interprets pronoun references as NOT counting toward "implicit mention," so it follows the exclusion rule and drops the referenced entity. This creates an **information disconnect** between the entity extraction phase and the edge creation phase.

Interestingly, the edge extraction prompt (`extract_edges.py`) handles pronoun disambiguation correctly in the `fact` field — the LLM understands "He" refers to John. But since John wasn't extracted as an entity, the edge cannot be created correctly.

### Suggested Fix

**Two modifications are required** (both are necessary, as tested):

#### 1. Add explicit instruction at the beginning (around line 105-107)

**Current:**
```python
[...]

PR #1267, opened 2026-02-24
fix: resilient async worker and phantom UUID edge filter

rafaelreis-r:
## Summary

Two bug fixes developed and validated in a self-hosted Docker deployment. Each fix addresses a distinct failure mode observed in production with `gpt-4.1-mini`/`gpt-4.1-nano` models.

> **Note:** A third planned fix (`DEFAULT_MAX_TOKENS` 8192 → 16384) turned out to be already present in `upstream/main` — so it is not included here.

---

## Fix 1: Resilient async worker with retry and error classification (`server/graph_service/routers/ingest.py`)

**Problem:** `AsyncWorker.worker()` only caught `asyncio.CancelledError`. Any other exception (e.g., `RateLimitError`, Neo4j `SocketDeadlineExceededError`) would silently kill the coroutine, leaving jobs permanently stuck in the `asyncio.Queue`. The endpoint returned `202 Accepted` but jobs were never processed.

**Fix:** Added a retry loop with exponential backoff (up to 5 attempts: 5/10/20/40 s) and error classification:
- **Permanent errors** (`IndexError`, `KeyError`, `ValueError`, `BadRequestError`, HTTP 400): skipped immediately with a clear log message — no point retrying the same data.
- **Transient errors** (rate limits, network/connection issues): retried up to `MAX_RETRIES` times with exponential backoff.
- All outcomes are logged at appropriate levels (`logger.warning` for retries, `logger.error` for final failures) for observability.

Related: #1177

---

## Fix 2: Filter phantom UUID references in `resolve_extracted_edges` (`graphiti_core/utils/maintenance/edge_operations.py`)

**Problem:** When the LLM is inconsistent between the node-extraction and edge-extraction steps (a known LLM non-determinism issue), edges may reference UUIDs of nodes that were never created in the same batch. The upstream code already attempts to fetch those missing nodes from the database via `EntityNode.get_by_uuids`, but if the UUID is truly phantom (hallucinated — never persisted anywhere), the node is still absent from `uuid_entity_map` and downstream code raises a `KeyError`, aborting the entire episode.

**Fix:** Immediately after the DB lookup, filter out any `extracted_edges` whose `source_node_uuid` or `target_node_uuid` is still not present in `uuid_entity_map`. A warning is logged when edges are dropped. The `related_edges_lists` and `edge_invalidation_candidates` parallel lists are filtered in lockstep to keep all indices consistent. No valid data is lost — these edges cannot be persisted regardless since their referenced nodes do not exist anywhere in the graph.

Related: #1171

---

## Testing

Both fixes were deployed and validated in a self-hosted Docker environment processing ~300 episodes across multiple knowledge groups.

Before the fixes:
- Worker coroutines died silently on `RateLimitError` (confirmed via queue monitoring) — `202 Accepted` returned but no processing occurred
- ~50% of episode batches failed with `KeyError` in `resolve_extracted_edges` due to phantom UUIDs from LLM inconsistency

After the fixes: zero silent failures observed, episode processing completed successfully across all batches.

### DIFF
--- a/server/graph_service/routers/ingest.py
+++ b/server/graph_service/routers/ingest.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 from contextlib import asynccontextmanager
 from functools import partial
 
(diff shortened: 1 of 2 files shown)


danielchalef:

danielchalef:
Maintainer edit — thanks for this, the worker-resilience fix is a real and valuable catch. I've rebased the branch onto current `main` and scoped it to **Fix 1 only**, with a few changes:

**Kept & hardened (`server/graph_service/routers/ingest.py`):**
- Extracted the retry/classify logic into `AsyncWorker._process_job()` so it's unit-testable, and consolidated the dual `print()`+`logger` output to a single `logger` path.
- Adjusted the permanent-error classification so HTTP **429 (rate limit)** and **408 (timeout)** are treated as *transient* (retryable). The previous `status_code == 400` plus name-matching was close, but a blanket check risked classifying rate limits as permanent — now permanent status codes are an explicit `{400, 404, 422}` set, and rate limits/network errors fall through to the backoff retry path (the primary motivating case).
- Added `server/tests/test_ingest_worker.py` covering: transient-retry-then-succeed, permanent-skip-no-retry, give-up-after-MAX_RETRIES, `CancelledError` propagation, and worker-survives-a-failing-job-and-processes-the-next.

**Dropped — Fix 2 (phantom-UUID edge filter in `edge_operations.py`):** this is already handled in current `main`, so the filter is redundant:
- Edges are now bound to extracted nodes **by name** in `extract_edges` — `source_entity_name`/`target_entity_name` are validated against the extracted node list and any out-of-list reference is dropped before a UUID is ever assigned (`edge_operations.py:216-239`). A hallucinated node reference can't produce a phantom UUID in the first place.
- `resolve_extracted_edges` reads endpoints via `uuid_entity_map.get(...)` with a `None` fallback (`edge_operations.py:462-463`), so the `KeyError` your repro hit no longer occurs.

The underlying phantom-reference behavior you saw is really a model-quality issue (smaller/local models hallucinating entities), and the name-based binding addresses it at the source.

Net: this PR now carries just the worker fix + tests. Appreciate the clear writeup and the honest note about the already-upstream max-tokens change.

PR #1294, opened 2026-03-03
fix: prevent shared driver state mutation in multi-group FalkorDB usage

StephenBadger:
## Summary

- **`add_episode()` and `add_episode_bulk()` mutate `self.driver` and `self.clients.driver`** when switching FalkorDB graphs for different `group_id`s. In a multi-tenant server with one shared `Graphiti` client, writes to one group silently redirect all subsequent reads for other groups to the wrong FalkorDB graph database, returning empty results.
- **Root cause**: FalkorDB uses a separate graph database per `group_id`. `add_episode()` clones the driver to point at the target group's graph but stores the clone back into `self.driver`/`self.clients.driver`, permanently changing which graph the shared instance targets. Concurrent or interleaved operations targeting different groups corrupt each other's queries.
- **`@handle_multiple_group_ids` decorator only cloned drivers for `len(group_ids) > 1`**, so single-group searches used whatever stale driver state the last `add_episode` left behind.

## Fix

- `add_episode()` and `add_episode_bulk()` now create **local** `driver`/`clients` via `clone()` and `model_copy()` instead of mutating shared instance state
- All private methods (`_extract_and_resolve_edges`, `_process_episode_data`, `_get_or_create_saga`, `_extract_and_dedupe_nodes_bulk`, `_resolve_nodes_and_edges_bulk`) accept optional `driver`/`clients` params so the local clone flows through without touching `self`
- `@handle_multiple_group_ids` decorator now clones the driver for single `group_id` queries too (`>= 1` instead of `> 1`)

## Reproduction

1. Create a `Graphiti` instance with a FalkorDB driver
2. Call `add_episode(group_id="group-a", ...)` — driver now points to `group-a` graph
3. Call `add_episode(group_id="group-b", ...)` — driver flips to `group-b` graph
4. Call `search_(group_ids=["group-a"], ...)` — **returns empty results** because driver is querying `group-b` graph

### DIFF
--- a/graphiti_core/decorators.py
+++ b/graphiti_core/decorators.py
@@ -28,8 +28,10 @@ F = TypeVar('F', bound=Callable[..., Awaitable[Any]])
 
 def handle_multiple_group_ids(func: F) -> F:
     """
-    Decorator for FalkorDB methods that need to handle multiple group_ids.
-    Runs the function for each group_id separately and merges results.
+    Decorator for FalkorDB methods that need to handle group_ids.
+    Each group_id maps to a separate FalkorDB graph database, so this
+    decorator clones the driver to point at the correct graph for each
+    group_id and runs the function per-group, merging results.
     """
 
     @functools.wraps(func)
@@ -44,13 +46,16 @@ def handle_multiple_group_ids(func: F) -> F:
         if group_ids is None and group_ids_pos is not None and len(args) > group_ids_pos:
             group_ids = args[group_ids_pos]
 
-        # Only handle FalkorDB with multiple group_ids
+        # Handle FalkorDB group_ids — each group_id is a separate graph database,
+        # so we must clone the driver to target the correct one.  Previously this
+        # only handled len > 1, which meant single-group queries used whatever
+        # driver state was left over from the last add_episode call.
         if (
             hasattr(self, 'clients')
             and hasattr(self.clients, 'driver')
             and self.clients.driver.provider == GraphProvider.FALKORDB
             and group_ids
-            and len(group_ids) > 1
+            and len(group_ids) >= 1
         ):
             # Execute for each group_id concurrently
             driver = self.clients.driver
--- a/graphiti_core/graphiti.py
+++ b/graphiti_core/graphiti.py
@@ -340,7 +340,13 @@ class Graphiti:
         """
         await self.driver.close()
 
-    async def _get_or_create_saga(self, saga_name: str, group_id: str, now: datetime) -> SagaNode:
+    async def _get_or_create_saga(
+        self,
+        saga_name: str,
+        group_id: str,
+        now: datetime,
+        driver: GraphDriver | None = None,
+    ) -> SagaNode:
         """
         Get an existing saga by name or create a new one.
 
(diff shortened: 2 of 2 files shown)


danielchalef:

StephenBadger:

StephenBadger:
recheck

nidhishgajjar:
**Orb Code Review** (powered by GLM 5.1 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR fixes a critical concurrency bug where `add_episode()` and `add_episode_bulk()` mutate shared instance state (`self.driver` and `self.clients.driver`) when switching FalkorDB graphs for different `group_id`s. In a multi-tenant server sharing a single `Graphiti` instance, this causes writes to one group to silently redirect all subsequent operations to the wrong FalkorDB graph database.

The fix creates **local** `driver`/`clients` variables via `clone()` and `model_copy()` instead of mutating shared state, and threads them through private helper methods via optional parameters.

## Architecture

The change correctly identifies and fixes the root cause in two places:
1. **`add_episode()`** and **`add_episode_bulk()`** — replaces `self.driver = self.driver.clone(...)` with local `driver`/`clients` variables
2. **`@handle_multiple_group_ids` decorator** — fixes the condition from `len(group_ids) > 1` to `>= 1` so single-group queries also get driver cloning
3. **Private methods** (`_extract_and_resolve_edges`, `_process_episode_data`, `_get_or_create_saga`, `_extract_and_dedupe_nodes_bulk`, `_resolve_nodes_and_edges_bulk`) accept optional `driver`/`clients` params to thread the local clone through

## Issues

No critical or blocking issues found. The approach is correct.

### Warning: Other methods may still have latent state issues

**File:** `graphiti_core/graphiti.py`  
**Severity:** suggestion  

I verified that `self.driver = self.driver.clone(...)` only occurs in `add_episode()` (line 1017) and `add_episode_bulk()` (line 1243), so this PR covers all current mutation points. However, methods like `search()`, `search_()`, `add_triplet()`, and `remove_episode()` all use `self.driver` or `self.clients` directly. These should be safe because they don't mutate the driver, but future contributors should be aware that FalkorDB multi-group support requires passing a cloned driver rather than relying on instance state.

**Suggestion:** Consider adding a code comment or guard near the `self.driver` property assignment (around line 208) warning against mutation in async contexts, e.g.:

```python
# WARNING: Do not mutate self.driver after initialization.
# FalkorDB requires per-call driver clones for multi-group support.
# See add_episode() for the correct pattern.
```

### Suggestion: Test coverage for the race condition

**File:** Tests  
**Severity:** suggestion  

The fix addresses a real concurrency bug but I don't see corresponding test coverage. A regression test that calls `add_episode()` with two different `group_id`s concurrently and verifies that each group's data is isolated would be valuable.

## Cross-file Impact

The `@handle_multiple_group_ids` decorator change in `decorators.py` affects all FalkorDB methods that use the decorator. The change from `> 1` to `>= 1` means single-group calls now go through the clone path — this is correct but slightly changes performance characteristics for single-group FalkorDB calls (one extra `clone()` call). This should be negligible since `clone()` is expected to be lightweight.

The use of `model_copy(update={'driver': driver})` on `GraphitiClients` (a Pydantic `BaseModel`) is the correct pattern — it creates a shallow copy with only the driver replaced, while `llm_client`, `embedder`, `cross_encoder`, and `tracer` remain shared references.

## Assessment

✅ **Approve** — This is a well-targeted fix for a real concurrency bug that would cause silent data corruption in multi-tenant FalkorDB deployments. The approach of using local variables instead of mutating shared state is the correct pattern. The changes are thorough and thread the driver clone through all relevant code paths.

Naseem77:
@StephenBadger This no longer applies to main, real conflicts. Could you rebase so it can be reviewed properly?

prasmussen15 (changes requested):
Please rebase this branch onto current main and resolve the resulting conflicts before review. The current diff no longer applies cleanly.

PR #1305, opened 2026-03-07
Fix/falkordb group id race condition

### LINKED ISSUE #1331: [BUG] FalkorDB: add_episode() mutates shared self.driver causing cross-group data contamination
## Bug Description

When multiple `group_id` episodes are processed concurrently (e.g., via the MCP server's per-group queue workers), `add_episode()` mutates `self.driver` on the shared `Graphiti` instance, causing data to be written to the wrong FalkorDB graph.

## Root Cause

In `graphiti.py`, lines 887-889 of `add_episode()`:

```python
```

This mutates the shared instance's `self.driver`. When two async tasks call `add_episode()` concurrently with different group_ids (e.g., `patroni` and `cicd`), they race to reassign `self.driver`. Whichever runs last wins, and the other task's writes go to the wrong FalkorDB graph.

## Reproduction

```python
import asyncio

graphiti = Graphiti(...)  # FalkorDB backend

# Concurrent add_episode calls with different group_ids
await asyncio.gather(
    graphiti.add_episode(name="ep1", episode_body="data for group A", group_id="group_a", ...),
    graphiti.add_episode(name="ep2", episode_body="data for group B", group_id="group_b", ...),
)

# Result: ep1's nodes may end up in the "group_b" FalkorDB graph (or vice versa)
```

Verified via direct FalkorDB query:
```bash
redis-cli GRAPH.QUERY group_a "MATCH (n) RETURN n.name, n.group_id LIMIT 10"
# Shows nodes with group_id="group_b" — cross-contaminated
```

## Impact

- Data silently written to wrong FalkorDB graph
- Search returns incomplete results (data is in a different graph than expected)
- `clear_graph` fails to clean up (nodes have mismatched group_id vs actual graph location)
- Affects the MCP server which uses per-group async queue workers sharing one Graphiti instance

## Proposed Fix

`add_episode()` should use a local driver clone instead of mutating `self.driver`:

```python
else:
    driver = self.driver
# Use `driver` for all operations in this call, don't assign to self
```

## Workaround

Serialize all `add_episode()` calls through a single global queue (no concurrent processing across groups).

## Related

- #1161 — FalkorDB search fails with single group_id
- #1325 / #1326 — handle_multiple_group_ids skips graph routing for single group_id

Both issues share the same root architectural problem: FalkorDB uses separate graph databases per group_id, but the driver selection is done via shared mutable state.

### LINKED ISSUE #1676: [BUG] Concurrent multi-group_id episode processing silently corrupts data across FalkorDB graphs (shared self.driver mutation race)
## Bug Description

## Steps to Reproduce
```python
# /// script
# ///
import asyncio
from mcp import ClientSession

        "group_id": group_id,
        "source": "text",
    })

async def main():
            await session.initialize()
            await asyncio.gather(
                add(session, "groupA", 1),
                add(session, "groupB", 1),
                add(session, "groupA", 2),
                add(session, "groupB", 2),
            )

asyncio.run(main())
```
```
```

## Expected Behavior

## Actual Behavior

## Environment
- **Python Version**: 3.12.3

## Installation Method

## Error Messages/Traceback

## Configuration

## Additional Context

  ```python
  ```

## Possible Solution

### LINKED ISSUE #1795: FalkorDB: writes mutate the shared driver (`self.driver = self.driver.clone(database=group_id)`) — reads then answer from the *last-written* group's graph, silently returning empty for other groups
## Summary

On the FalkorDB backend, each group gets its own graph, and `add_episode` /
`add_triplet` permanently switch the shared `Graphiti` client's driver to the
group they just wrote:

- `graphiti_core/graphiti.py:1081` — `self.driver = self.driver.clone(database=group_id)`
- `graphiti_core/graphiti.py:1309` — same in the triplet path
  (line numbers from `main` @ `993e081a`; behavior identical in 0.29.2)

Any long-lived multi-group client — the MCP server is exactly that — then runs
**all subsequent reads inside whichever group wrote last**. `get_episodes` /
searches for another group execute in the wrong graph with a
`group_id` property filter that matches nothing, and return empty with no
error. Scoped memory appears to work until a second group writes; after that,
cross-group reads silently return nothing.

## Environment

- graphiti-core 0.29.2 (code path unchanged on current `main` @ `993e081a`)
- MCP server (streamable HTTP), FalkorDB v4.20.4, falkordb-py 1.2.0

## Reproduction

1. One MCP server / one `Graphiti` client, FalkorDB backend.
2. `add_memory(..., group_id="a")` and wait for it to land.
3. `get_episodes(group_ids="a")` → returns the episode (driver sits on `a`).
4. `add_memory(..., group_id="b")` (or just let queued group-`b` work run).
5. `get_episodes(group_ids="a")` → **empty**, though the data exists.

## Evidence

Captured live with `redis-cli MONITOR` while calling
`get_episodes(group_ids="quarantine")` right after group `helpscout-sync` had
been written:

```
GRAPH.QUERY "helpscout-sync" "MATCH (e:Episodic) WHERE e.group_id IN $group_ids ..."
```

— the Episodic query for group `quarantine` executed against the
`helpscout-sync` graph. Direct inspection confirmed the data exists where it
was written:

```
GRAPH.QUERY quarantine "MATCH (e:Episodic) RETURN e.name"  → smoke-temp-fix
```

## Two related design problems

1. **Mutating shared state:** `clone()` returning a rerouted driver is fine;
   assigning it to `self.driver` on a shared client is the bug. Reads have no
   corresponding per-call routing.
2. **Per-group graphs make `group_ids` lists unsatisfiable:** a search with
   `group_ids=["a","b"]` can only ever execute inside one graph, so
   cross-group search silently degrades to single-group even when routing is
   "correct".

## Suggested fix

Route per operation instead of mutating: use a local
`driver = self.driver.clone(database=group_id)` for writes, and route reads by
their requested `group_ids`. For the list-search case, either fan out across
graphs and merge, or offer a **single-graph mode** for FalkorDB in which all
groups live in the configured graph and scoping relies on the existing
`group_id` property filters (this is how the Neo4j path already behaves).

We deployed the single-graph variant locally (make `FalkorDriver.clone()` a
no-op so `_database` stays fixed) and all semantics — scalar reads after
interleaved writes, list searches, isolation between groups — behave correctly
since. The 0.29.3 "namespace operations routing" fixes appear related but do
not cover this.

## Workaround

Pin all groups to one graph by neutering `FalkorDriver.clone()` (return
`self`), accepting the loss of graph-level isolation; scoping via the
`group_id` property filters already present in every query.

kraft87:
Closes #1331
Closes #1795
Closes #1676

Related: #1651 (MCP `get_episodes` / `delete_episode` bypass group routing in `mcp_server`; separate fix, but the "last-written group" half of that symptom goes away once the shared driver stops being rebound here).

## Summary
Fix a race condition where FalkorDB driver state was mutated in-place when `group_id` differed from the current database, causing concurrent calls to corrupt each other's database context.

Scope after the 2026-07-24 rebase: this PR covers the write path (`add_episode` / `add_episode_bulk`) only. Read-path routing for `get_by_group_ids` landed in #1675.

## Type of Change
- [x] Bug fix
- [x] Documentation/Tests

## Objective
When `add_episode` or `add_episode_bulk` was called with a `group_id` different from the FalkorDB driver's current database, the code mutated shared state:
```python
```
Under concurrent calls (e.g., `asyncio.gather` with different tenants), coroutines would overwrite each other's driver, writing data to the wrong database. On a long-lived multi-group client (the MCP server), every subsequent read also ran inside whichever group wrote last and silently returned empty for other groups (#1795).

### Changes

**Core fix -- driver isolation (`graphiti_core/graphiti.py`):**
- Removed the in-place `self.driver` / `self.clients.driver` mutation from both `add_episode` and `add_episode_bulk`
- Both methods now accept an optional `driver` kwarg, create a local `clients = self.clients.model_copy(update={'driver': driver})`, and thread that through all internal calls
- All internal helpers (`_process_episode_data`, `_extract_and_dedupe_nodes_and_edges`, `_resolve_nodes_and_edges_bulk`, `_get_or_create_saga`, `_saga_get_previous_episode_uuid`, `_extract_and_resolve_edges`) updated to accept and propagate `driver`
- Every `self.driver` / `self.clients` reference in these methods replaced with the local `driver` / `clients`

**Decorator-based group_id handling (`graphiti_core/decorators.py`):**
- Added `@handle_single_group_id` -- detects FalkorDB provider, clones the driver scoped to the group_id, calls `ensure_database_initialized()`, and injects as `driver` kwarg. The original `self.clients.driver` is never touched.
- Updated `@handle_multiple_group_ids` -- now calls `ensure_database_initialized()` on each cloned driver

**FalkorDB driver (`graphiti_core/driver/falkordb_driver.py`):**
- `clone()` now uses `copy.copy(self)` instead of constructing a new `FalkorDriver`, avoiding re-running `__init__` (which schedules index builds, creates new operation objects, etc.)
- Added `_initialized_databases: set[str]` -- shared by reference across all clones via shallow copy
- Added `ensure_database_initialized()` -- idempotent guard that calls `build_indices_and_constraints()` only once per database across all clones

**Base driver (`graphiti_core/driver/driver.py`):**
- Added no-op `ensure_database_initialized()` to `GraphDriver` base class

## Testing
- [x] All existing tests pass

| File | Tests | Type |
|------|-------|------|
| `tests/test_falkor_concurrency.py` | 3 | Unit -- `clone()` is side-effect free, `ensure_database_initialized()` is idempotent, concurrent `add_episode` calls with different `group_id`s never share or mutate a driver |
| `tests/test_handle_multiple_group_ids.py` | updated | Unit -- decorator calls `ensure_database_initialized()` on each cloned driver |

Repro from #1331 / #1676 / #1795 still fails on current `main` without this change and passes with it.

## Checklist
- [x] Self-review completed

### DIFF
--- a/graphiti_core/decorators.py
+++ b/graphiti_core/decorators.py
@@ -61,10 +61,12 @@ def handle_multiple_group_ids(func: F) -> F:
             gid = group_ids[0]
             driver = self.clients.driver
             if gid != getattr(driver, '_database', None):
+                cloned = driver.clone(database=gid)
+                await cloned.ensure_database_initialized()
                 return await func(
                     self,
                     *args,
-                    **{**kwargs, 'driver': driver.clone(database=gid)},
+                    **{**kwargs, 'driver': cloned},
                 )
             return await func(self, *args, **kwargs)
 
@@ -79,10 +81,13 @@ def handle_multiple_group_ids(func: F) -> F:
                 if group_ids_pos is not None and len(args) > group_ids_pos:
                     filtered_args.pop(group_ids_pos)
 
+                cloned = driver.clone(database=gid)
+                await cloned.ensure_database_initialized()
+
                 return await func(
                     self,
                     *filtered_args,
-                    **{**kwargs, 'group_ids': [gid], 'driver': driver.clone(database=gid)},
+                    **{**kwargs, 'group_ids': [gid], 'driver': cloned},
                 )
 
             results = await semaphore_gather(
@@ -111,6 +116,34 @@ def handle_multiple_group_ids(func: F) -> F:
                 return results
 
         # Normal execution
+        return await func(self, *args, **kwargs)
+
+    return wrapper  # type: ignore
+
+
+def handle_single_group_id(func: F) -> F:
+    """
+    Decorator for FalkorDB write methods that need to scope to a single group_id.
+    Injects a cloned driver targeted at the group's graph instead of mutating self.driver.
+    """
+
+    @functools.wraps(func)
+    async def wrapper(self, *args, **kwargs):
+        group_id = kwargs.get('group_id')
+        if group_id is None:
+            pos = get_parameter_position(func, 'group_id')
+            if pos is not None and len(args) > pos - 1:
+                group_id = args[pos - 1]  # adjust for self
+        if (
+            group_id
+            and hasattr(self, 'clients')
+            and hasattr(self.clients, 'driver')
+            and self.clients.driver.provider == GraphProvider.FALKORDB
+            and group_id != self.clients.driver._database
+        ):
+            cloned = self.clients.driver.clone(database=group_id)
+            await cloned.ensure_database_initialized()
+            kwargs['driver'] = cloned
         return await func(self, *args, **kwargs)
 
     return wrapper  # type: ignore
--- a/graphiti_core/driver/driver.py
+++ b/graphiti_core/driver/driver.py
@@ -132,6 +132,14 @@ class GraphDriver(QueryExecutor, ABC):
         """Clone the driver with a different database or graph name."""
         return self
 
+    async def ensure_database_initialized(self) -> None:
+        """Ensure indices and constraints exist for the current database.
+
+        FalkorDB overrides this to track per-graph initialization.
+        Other backends are no-ops since they share a single database.
+        """
+        pass
+
     def build_fulltext_query(
         self, query: str, group_ids: list[str] | None = None, max_query_length: int = 128
     ) -> str:
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 import asyncio
+import copy
 import datetime
 import logging
 import re
@@ -173,12 +174,16 @@ class FalkorDriver(GraphDriver):
         self._search_ops = FalkorSearchOperations()
         self._graph_ops = FalkorGraphMaintenanceOperations()
 
+        # Track which databases have had indices built (shared across clones via shallow copy)
+        self._initialized_databases: set[str] = set()
+
         # Schedule the indices and constraints to be built
         try:
             # Try to get the current event loop
             loop = asyncio.get_running_loop()
             # Schedule the build_indices_and_constraints to run
             loop.create_task(self.build_indices_and_constraints())
+            self._initialized_databases.add(self._database)
         except RuntimeError:
             # No event loop running, this will be handled later
             pass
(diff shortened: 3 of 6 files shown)


danielchalef:

kraft87:

kraft87:
Rebased on current main. The fix now also threads driver through the new _saga_get_previous_episode_uuid helper and saga-tracking calls introduced after the original submission. Added test_falkor_concurrency.py with coverage for clone side-effect-freeness, ensure_database_initialized idempotency, and concurrent add_episode isolation.

Naseem77:
@kraft87 This no longer applies to main, real conflicts. Could you rebase so it can be reviewed properly?

prasmussen15 (changes requested):
Please rebase this branch onto current main and resolve the resulting conflicts before review. The current diff no longer applies cleanly.

kraft87:
@Naseem77 @prasmussen15 Rebased. It's smaller now since #1675 covers the read path; this handles the writes in `add_episode`/`add_episode_bulk`. Repro still fails on current main without it.

prasmussen15 (approved):

kraft87:
@prasmussen15 do the workflows need to be approved as well?

Issue #945, opened 2025-09-28
[BUG]  Custom OPENAI_BASE_URL causes NaN embeddings in MCP server


GarbageGoobler:
## Bug Description
Setting a custom OPENAI_BASE_URL for the MCP server causes embedding generation to return NaN values, breaking memory addition functionality.

## Steps to Reproduce
1. Configure a custom OPENAI_BASE_URL in the `.env` file (e.g., pointing to a different OpenAI-compatible API endpoint)
2. Start the graphiti MCP server using docker-compose
3. Connect Claude Code or another MCP client to the server
4. Attempt to add a memory using the graphiti MCP tools (e.g., `graphiti-memory_add_memory`)

## Expected Behavior
The custom OpenAI-compatible endpoint should return valid numeric embeddings, allowing memories to be successfully added to the graph database with proper vector embeddings for similarity search and retrieval.

## Actual Behavior
The embedding generation returns vectors containing NaN (Not a Number) values, causing the Neo4j vector storage procedure `db.create.setNodeVectorProperty` to fail with "Vector must only contain finite values" error. The memory addition process fails and the episode is not stored.

## Environment
- **Graphiti Version**: Latest (MCP server)
- **Python Version**: 3.x (Docker container)
- **Operating System**: Linux (Docker)
- **Database Backend**: Neo4j 5.26.0
- **LLM Provider & Model**: Custom OpenAI-compatible endpoint (nano-gpt.com in test case)

## Installation Method
- [x] Development installation (git clone + Docker)

## Error Messages/Traceback
```
2025-09-28 22:22:44,147 - __main__ - ERROR - Error processing episode 'Ada Lovelace Biography' for group_id default: {code: Neo.ClientError.Procedure.ProcedureCallFailed} {message: Failed to invoke procedure `db.create.setNodeVectorProperty`: Caused by: java.lang.IllegalArgumentException: Vector must only contain finite values. Provided: SequenceValueVectorCandidate[sequence=List{Double(NaN), Double(NaN), Double(NaN), ... [truncated list of NaN and valid numeric values]]}
```

## Configuration
```python
# .env configuration
OPENAI_BASE_URL=https://nano-gpt.com/api/v1  # Custom endpoint
OPENAI_API_KEY=your_api_key
# Other graphiti MCP server environment variables...
```

## Additional Context
- **Consistency**: Happens consistently when using custom OPENAI_BASE_URL
- **Component**: MCP server (graphiti/mcp_server)
[...]

PR #1332, opened 2026-03-19
Guard against null/invalid embeddings in similarity search queries

### LINKED ISSUE #1328: [BUG] `edge_similarity_search` crashes on edges with null/invalid `fact_embedding`
# Graphiti Bug Report: `edge_similarity_search` crashes on edges with null/invalid `fact_embedding`

## Bug Description

When a Neo4j graph grows over time, `edge_similarity_search` fails with `Neo.ClientError.Statement.ArgumentError` because some `RELATES_TO` edges have a `null`, empty, or dimension-mismatched `fact_embedding` property. The `vector.similarity.cosine()` function in Neo4j does not accept `null` or invalid vectors, causing the entire search query to fail.

The root cause is in `Neo4jSearchOperations.edge_similarity_search` (`graphiti_core/driver/neo4j/operations/search_ops.py`), which builds a Cypher query that computes cosine similarity on **all** matched edges without first filtering out edges where `fact_embedding` is `null` or invalid:

```cypher
MATCH (n:Entity)-[e:RELATES_TO]->(m:Entity)
WITH DISTINCT e, n, m, vector.similarity.cosine(e.fact_embedding, $search_vector) AS score
WHERE score > $min_score
...
```

The same issue likely affects `node_similarity_search` (with `n.name_embedding`) and `community_similarity_search` (with `c.name_embedding`).

## Steps to Reproduce

```python

# Initialize Graphiti with Neo4j and a custom embedder
client = Graphiti(
    uri="bolt://localhost:7687",
    user="neo4j",
    password="password",
    llm_client=my_llm_adapter,
    embedder=my_embedder_adapter,  # e.g., 1024-dim embeddings
)

# Add multiple episodes over time — some embedding calls may silently fail
# or edges may be created with null/mismatched fact_embedding
for episode in episodes:
    await client.add_episode(
        name=episode.name,
        episode_body=episode.body,
        source_description=episode.description,
        reference_time=episode.timestamp,
        group_id="user_123",
    )

# Eventually, search fails when it encounters an edge with invalid embedding
    query="list today's emails",
    group_ids=["user_123"],
    limit=50,
)
# ^ Raises Neo.ClientError.Statement.ArgumentError
```

## Expected Behavior

The search query should gracefully skip edges (and nodes) where the embedding property is `null` or invalid, and return results only from edges with valid embeddings. A `WHERE e.fact_embedding IS NOT NULL` clause (or equivalent) should be added before the `vector.similarity.cosine()` call.

## Actual Behavior

The entire search query fails with:

```
neo4j.exceptions.ClientError: {neo4j_code: Neo.ClientError.Statement.ArgumentError}
{message: Invalid input for 'vector.similarity.cosine()': Argument a is not a valid vector for this similarity function.}
```

This causes complete search failure — no results are returned, even though most edges have valid embeddings.

## Environment

- **Graphiti Version**: 0.28.0 (also verified code is unchanged in 0.28.2)
- **Python Version**: 3.12
- **Operating System**: Linux (Docker container)
- **Database Backend**: Neo4j 5.x (via `neo4j==6.1.0` Python driver)
- **LLM Provider & Model**: Custom LLM adapter (T-Systems OTC / `text-embedding-bge-m3`, 1024 dimensions)

## Installation Method

- [x] pip install

## Error Messages/Traceback

```
ERROR: Error executing Neo4j query: {neo4j_code: Neo.ClientError.Statement.ArgumentError} {message: Invalid input for 'vector.similarity.cosine()': Argument a is not a valid vector for this similarity function.} {gql_status: 22N38} {gql_status_description: error: data exception - invalid function argument. Invalid argument to the function vector.similarity.cosine.}
        MATCH (n:Entity)-[e:RELATES_TO]->(m:Entity)
            WITH DISTINCT e, n, m, vector.similarity.cosine(e.fact_embedding, $search_vector) AS score
            WHERE score > $min_score
            RETURN
            e.uuid AS uuid,
            n.uuid AS source_node_uuid,
            m.uuid AS target_node_uuid,
            ...
            ORDER BY score DESC
            LIMIT $limit

  File ".../graphiti_core/search/search_utils.py", line 431, in edge_similarity_search
  ...
  File ".../neo4j/_async/io/_common.py", line 263, in on_failure
    raise self._hydrate_error(metadata)
neo4j.exceptions.ClientError: {neo4j_code: Neo.ClientError.Statement.ArgumentError}
{message: Invalid input for 'vector.similarity.cosine()': Argument a is not a valid vector for this similarity function.}
```

## Configuration

```python
# Custom embedder producing 1024-dim vectors (bge-m3)
# Graphiti initialized with:
client = Graphiti(
    uri="bolt://neo4j:7687",
    user="neo4j",
    password="***",
    llm_client=amas_llm_adapter,
    embedder=amas_embedder_adapter,  # 1024-dim
)
```

## Additional Context

- **Consistency**: Happens consistently once the graph has enough edges — as the graph grows, the probability of encountering an edge with a null/invalid `fact_embedding` increases. Small graphs may work fine.
- **Component**: Core library (`graphiti-core`)
- **Root cause**: Edges can end up with `null` or invalid `fact_embedding` when:
  1. The embedding API call fails/times out during `add_episode()` but the edge itself is still created in Neo4j
  2. A different embedding model with different dimensions was used at some point (dimension mismatch)
  3. The embedding returns NaN values (related: [#945](https://github.com/getzep/graphiti/issues/945))
- **Related issues**: [#945 - Custom OPENAI_BASE_URL causes NaN embeddings](https://github.com/getzep/graphiti/issues/945), [neo4j/graph-data-science#26 - Cosine similarity with null value](https://github.com/neo4j/graph-data-science/issues/26)
- **Scope**: Affects `edge_similarity_search`, and potentially `node_similarity_search` and `community_similarity_search` in the same file

## Possible Solution

Add a `WHERE` clause to filter out edges/nodes with null embeddings **before** computing cosine similarity. For `edge_similarity_search` in `neo4j/operations/search_ops.py`:

```cypher
MATCH (n:Entity)-[e:RELATES_TO]->(m:Entity)
  AND e.fact_embedding IS NOT NULL
  AND size(e.fact_embedding) = size($search_vector)
WITH DISTINCT e, n, m, vector.similarity.cosine(e.fact_embedding, $search_vector) AS score
WHERE score > $min_score
...
```

The fix should be applied to all three similarity search methods:
1. `edge_similarity_search` → guard `e.fact_embedding`
2. `node_similarity_search` → guard `n.name_embedding`
3. `community_similarity_search` → guard `c.name_embedding`

Additionally, consider adding a data validation/cleanup utility to detect and optionally re-embed or remove edges/nodes with invalid embeddings.

pratyush618:
## Summary

Fixes #1328

Edges and nodes can end up with `null` or dimension-mismatched embeddings in the database when:
- The embedding API call fails or times out during `add_episode()` but the edge/node is still written
- A different embedding model with different dimensions was used at some point
- The embedding returns NaN values (related to #945)

In these cases, calling `vector.similarity.cosine()` throws `Neo.ClientError.Statement.ArgumentError`, failing the entire search query.

This PR adds `IS NOT NULL` and dimension-size guard clauses **before** the cosine similarity call in all three affected methods, for both Neo4j and FalkorDB backends:

- `node_similarity_search` — guards `n.name_embedding`
- `edge_similarity_search` — guards `e.fact_embedding`
- `community_similarity_search` — guards `c.name_embedding`

The generated Cypher now looks like:
```cypher
WHERE e.fact_embedding IS NOT NULL
  AND size(e.fact_embedding) = size($search_vector)
```

Invalid records are silently skipped and valid results are still returned, rather than the entire query crashing.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -188,6 +188,9 @@ class FalkorSearchOperations(SearchOperations):
             filter_queries.append('n.group_id IN $group_ids')
             filter_params['group_ids'] = group_ids
 
+        filter_queries.append('n.name_embedding IS NOT NULL')
+        filter_queries.append('size(n.name_embedding) = size($search_vector)')
+
         filter_query = ''
         if filter_queries:
             filter_query = ' WHERE ' + (' AND '.join(filter_queries))
@@ -352,6 +355,9 @@ class FalkorSearchOperations(SearchOperations):
             if target_node_uuid is not None:
                 filter_params['target_uuid'] = target_node_uuid
                 filter_queries.append('m.uuid = $target_uuid')
+
+        filter_queries.append('e.fact_embedding IS NOT NULL')
+        filter_queries.append('size(e.fact_embedding) = size($search_vector)')
 
         filter_query = ''
         if filter_queries:
@@ -537,14 +543,19 @@ class FalkorSearchOperations(SearchOperations):
     ) -> list[CommunityNode]:
         query_params: dict[str, Any] = {}
 
-        group_filter_query = ''
+        filter_clauses = [
+            'c.name_embedding IS NOT NULL',
+            'size(c.name_embedding) = size($search_vector)',
+        ]
         if group_ids is not None:
-            group_filter_query += ' WHERE c.group_id IN $group_ids'
+            filter_clauses.append('c.group_id IN $group_ids')
             query_params['group_ids'] = group_ids
+
+        filter_query = ' WHERE ' + ' AND '.join(filter_clauses)
 
         cypher = (
             'MATCH (c:Community)'
-            + group_filter_query
+            + filter_query
             + """
             WITH c,
             """
--- a/graphiti_core/driver/neo4j/operations/search_ops.py
+++ b/graphiti_core/driver/neo4j/operations/search_ops.py
@@ -142,6 +142,9 @@ class Neo4jSearchOperations(SearchOperations):
             filter_queries.append('n.group_id IN $group_ids')
             filter_params['group_ids'] = group_ids
 
+        filter_queries.append('n.name_embedding IS NOT NULL')
+        filter_queries.append('size(n.name_embedding) = size($search_vector)')
+
         filter_query = ''
         if filter_queries:
             filter_query = ' WHERE ' + (' AND '.join(filter_queries))
@@ -307,6 +310,9 @@ class Neo4jSearchOperations(SearchOperations):
             if target_node_uuid is not None:
                 filter_params['target_uuid'] = target_node_uuid
                 filter_queries.append('m.uuid = $target_uuid')
+
+        filter_queries.append('e.fact_embedding IS NOT NULL')
+        filter_queries.append('size(e.fact_embedding) = size($search_vector)')
 
         filter_query = ''
         if filter_queries:
@@ -490,14 +496,19 @@ class Neo4jSearchOperations(SearchOperations):
     ) -> list[CommunityNode]:
         query_params: dict[str, Any] = {}
 
-        group_filter_query = ''
+        filter_clauses = [
+            'c.name_embedding IS NOT NULL',
+            'size(c.name_embedding) = size($search_vector)',
+        ]
         if group_ids is not None:
-            group_filter_query += ' WHERE c.group_id IN $group_ids'
+            filter_clauses.append('c.group_id IN $group_ids')
             query_params['group_ids'] = group_ids
+
+        filter_query = ' WHERE ' + ' AND '.join(filter_clauses)
 
         cypher = (
             'MATCH (c:Community)'
-            + group_filter_query
+            + filter_query
             + """
             WITH c,
             """


danielchalef:

pratyush618:

themavik (commented):
good defensive change — null embeddings hitting cosine distance would crash the query. the size() check is smart too since mismatched dimensions would give garbage results. one question: does this silently filter out nodes that haven't been embedded yet? if a node is newly created and hasn't gone through the embedding pipeline, it'd be invisible to all similarity searches. might want an index-time invariant instead (reject inserts without embeddings) vs query-time filtering.

pratyush618 (commented):
Thanks for the review! Good question — yes, this will filter out nodes/edges that haven't been embedded yet from similarity searches. But I think that's the correct behavior here since `vector.similarity.cosine()` simply can't produce a meaningful score without a valid embedding vector.

These records would still be discoverable through the other search paths (fulltext search, BFS traversal) — they're only invisible to similarity-based retrieval, which inherently requires a valid embedding to function.

An index-time invariant (rejecting writes without embeddings) is an interesting idea, but it could be worse in practice — if the embedding API is temporarily down or times out, you'd lose the data entirely rather than just having it temporarily excluded from one search method. The current approach is more resilient: the data is persisted, and once re-embedded (e.g. via a cleanup/backfill utility), it becomes searchable again.

Naseem77:
@pratyush618 Tested after applying: the public node_similarity_search still crashes with a dimension mismatch. Only the unused driver-ops paths are guarded, the live path is search_utils.py.

prasmussen15 (changes requested):
Please move the embedding validation to the live public similarity-search path in graphiti_core/search/search_utils.py. The currently guarded driver-ops code is not the path FalkorDB executes.

PR #1335, opened 2026-03-20
feat(falkordb): Use HNSW vector index procedures for similarity search

### LINKED ISSUE #1055: [Question] Why vector index is not used in similarity search?
In similarity search functions, `vector.similarity.cosine` is used intead of `db.index.vector.queryNodes/queryRelationships`? This change is made in https://github.com/getzep/graphiti/pull/157/.

What is the reason for this?

rhlsthrm:
## Problem

All FalkorDB similarity searches (`node_similarity_search`, `edge_similarity_search`, `community_similarity_search`) use brute-force `vec.cosineDistance()` which performs a full scan of every node/edge — O(n) with high-dimensional vectors (e.g. 3072-dim). On large graphs this takes 6-8 seconds per query.

FalkorDB already supports HNSW vector indexes and provides `db.idx.vector.queryNodes` and `db.idx.vector.queryRelationships` procedures that leverage these indexes for approximate nearest neighbor search.

## Solution

Replace brute-force cosine distance scans with HNSW vector index procedures:

- **`node_similarity_search`**: Uses `db.idx.vector.queryNodes("Entity", "name_embedding", ...)` 
- **`edge_similarity_search`**: Uses `db.idx.vector.queryRelationships("RELATES_TO", "fact_embedding", ...)`
- **`community_similarity_search`**: Uses `db.idx.vector.queryNodes("Community", "name_embedding", ...)`

### Key implementation details:

1. **Score conversion**: HNSW procedures return cosine *distance* (0=identical). Convert to similarity: `similarity = 1.0 - distance`
2. **Post-filtering**: HNSW index doesn't support pre-filtering, so we over-fetch candidates (`limit * 3`) and apply group_id/search_filter/source/target UUID filters after retrieval
3. **Graceful fallback**: If the vector index doesn't exist, falls back to the original brute-force approach with a warning log
4. **Vector index creation**: Added `get_vector_indices()` to `graph_queries.py` and updated `FalkorDriver.build_indices_and_constraints()` to create vector indices on startup

## Performance Impact

HNSW index queries return results in ~0.002s vs ~6-8s for brute-force scans — approximately **3000x speedup** on the vector query portion for large graphs.

## Files Changed

- `graphiti_core/driver/falkordb/operations/search_ops.py` — Main changes: 3 search methods use HNSW, 3 brute-force fallback methods added
- `graphiti_core/graph_queries.py` — Added `get_vector_indices()` function
- `graphiti_core/driver/falkordb_driver.py` — Creates vector indices on startup

Closes #1055

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -190,6 +190,66 @@ class FalkorSearchOperations(SearchOperations):
 
         filter_query = ''
         if filter_queries:
+            filter_query = ' AND ' + (' AND '.join(filter_queries))
+
+        # Use HNSW vector index for fast approximate nearest neighbor search
+        # Over-fetch candidates to allow for post-filtering
+        candidate_limit = limit * 3
+
+        # score from db.idx.vector.queryNodes is cosine DISTANCE (0=identical, 1=orthogonal)
+        # Convert to similarity: similarity = 1.0 - distance
+        cypher = f"""
+            CALL db.idx.vector.queryNodes('Entity', 'name_embedding', $candidate_limit, vecf32($search_vector))
+            YIELD node AS n, score
+            WITH n, (1.0 - score) AS similarity
+            WHERE similarity > $min_score
+            {filter_query}
+            WITH n, similarity
+            ORDER BY similarity DESC
+            LIMIT $limit
+            RETURN
+            {get_entity_node_return_query(GraphProvider.FALKORDB)}
+        """
+
+        try:
+            records, _, _ = await executor.execute_query(
+                cypher,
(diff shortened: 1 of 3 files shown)


danielchalef:

rhlsthrm:

Naseem77:
@rhlsthrm Nice idea, but current FalkorDB rejects CALL db.idx.vector.createNodeIndex (procedure not registered). Switching to the CREATE VECTOR INDEX ... OPTIONS syntax would make this work.

prasmussen15 (changes requested):
Please replace the unavailable db.idx.vector.createNodeIndex procedure with current FalkorDB CREATE VECTOR INDEX ... OPTIONS syntax, and add a regression test against that supported syntax.

PR #1395, opened 2026-04-11
fix: replace 'minimal' with 'low' reasoning effort for gpt-5.4-mini

sgaluza:
## Summary

`gpt-5.4-mini` does not support `reasoning.effort: "minimal"`. The supported values are: `none`, `low`, `medium`, `high`, `xhigh`.

This PR:
- Updates `DEFAULT_REASONING` from `'minimal'` to `'low'` in `openai_base_client.py`
- Adds `VALID_REASONING_VALUES` and `VALID_VERBOSITY_VALUES` constants with early validation in `BaseOpenAIClient` constructor — invalid values now raise `ValueError` immediately instead of failing at API call time
- Updates existing tests to use valid reasoning values (`'minimal'` → `'low'`, `'intense'` → `'high'`)
- Adds parametrized tests for validation of both `reasoning` and `verbosity` parameters

## Motivation

When using `gpt-5.4-mini` as the extraction model, batch operations fail with:

```
Unsupported value: 'minimal' is not supported with the 'gpt-5.4-mini' model.
Supported values are: 'none', 'low', 'medium', 'high', and 'xhigh'.
```

`'low'` was chosen over `'none'` because edge extraction and node deduplication require minimal reasoning for quality results.

## Test plan

- [x] All existing tests pass with updated values
- [x] New parametrized tests verify validation rejects invalid values
- [x] Tested in production with gpt-5.4-mini batch sync (78 documents)

### DIFF
--- a/graphiti_core/llm_client/openai_base_client.py
+++ b/graphiti_core/llm_client/openai_base_client.py
@@ -33,8 +33,14 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_MODEL = 'gpt-4.1-mini'
 DEFAULT_SMALL_MODEL = 'gpt-4.1-nano'
-DEFAULT_REASONING = 'minimal'
+DEFAULT_REASONING = 'low'
 DEFAULT_VERBOSITY = 'low'
+VALID_REASONING_VALUES: frozenset[str | None] = frozenset(
+    {None, 'none', 'low', 'medium', 'high', 'xhigh'}
+)
+VALID_VERBOSITY_VALUES: frozenset[str | None] = frozenset(
+    {None, 'low', 'medium', 'high'}
+)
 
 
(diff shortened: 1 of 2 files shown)


danielchalef:
Thanks for flagging the `minimal`-vs-`low` reasoning-effort issue. The reasoning-effort handling was reworked in **#1551** (merged): rather than a hardcoded cross-model value, the default effort is now **model-tied** — `gpt-5.5` uses `'none'`, and the stale `'minimal'` hardcode (which `gpt-5.4-mini` rejected) is gone. Because the lowest valid effort differs by snapshot, it's resolved per-model with an explicit `reasoning=` override always available, and temperature is never sent to reasoning models. That covers the `gpt-5.4-mini` case this PR targeted.

Closing as **superseded by #1551** — appreciate the contribution!

PR #1414, opened 2026-04-14
fix(mcp): fail loudly when embedder initialization fails

chrisgscott:
## Summary

- Embedder initialization failure is now a hard error instead of a warning that silently continues with `embedder=None`
- Added search health check at startup that verifies the embedder can produce vectors
- Added defensive checks in `search_nodes` and `search_memory_facts` that return a clear error if embedder is unexpectedly `None`

## Problem

When the embedder client fails to initialize (invalid API key, network issue, etc.), the MCP server logs a warning and continues with `embedder_client = None`. This causes `search_nodes` and `search_memory_facts` to silently return empty results — the hybrid search (`NODE_HYBRID_SEARCH_RRF`) needs the embedder to generate query vectors for the cosine similarity leg, and without it, RRF reranking produces nothing.

The failure mode is especially confusing because `add_memory` continues to work normally (it uses the LLM client, not the embedder). Users see data being written successfully but get empty results on every search, with no error indicating why.

## Test plan

- [ ] Start the MCP server with a valid embedder config — verify "Search health check passed" in logs
- [ ] Start the MCP server with an invalid/missing API key — verify it fails at startup with a clear error instead of silently degrading
- [ ] Call `search_nodes` when embedder is available — verify results are returned normally
- [ ] If embedder is somehow `None` at search time, verify the tool returns an explicit error message instead of empty results

### DIFF
--- a/graphiti_core/llm_client/openai_client.py
+++ b/graphiti_core/llm_client/openai_client.py
@@ -90,8 +90,11 @@ class OpenAIClient(BaseOpenAIClient):
             request_kwargs['temperature'] = temperature_value
 
         # Only include reasoning and verbosity parameters for reasoning models
+        # Map legacy reasoning effort values to current OpenAI API values
+        _REASONING_MAP = {'minimal': 'low', 'default': 'medium', 'maximum': 'high'}
         if is_reasoning_model and reasoning is not None:
-            request_kwargs['reasoning'] = {'effort': reasoning}  # type: ignore
+            effort = _REASONING_MAP.get(reasoning, reasoning)
+            request_kwargs['reasoning'] = {'effort': effort}  # type: ignore
 
         if is_reasoning_model and verbosity is not None:
             request_kwargs['text'] = {'verbosity': verbosity}  # type: ignore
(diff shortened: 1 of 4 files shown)


danielchalef:
   - - -
   - - -

xkonjin (commented):
Good iterative improvements here. Some concerns before merging:

1. **Critical: Silent fallback on embedder health check.** In `initialize()`, if `embedder_client.create(['health check'])` throws, you catch it and log a warning that "Search may not work correctly." But `embedder_client` was already created successfully above — this branch means the *runtime* embedder call failed (network, quota, auth). That is a real failure mode. Logging a warning and continuing means users will see confusing search errors later instead of a clear startup failure. Consider whether this should be a hard error, or at minimum expose a /health endpoint that reports it so operators can detect it.

2. **Model string assumptions.** `model.startswith('gpt-5')` will match any future `gpt-5.x` model, which is okay today, but it is fragile if OpenAI releases a non-reasoning `gpt-5` variant or changes naming. There is no action needed now, but consider centralizing reasoning-model detection into a shared helper so you don't have to update every `startswith` call later.

3. **Missing tests for token param mapping.** `_create_completion` now switches between `max_tokens` and `max_completion_tokens` based on `is_reasoning_model`. Add a unit test that asserts the correct kwarg is sent for both reasoning and non-reasoning model strings.

4. **Retry loop doubles imports.** You import `asyncio as _asyncio` inside the retry loop body. Move that to the top of the file with the other imports.

5. **Config YAMLs contain hardcoded user.** `config-second-brain.yaml` and `config-second-brain-stdio.yaml` both set `user_id: chris`. If these are committed examples, they should either be parameterized (`${USER_ID:chris}` already exists in the same file) or the hardcoded default should be something neutral like `default_user`.

Fix the import placement and decide on the embedder health-check failure semantics before merging.

PR #1421, opened 2026-04-18
fix: correctly flatten update_community results in add_episode

### LINKED ISSUE #1396: add_episode tuple-unpack fails when extracted node count is not exactly 2
# `add_episode` crashes with `ValueError: too many / not enough values to unpack` when extracted node count is not exactly 2

## Summary

In `graphiti_core/graphiti.py` line 992, the `update_communities`
branch of `add_episode()` does:

```python
communities, community_edges = await semaphore_gather(
    *[
        update_community(self.driver, self.llm_client, self.embedder, node)
        for node in nodes
    ],
    max_coroutines=self.max_coroutines,
)
```

`semaphore_gather` returns a list of N tuples (one per coroutine),
where each tuple is `(list[CommunityNode], list[CommunityEdge])`
(from `update_community`'s return type in
`graphiti_core/utils/maintenance/community_operations.py:326`). The
destructuring `communities, community_edges = <list of N tuples>`
treats the list as an iterable of length 2, so:

- **N == 2** → lucky coincidence: `communities` becomes the first
  tuple and `community_edges` becomes the second. The types are
  now wrong (both are tuples, not lists) but Python doesn't notice
  until a downstream consumer uses them.
- **N == 1** → `ValueError: not enough values to unpack (expected 2, got 1)`
- **N == 3+** → `ValueError: too many values to unpack (expected 2)`

Any episode that happens to extract a number of entity nodes other
than exactly 2 will crash the whole `add_episode` call when
`update_communities=True` (the default).

## Environment

- graphiti-core 0.28.2 (PyPI)
- Python 3.12
- FalkorDB driver
- Any LLM backend

## Reproducer

```python
import asyncio

async def main():
    g = Graphiti(
        uri='falkor://localhost:6379',
        # ... your credentials ...
    )
    await g.build_indices_and_constraints()

    # Craft an episode that extracts exactly 1 entity — e.g. a
    # short sentence that only mentions one person:
    await g.add_episode(
        name='solo',
        episode_body='Alice went to the store.',
        source=EpisodeType.text,
        source_description='test',
        group_id='test',
        # update_communities defaults to True
    )

asyncio.run(main())
```

Expected: episode added cleanly.

Observed:
```
ValueError: not enough values to unpack (expected 2, got 1)
  File ".../graphiti_core/graphiti.py", line 992, in add_episode
    communities, community_edges = await semaphore_gather(
```

Try again with an episode that extracts 3 entities
("Alice, Bob, and Carol went to the store together.") — same crash,
different message: `too many values to unpack (expected 2)`.

## Root cause

`update_community` in `community_operations.py:326` returns
`tuple[list[CommunityNode], list[CommunityEdge]]` — two parallel
lists **per node**. `semaphore_gather` returns a list of these tuples,
one per coroutine. The caller needs to flatten them, not unpack them
as if they were two top-level parallel lists.

## Suggested fix

```python
# graphiti.py:992
if update_communities:
    per_node_results = await semaphore_gather(
        *[
            update_community(self.driver, self.llm_client, self.embedder, node)
            for node in nodes
        ],
        max_coroutines=self.max_coroutines,
    )
    communities: list[CommunityNode] = []
    community_edges: list[CommunityEdge] = []
    for node_communities, node_community_edges in per_node_results:
        communities.extend(node_communities)
        community_edges.extend(node_community_edges)
```

This correctly accumulates the lists across all nodes regardless of
how many were extracted from the episode.

## Our workaround

We set `update_communities=False` on every `add_episode` call from
our application code (`lilly/graphiti_writer.py` and
`services/graphiti-mcp`). This means we don't benefit from
incremental community updates on ingest, but we work around the
crash. We rebuild communities in a nightly `build_communities()`
task instead (which has its own issues — see issue #3 and #4 in
this series).

## Impact

Any caller who uses the default `update_communities=True` and whose
episodes extract any number of entities other than exactly 2 will
hit this. It's surprising that the bug survives because lots of
two-participant conversational episodes happen to extract exactly
2 entities and work by accident.

## Labels suggestion

`bug`, `add_episode`, `community`

octo-patch:
Fixes #1396

## Problem

`add_episode` crashes with `ValueError: too many / not enough values to unpack` when the number of extracted entity nodes is anything other than exactly 2.

The root cause: `semaphore_gather` returns a **list of N tuples** (one per coroutine), where each tuple is `(list[CommunityNode], list[CommunityEdge])`. The previous code did:

```python
communities, community_edges = await semaphore_gather(...)
```

This Python unpacking treats the returned list as if it has exactly 2 items, so it only worked by accident when `N == 2`. For `N == 1` it raises `ValueError: not enough values to unpack`; for `N >= 3` it raises `ValueError: too many values to unpack`.

## Solution

Replace the tuple unpack with an explicit accumulation loop:

```python
per_node_results = await semaphore_gather(...)
for node_communities, node_community_edges in per_node_results:
    communities.extend(node_communities)
    community_edges.extend(node_community_edges)
```

This correctly aggregates results across all nodes regardless of how many entity nodes were extracted from the episode. Also adds explicit type annotations to `communities` and `community_edges` since they were previously inferred as `list[Any]`.

## Testing

The fix is a straightforward logic correction with no behaviour change for callers — the returned `communities` and `community_edges` lists now have the correct types and values for all N. No external dependencies are required to verify the logic.

### DIFF
--- a/graphiti_core/graphiti.py
+++ b/graphiti_core/graphiti.py
@@ -1114,16 +1114,19 @@ class Graphiti:
                 )
 
                 # Update communities if requested
-                communities = []
-                community_edges = []
+                communities: list[CommunityNode] = []
+                community_edges: list[CommunityEdge] = []
                 if update_communities:
-                    communities, community_edges = await semaphore_gather(
+                    per_node_results = await semaphore_gather(
                         *[
                             update_community(self.driver, self.llm_client, self.embedder, node)
                             for node in nodes
                         ],
                         max_coroutines=self.max_coroutines,
                     )
+                    for node_communities, node_community_edges in per_node_results:
+                        communities.extend(node_communities)
+                        community_edges.extend(node_community_edges)
 
                 end = time()
 


danielchalef:
Thanks for this — correct fix for the `update_community` results not being flattened from `semaphore_gather` in `add_episode` (`graphiti.py:1185`). Consolidating on #1486 as the canonical fix (same fix + a regression test, rebased on main). Closing as superseded by #1486; credit to you for surfacing it too.

PR #1450, opened 2026-04-28
fix(falkordb): strip backticks and guard empty token list in fulltext query builder

### LINKED ISSUE #1440: [BUG] FalkorDB fulltext query builder produces malformed RediSearch queries on backtick tokens and stopword-only inputs
## Bug Description

`_build_falkor_fulltext_query` in `graphiti_core` produces malformed RediSearch queries in two distinct cases, both surfacing during normal dedup-search operations:

1. **Backtick tokens leak through to RediSearch.** When an extracted entity name contains backticks (common when an LLM names entities from code-fenced text such as `` `add_episode` ``), the separator translation map omits backtick, so the raw character survives sanitization and breaks RediSearch parsing.

2. **Empty paren on stopword-only or punctuation-only queries.** When the stopword filter removes every token (or the input is short / all-separators), the builder still constructs a query like `(@group_id:"…") ()`, which RediSearch rejects with `Syntax error at offset N`. There is no early-return guard for the "no usable tokens" case.

Both failure modes occur in two upstream copies of the function and both must be fixed:

- `graphiti_core.driver.falkordb.operations.search_ops._build_falkor_fulltext_query`
  (called from `FalkorSearchOperations.{node,edge,episode,community}_fulltext_search`)
- `graphiti_core.driver.falkordb_driver.FalkorDriver.build_fulltext_query`
  (called from `graphiti_core.search.search_utils` via `driver.build_fulltext_query`)

## Steps to Reproduce

Phase-1 of a corpus rerender on FalkorDB using Sonnet 4.6 as the extractor. With richer entity names the failure rate jumped from 0/55 episodes (under Haiku 4.5) to 16/55 episodes (under Sonnet 4.6) — Sonnet's more descriptive entity names trigger backtick / stopword edge cases more often.

Symptoms in logs:

```
ResponseError: Syntax error at offset N near …
```

Either issued by RediSearch directly, or surfaced via Graphiti as a failed dedup-search step that aborts ingestion of the episode.

## Expected Behavior

The query builder should:

- Strip backticks from the input the same way it strips other RediSearch special characters (`,.<>{}[]"':;!@#$%^&*()-+=~?|/\`).
- Return an empty string (or otherwise short-circuit) when no usable tokens remain after sanitization + stopword filtering, so callers can skip the search rather than hit RediSearch with a malformed query.

## Actual Behavior

- Backticks survive sanitization and reach RediSearch, breaking parsing.
- Empty filtered token list produces `(@group_id:"…") ()`, which RediSearch rejects.

## Environment

- **Graphiti Version**: 0.28.2
- **Python Version**: 3.12
- **Operating System**: Windows 11 host (FalkorDB running in Docker)
- **Database Backend**: FalkorDB edge
- **LLM Provider & Model**: Anthropic `claude-sonnet-4-6` and `claude-haiku-4-5`

## Installation Method

- [x] pip install (`graphiti-core==0.28.2`)

## Error Messages/Traceback

Per-episode failure in a corpus rerender; dedup-search aborts the episode after retry exhaustion. Specific raw error from FalkorDB:

```
ResponseError: Syntax error at offset N near `<character>`
```

(Both backtick-leak and empty-paren cases produce a `ResponseError` with offsets pointing at the offending construct.)

## Configuration

Standard `Graphiti(falkor_driver, llm_client, embedder)` setup. Failure is independent of any schema configuration — it triggers on the raw entity-name search query path inside dedup-search.

## Additional Context

- **Consistency**: deterministic given the input. Any episode whose extracted entity names contain backticks fails; any episode whose dedup-search query reduces to stopwords-only fails.
- **Component**: `graphiti_core` library (not REST/MCP-specific).
- **Scaling**: the failure rate increases with extractor verbosity. Same corpus that runs 55/55 with Haiku-named entities runs 39/55 with Sonnet-named entities.
- **Related issues**: #1269 reports a different RediSearch syntax error around hyphens inside quoted `group_ids`; that is a separate code path and is not addressed here. PR #1175 quoted group_ids; this issue is about the *query text* sanitization.

## Possible Solution

Two surgical changes to both copies of `_build_falkor_fulltext_query`:

1. Add backtick (`` ` ``) to the separator translation map so it is replaced with whitespace before tokenization.
2. After stopword filtering, early-return an empty string (or a sentinel that callers handle) when `filtered_words` is empty, so no query is issued.

Validated locally by monkey-patching both function copies (and the driver-level method) with these two fixes — corpus rerender failure rate drops from 27 % (16/55 + 8/20) to 0 % (35/35 + 20/20) on the same Sonnet 4.6 + FalkorDB combination.

Happy to follow up with a PR once an approach is acked.

octo-patch:
Fixes #1440

## Problem

`_build_falkor_fulltext_query` (and `FalkorDriver.build_fulltext_query`) had two bugs that caused RediSearch `Syntax error` failures during dedup-search:

1. **Backtick tokens leak through to RediSearch.** The separator translation map omitted the backtick character `` ` ``, so entity names from code-fenced text (e.g. `` `add_episode` ``) survived sanitization and broke RediSearch parsing.

2. **Empty parenthesis on stopword-only or punctuation-only queries.** When every token was filtered out as a stopword (or the input was all separators/backticks), `filtered_words` became empty but the builder still constructed `(@group_id:"…") ()`, which RediSearch rejects with a syntax error.

Both failure modes existed in two upstream copies of the function:
- `graphiti_core/driver/falkordb/operations/search_ops.py` — `_build_falkor_fulltext_query`
- `graphiti_core/driver/falkordb_driver.py` — `FalkorDriver.sanitize` / `build_fulltext_query`

## Solution

Two surgical changes applied to both copies:

1. Add `` ` `` to the separator translation map so backticks are replaced with whitespace before tokenization.
2. Return `''` early when `filtered_words` is empty after stopword filtering, so callers skip the RediSearch query rather than issuing a malformed one.

## Testing

Added unit tests in `tests/driver/test_falkordb_driver.py` covering:
- Backtick tokens are stripped from the query string
- Stopword-only queries return `''`
- Punctuation-only queries return `''`
- Backtick-only queries return `''`
- Normal queries still produce valid non-empty results
- Backtick mixed with normal tokens strips backticks but keeps the tokens

All 6 new tests pass. No existing tests were broken.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -83,6 +83,7 @@ _SEPARATOR_MAP = str.maketrans(
         '|': ' ',
         '/': ' ',
         '\\': ' ',
+        '`': ' ',
     }
 )
 
@@ -111,6 +112,12 @@ def _build_falkor_fulltext_query(
     # Remove stopwords and empty tokens
     query_words = sanitized_query.split()
     filtered_words = [word for word in query_words if word and word.lower() not in STOPWORDS]
+
+    # If all tokens were stopwords or the query was empty/punctuation-only, return ''
+    # so callers skip the search rather than issue a malformed "(@group_id:…) ()" query.
+    if not filtered_words:
+        return ''
+
     sanitized_query = ' | '.join(filtered_words)
 
     if len(sanitized_query.split(' ')) + len(group_ids or '') >= max_query_length:
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -379,6 +379,7 @@ class FalkorDriver(GraphDriver):
                 '|': ' ',
                 '/': ' ',
                 '\\': ' ',
+                '`': ' ',
             }
         )
         sanitized = query.translate(separator_map)
@@ -414,6 +415,12 @@ class FalkorDriver(GraphDriver):
         # Remove stopwords and empty tokens from the sanitized query
         query_words = sanitized_query.split()
         filtered_words = [word for word in query_words if word and word.lower() not in STOPWORDS]
+
+        # If all tokens were stopwords or the query was empty/punctuation-only, return ''
+        # so callers skip the search rather than issue a malformed "(@group_id:…) ()" query.
+        if not filtered_words:
+            return ''
+
         sanitized_query = ' | '.join(filtered_words)
 
         # If the query is too long return no query
--- a/tests/driver/test_falkordb_driver.py
+++ b/tests/driver/test_falkordb_driver.py
@@ -362,6 +362,55 @@ class TestDatetimeConversion:
         assert convert_datetimes_to_strings(True) is True
 
 
+class TestFalkorDriverFulltextQuery:
+    """Tests for FalkorDriver.build_fulltext_query and _build_falkor_fulltext_query."""
+
+    @unittest.skipIf(not HAS_FALKORDB, 'FalkorDB is not installed')
+    def setup_method(self):
+        with patch('graphiti_core.driver.falkordb_driver.FalkorDB'):
+            self.driver = FalkorDriver()
+
+    @unittest.skipIf(not HAS_FALKORDB, 'FalkorDB is not installed')
+    def test_backtick_tokens_are_stripped(self):
+        """Backtick-quoted tokens (e.g. `add_episode`) must not reach RediSearch."""
+        result = self.driver.build_fulltext_query('`add_episode`')
+        # backtick stripped → token is "add_episode" → underscore survives, no raw backtick
+        assert '`' not in result
+
+    @unittest.skipIf(not HAS_FALKORDB, 'FalkorDB is not installed')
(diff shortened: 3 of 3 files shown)


robowen1:
Reporter here — thanks for picking this up so quickly. The diff matches what I monkey-patched locally to validate the issue, and I can confirm the same fix took my Sonnet 4.6 + FalkorDB corpus rerender from ~30% failure (16/55 episodes) to 0% 

Quick note on correctness: returning '' is the right contract — it matches the existing max_query_length overflow branch, and the FalkorSearchOperations.{node,edge,episode,community}_fulltext_search callers already short-circuit on fuzzy_query == '' with return []. So no caller changes are needed and the malformed (@group_id:…) () can't be constructed.

LGTM — happy to see this land.

danielchalef:
Thanks for this fix! Closing as a duplicate of #1475, which we're taking as the fix for FalkorDB fulltext query-text sanitization — it covers the same backtick stripping + empty/stopword-only-token guard across both build paths, with regression tests co-located in the same file the maintainer fix (#1549) uses. (These query-text fixes are separate from #1549's group_id escaping.) Appreciate the contribution!

PR #1459, opened 2026-05-01
fix: mount FalkorDB volume to correct data path

Tunde-cs:
## Summary
Fix incorrect Docker volume mount path in `mcp_server/docker/docker-compose-falkordb.yml`. 
The volume was mounted to `/data` (a symlink into the container's writable layer) instead 
of `/var/lib/falkordb/data` (where FalkorDB actually writes `dump.rdb`). Data appeared to 
persist across restarts but was silently lost on `docker compose down`, `--force-recreate`, 
or image rebuilds.

## Type of Change
- [x] Bug fix

## Objective
N/A

## Testing
- [ ] Unit tests added/updated
- [x] All existing tests pass

Verified on a running stack:
- `docker exec docker-falkordb-1 ls /var/lib/falkordb/data/` → `dump.rdb` present
- `docker run --rm -v docker_falkordb_data:/check alpine ls /check/` → `dump.rdb` in the named volume

`make check` passes (format, lint, pyright, unit tests).

## Breaking Changes

## Checklist
- [x] Self-review completed

### DIFF
--- a/mcp_server/docker/docker-compose-falkordb.yml
+++ b/mcp_server/docker/docker-compose-falkordb.yml
@@ -8,7 +8,7 @@ services:
       - FALKORDB_PASSWORD=${FALKORDB_PASSWORD:-}
       - BROWSER=${BROWSER:-1}  # Enable FalkorDB Browser UI (set to 0 to disable)
     volumes:
-      - falkordb_data:/data
+      - falkordb_data:/var/lib/falkordb/data
     healthcheck:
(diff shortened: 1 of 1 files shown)


Naseem77:
@Tunde-cs Good news: this exact change is already on main, so this PR can be closed. Thanks anyway!

prasmussen15:
The change is already on main, closing the PR.

PR #1464, opened 2026-05-03
fix(mcp): preserve default temperature for non-openai llms

### LINKED ISSUE #1103: [BUG] MCP Server: Anthropic provider fails with 400 error when temperature not explicitly set
# MCP Server: Anthropic provider fails with 400 error when temperature not explicitly set in config

## Description

When using the Anthropic LLM provider without explicitly setting `temperature` in `config.yaml`, the API returns a 400 error because `None` is passed instead of a valid float.

The MCP server's `LLMConfig` model in `mcp_server/src/config/schema.py` defaults temperature to `None`:

```python
temperature: float | None = Field(
    default=None, description='Temperature (optional, defaults to None for reasoning models)'
)
```

This `None` value is then passed directly to the Anthropic API, which expects a numeric value.

## Error

```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'temperature: Input should be a valid number'}}
```

## Environment

- graphiti-mcp-server v0.24.3
- graphiti-core v0.24.3
- Provider: Anthropic (claude-sonnet-4-5-20250929)

## Root Cause

1. `graphiti_core/llm_client/config.py` defines a sensible default:

```python
DEFAULT_TEMPERATURE = 1

class LLMConfig:
    def __init__(
        self,
        temperature: float = DEFAULT_TEMPERATURE,  # Works correctly
        ...
    ):
```

2. `mcp_server/src/config/schema.py` (L151-153) overrides this with `None`:

```python
temperature: float | None = Field(
    default=None, description='Temperature (optional, defaults to None for reasoning models)'
)
```

3. `mcp_server/src/services/factories.py` (L206-212) passes `None` explicitly, overriding the core default:

```python
llm_config = GraphitiLLMConfig(
    api_key=api_key,
    model=config.model,
    temperature=config.temperature,  # Passes None, overriding core default
    max_tokens=config.max_tokens,
)
return AnthropicClient(config=llm_config)
```

4. Anthropic API receives `None` and returns 400.

## Why This Happens

The `None` default was added to support OpenAI reasoning models (o1, o3, gpt-5) which reject the temperature parameter entirely. However, the implementation is backwards - the edge case (OpenAI reasoning models) drives the default, breaking the normal case (all other providers).

When using graphiti-core directly (without MCP server), the default works correctly because the core's `LLMConfig` defaults temperature to `1.0`.

## Workaround

Explicitly set temperature in config.yaml:

```yaml
llm:
  provider: "anthropic"
  model: "claude-sonnet-4-5-20250929"
  temperature: 1.0
```

## Suggested Fix

The `None` default exists to support OpenAI reasoning models, but this is the edge case, not the norm. The fix should be:

1. **Change the schema default to `1.0`** in `mcp_server/src/config/schema.py`:

```python
temperature: float | None = Field(
    default=1.0, description='Temperature (optional, set to None for OpenAI reasoning models)'
)
```

2. **Handle OpenAI reasoning models explicitly** in `mcp_server/src/services/factories.py` OpenAI case - detect reasoning models (o1, o3, gpt-5) and omit temperature for those.

This way:
- Anthropic/Gemini/Groq work out of the box with sensible default
- OpenAI reasoning models are handled as the exception they are
- Users can still override temperature in config if needed

## Related

- #874 (GPT-5 temperature parameter handling)
- #878 (OpenAI reasoning model temperature fix)

fengfeng-zi:
## Summary
- stop overriding Graphiti core's default temperature when MCP config leaves temperature unset for Anthropic, Gemini, and Groq providers
- keep explicit temperature overrides intact
- add focused factory regression tests for unset and explicit temperature behavior

## Testing
- .venv\\Scripts\\python.exe -m pytest mcp_server\\tests\\test_factories.py -q

Closes #1103

### DIFF
--- a/mcp_server/src/services/factories.py
+++ b/mcp_server/src/services/factories.py
@@ -94,6 +94,26 @@ def _validate_api_key(provider_name: str, api_key: str | None, logger) -> str:
     return api_key
 
 
+def _build_graphiti_llm_config(
+    *,
+    api_key: str,
+    model: str,
+    max_tokens: int,
+    temperature: float | None,
+    base_url: str | None = None,
+) -> GraphitiLLMConfig:
+    kwargs = {
+        'api_key': api_key,
+        'model': model,
+        'max_tokens': max_tokens,
+    }
+    if temperature is not None:
+        kwargs['temperature'] = temperature
+    if base_url is not None:
+        kwargs['base_url'] = base_url
+    return GraphitiLLMConfig(**kwargs)
+
+
 class LLMClientFactory:
     """Factory for creating LLM clients based on configuration."""
 
@@ -199,7 +219,7 @@ class LLMClientFactory:
                 api_key = config.providers.anthropic.api_key
                 _validate_api_key('Anthropic', api_key, logger)
 
-                llm_config = GraphitiLLMConfig(
+                llm_config = _build_graphiti_llm_config(
                     api_key=api_key,
                     model=config.model,
                     temperature=config.temperature,
@@ -216,7 +236,7 @@ class LLMClientFactory:
                 api_key = config.providers.gemini.api_key
                 _validate_api_key('Gemini', api_key, logger)
 
-                llm_config = GraphitiLLMConfig(
+                llm_config = _build_graphiti_llm_config(
                     api_key=api_key,
                     model=config.model,
                     temperature=config.temperature,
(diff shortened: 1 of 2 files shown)


danielchalef:
Thanks for this — you correctly diagnosed that the MCP factory sends `temperature=None` to the non-OpenAI clients (Anthropic/Gemini/Groq), which pass it to their APIs verbatim and 400 (#1103). I've reimplemented the fix directly on main as #1570 (this branch was CI-red and behind), using the same omit-when-unset approach with a parametrized test across all three providers. Closing as superseded by #1570 — credit to you for the fix.

PR #1471, opened 2026-05-06
Fix FalkorDB fulltext search performance

XindongG:
## Summary

This fixes FalkorDB fulltext search paths that can become pathological on relationship-heavy graphs.

Changes:
- route high-level `search_utils` calls through `FalkorSearchOperations` by binding `FalkorDriver.search_interface`
- cap FalkorDB/RedisSearch fulltext OR terms by default, with `FALKORDB_FULLTEXT_MAX_TERMS=0` available to disable the cap
- avoid endpoint node `MATCH` work for edge fulltext results when no node-label filter is present, returning persisted `source_node_uuid` / `target_node_uuid` relationship properties instead
- when node labels require endpoint lookup, apply edge-only filters and `ORDER BY ... LIMIT` before matching endpoint nodes
- avoid the redundant episodic node rebind in `episode_fulltext_search`
- add regression coverage for the FalkorDB query builder and query generation

## Why

On FalkorDB, long RedisSearch OR queries and fulltext edge searches that `MATCH` endpoint nodes before `LIMIT` can time out even on modest graphs. In a real graph with roughly 7k `RELATES_TO` edges, the previous edge fulltext query shape consistently hit a 15s timeout. The optimized shape limits fulltext hits before endpoint lookup, and skips endpoint lookup entirely when endpoint labels are not part of the filter.

## Validation

- `uv run --extra dev pytest tests/driver/test_falkordb_driver.py` passed: 29 passed, 1 skipped
- `uv run --extra dev pytest tests/driver/test_falkordb_driver.py tests/utils/search/test_search_security.py` passed: 37 passed, 1 skipped
- `uv run --extra dev ruff format --check graphiti_core/driver/falkordb_driver.py graphiti_core/driver/falkordb/operations/search_ops.py tests/driver/test_falkordb_driver.py` passed
- `uv run --extra dev ruff check graphiti_core/driver/falkordb_driver.py graphiti_core/driver/falkordb/operations/search_ops.py tests/driver/test_falkordb_driver.py` passed
- `uv run --extra dev pyright graphiti_core/driver/falkordb_driver.py graphiti_core/driver/falkordb/operations/search_ops.py` passed
- `make lint` passed

`make test` was attempted, but this local environment hung for several minutes on `tests/cross_encoder/test_bge_reranker_client_int.py` while running `pytest -m "not integration"`. I stopped that run and validated the changed FalkorDB/search-security scope directly.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 import logging
+import os
 from typing import Any
 
 from graphiti_core.driver.driver import GraphProvider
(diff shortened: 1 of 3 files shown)


Naseem77:
@XindongG Heads up: FalkorSearchOperations is never invoked by the search pipeline (search_interface is None for FalkorDB), so this doesn't affect the live path. The real hot path is search_utils.py.

prasmussen15 (changes requested):
Please move the fulltext-search fix to graphiti_core/search/search_utils.py, which is the path the FalkorDB pipeline actually executes. The FalkorSearchOperations-only change does not affect production behavior.

PR #1474, opened 2026-05-07
fix: force batch_size=1 for gemini-embedding-2 models

### LINKED ISSUE #1467: GeminiEmbedder.create_batch silently returns wrong number of vectors for gemini-embedding-2-preview (and -2)
## Summary

`GeminiEmbedder.create_batch(input_data_list)` silently returns fewer vectors than inputs when used with `gemini-embedding-2-preview` or `gemini-embedding-2`, causing the caller's `zip(..., strict=True)` to raise `ValueError`. The root cause is that `gemini-embedding-2*` does not batch via `embed_content(contents=[...])` — it returns a single embedding regardless of how many strings are in the list. `gemini-embedding-001` does batch correctly. graphiti already pre-empts this for `-001` by hard-coding `batch_size = 1`; the same defence is missing for the `-2*` family.

## Reproduction

Tested against graphiti-core 0.29.0, google-genai 1.74.0, Python 3.12.

```python
import asyncio
from google import genai
from graphiti_core.embedder.gemini import GeminiEmbedder, GeminiEmbedderConfig

async def main():
    client = genai.Client()
    for model in ["gemini-embedding-2-preview", "gemini-embedding-001"]:
        cfg = GeminiEmbedderConfig(embedding_model=model)
        emb = GeminiEmbedder(cfg, client=client)
        print(f"{model}: batch_size={emb.batch_size}")
        vectors = await emb.create_batch(["Eli", "Anthropic", "Sydney"])
        print(f"  create_batch(3 inputs): got {len(vectors)} vectors")

asyncio.run(main())
```

Output:

```
gemini-embedding-2-preview: batch_size=100
  create_batch(3 inputs): got 1 vectors        ← BUG
gemini-embedding-001: batch_size=1
  create_batch(3 inputs): got 3 vectors        ← graphiti's special case saves it
```

`gemini-embedding-2` (the production sibling, no `-preview` suffix) behaves the same as `-2-preview`.

## Symptom in real use

When `gemini-embedding-2-preview` is configured and an `add_episode` call extracts ≥2 entities, graphiti's dedup pipeline trips on the partial result. The traceback (graphiti-core 0.29.0):

```
File ".../graphiti_core/utils/maintenance/node_operations.py", line 446, in _semantic_candidate_search
    for node, query_vector in zip(extracted_nodes, query_vectors, strict=True)
ValueError: zip() argument 2 is shorter than argument 1
```

In 0.28.2 there is no `_semantic_candidate_search`, but the same bug is latent in `nodes.py:1079` (`create_entity_node_embeddings`) — `zip(filtered_nodes, name_embeddings, strict=True)` will fail the moment a single episode produces ≥2 entities.

## Why `create_batch` returns fewer than expected

`GeminiEmbedder.create_batch` calls:

```python
result = await self.client.aio.models.embed_content(
    model=...,
    contents=batch,                     # list of N strings
    config=types.EmbedContentConfig(output_dimensionality=...),
)
for embedding in result.embeddings:    # only iterates len(result.embeddings) times
    all_embeddings.append(embedding.values)
```

For `gemini-embedding-2-preview`/`-2`, `result.embeddings` has length 1 regardless of `len(batch)`. The for-loop appends one item, no exception is raised, and the function returns a list shorter than `input_data_list`. Callers that use `zip(..., strict=True)` then fail.

This isn't a transient API hiccup — it's reproducible across runs and across `EmbedContentConfig` variations (with/without `task_type`, `output_dimensionality`). Models matching `embed*` reported by `models.list()`:

```
models/gemini-embedding-001         supports embedContent + asyncBatchEmbedContent
models/gemini-embedding-2-preview   supports embedContent + asyncBatchEmbedContent
models/gemini-embedding-2           supports embedContent + asyncBatchEmbedContent
```

`asyncBatchEmbedContent` is a separately-supported action — the proper batched API for the `-2*` family. `embed_content(contents=[…])` with `-2*` treats the list as parts of a single document, not as a batch.

## Suggested fix

Either (a) extend the existing special case to the `-2*` family:

```python
# graphiti_core/embedder/gemini.py
if batch_size is None and self.config.embedding_model in (
    "gemini-embedding-001",
    "gemini-embedding-2-preview",
    "gemini-embedding-2",
):
    self.batch_size = 1
```

or (b) detect the partial-batch return and fall back to per-input calls:

```python
result = await self.client.aio.models.embed_content(...)
if not result.embeddings:
    raise Exception("No embeddings returned")
if len(result.embeddings) != len(batch):
    # Fall back to per-input mode
    for item in batch:
        single = await self.client.aio.models.embed_content(model=..., contents=[item], config=...)
        all_embeddings.append(single.embeddings[0].values)
    continue
for embedding in result.embeddings:
    all_embeddings.append(embedding.values)
```

(a) is consistent with the existing pattern. (b) is more defensive against future model behavior changes.

## Workaround for users today

Pass `batch_size=1` explicitly:

```python
GeminiEmbedder(
    GeminiEmbedderConfig(embedding_model="gemini-embedding-2-preview"),
    client=genai_client,
    batch_size=1,
)
```

## Environment

- graphiti-core: 0.29.0 (also reproduces in 0.28.2 via the `nodes.py` caller)
- google-genai: 1.74.0
- Python: 3.12.11
- Platform: macOS arm64

Maanik23:
## Summary

Fix `GeminiEmbedder.create_batch()` silently returning fewer vectors than inputs when using `gemini-embedding-2-preview` or `gemini-embedding-2` models.

Closes #1467

## Problem

The `gemini-embedding-2*` models do not support batching via `embed_content(contents=[...])`. They treat the list as parts of a single document and return only **one** embedding vector regardless of how many strings are passed.

This caused `create_batch()` to silently return fewer vectors than inputs, leading to `ValueError` in downstream `zip(..., strict=True)` calls during entity deduplication (e.g. in `_semantic_candidate_search`).

## Fix

1. **Extend the `batch_size=1` special case** — the same defense already applied to `gemini-embedding-001` is now applied to the `-2` and `-2-preview` models. This ensures each input is sent individually, matching the API's actual behavior.

2. **Add a defensive length-mismatch check** — if a future model (or misconfiguration) causes the API to return fewer embeddings than requested, `create_batch()` now detects this and triggers the existing per-item fallback path instead of silently returning partial results.

## Tests

Added 6 new tests covering:
- `gemini-embedding-2-preview` defaults to `batch_size=1`
- `gemini-embedding-2` defaults to `batch_size=1`
- `gemini-embedding-001` still defaults to `batch_size=1` (existing behavior)
- Other models use default batch size (100)
- Explicit `batch_size` overrides model-specific default
- Length-mismatch triggers fallback to individual processing

### DIFF
--- a/graphiti_core/embedder/gemini.py
+++ b/graphiti_core/embedder/gemini.py
@@ -76,8 +76,13 @@ class GeminiEmbedder(EmbedderClient):
         else:
             self.client = client
 
-        if batch_size is None and self.config.embedding_model == 'gemini-embedding-001':
-            # Gemini API has a limit on the number of instances per request
+        if batch_size is None and self.config.embedding_model in (
+            'gemini-embedding-001',
+            'gemini-embedding-2-preview',
+            'gemini-embedding-2',
+        ):
+            # These models do not support batching via embed_content(contents=[...]).
+            # They treat the list as parts of a single document and return one embedding.
             # https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api
             self.batch_size = 1
         elif batch_size is None:
@@ -145,6 +150,12 @@ class GeminiEmbedder(EmbedderClient):
 
                 if not result.embeddings or len(result.embeddings) == 0:
                     raise Exception('No embeddings returned')
+
+                if len(result.embeddings) != len(batch):
+                    raise Exception(
+                        f'Batch embedding returned {len(result.embeddings)} vectors '
+                        f'for {len(batch)} inputs; model may not support batching'
+                    )
 
                 # Process embeddings from this batch
                 for embedding in result.embeddings:
--- a/tests/embedder/test_gemini.py
+++ b/tests/embedder/test_gemini.py
@@ -390,6 +390,82 @@ class TestGeminiEmbedderCreateBatch:
         assert len(result) == 2
         assert all(len(embedding) == 512 for embedding in result)
 
+    @pytest.mark.asyncio
+    @patch('google.genai.Client')
+    async def test_create_batch_length_mismatch_falls_back_to_individual(
+        self, mock_client_class, mock_gemini_client: Any
+    ) -> None:
+        """Test that create_batch falls back to individual processing when batch returns wrong count."""
+        config = GeminiEmbedderConfig(api_key='test_api_key', embedding_model='future-model')
+        embedder = GeminiEmbedder(config=config, batch_size=100)
+        embedder.client = mock_gemini_client
+
+        # Batch call returns only 1 embedding for 3 inputs (simulates gemini-embedding-2 behavior)
+        mock_batch_response = MagicMock()
+        mock_batch_response.embeddings = [create_gemini_embedding(0.1)]
+
+        # Individual calls each return 1 embedding
+        mock_individual_responses = [
+            MagicMock(embeddings=[create_gemini_embedding(0.1)]),
+            MagicMock(embeddings=[create_gemini_embedding(0.2)]),
+            MagicMock(embeddings=[create_gemini_embedding(0.3)]),
+        ]
+
+        mock_gemini_client.aio.models.embed_content.side_effect = [
+            mock_batch_response,
+            *mock_individual_responses,
+        ]
+
+        result = await embedder.create_batch(['Input 1', 'Input 2', 'Input 3'])
+
+        # Should fall back and produce 3 embeddings
+        assert len(result) == 3
+        # First call is the batch, next 3 are individual
+        assert mock_gemini_client.aio.models.embed_content.call_count == 4
+
+
+class TestGeminiEmbedderBatchSizeConfig:
+    """Tests for GeminiEmbedder batch_size configuration for different models."""
+
+    @patch('google.genai.Client')
+    def test_gemini_embedding_001_forces_batch_size_1(self, mock_client):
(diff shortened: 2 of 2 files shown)


brentkearney:
We hit the same bug (#1467) in production — batched Gemini embedding calls returning fewer vectors than inputs, which then surfaces as `ValueError: zip() argument 2 is shorter than argument 1` deep in entity dedup. In the MCP server that error gets swallowed by the queue worker, so episodes were being silently dropped.

We fixed it on a branch with the same core idea as this PR, but two deliberately different choices, in case they're useful for comparison: https://github.com/getzep/graphiti/compare/main...brentkearney:graphiti:fix/gemini-embedding-batch-mismatch

1. **`batch_size=1` for all `gemini-embedding-*` models, not just the `-2` family.** The allowlist approach has already needed extending once (`-001`, now `-2`/`-2-preview`), and each new model name that misbehaves means another silent data-loss window until someone adds it. Treating Gemini batching as unsupported by default — with the explicit `batch_size` argument as the override for anyone who has verified batching works for their model — fails safe for future model names.

2. **Raise on a count mismatch instead of falling back to per-item retries.** A short result means the API misbehaved; retrying individually can hide that indefinitely. Raising makes the mismatch visible to the caller immediately (and with `batch_size=1` as the default, the check is a backstop rather than a hot path).

Happy for any of this to be folded into this PR if the maintainers prefer one PR over two — mainly wanted to leave the field notes and an alternative diff here rather than open a competing PR.

PR #1475, opened 2026-05-07
fix: handle backtick tokens and empty queries in FalkorDB fulltext search

### LINKED ISSUE #1440: [BUG] FalkorDB fulltext query builder produces malformed RediSearch queries on backtick tokens and stopword-only inputs

Maanik23:
## Summary

Fix `_build_falkor_fulltext_query` producing malformed RediSearch queries that cause `ResponseError: Syntax error` during entity dedup-search.

Closes #1440

## Problem

Two distinct failure modes in the FalkorDB fulltext query builder:

1. **Backtick tokens leak through** — When an LLM extracts entity names from code-fenced text (e.g. `` `add_episode` ``), the backtick character was not in the separator map and survived sanitization, causing RediSearch parsing errors.

2. **Empty parentheses on stopword/punctuation-only input** — When stopword filtering removes all tokens, the builder still constructed `(@group_id:"…") ()` which RediSearch rejects. There was no early-return guard for the "no usable tokens" case.

## Fix

Both copies of the query builder are fixed:
- `graphiti_core/driver/falkordb/operations/search_ops.py`
- `graphiti_core/driver/falkordb_driver.py`

Changes:
1. Added backtick (`` ` ``) to the separator character map
2. Added early return of empty string when no tokens remain after sanitization + stopword filtering

## Tests

Added 5 new tests covering:
- Backtick stripping in `_build_falkor_fulltext_query`
- Empty return on stopword-only input
- Empty return on punctuation-only input
- Backtick stripping in `FalkorDriver.build_fulltext_query`
- Empty return on stopword-only input via driver method

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -84,6 +84,7 @@ _SEPARATOR_MAP = str.maketrans(
         '|': ' ',
         '/': ' ',
         '\\': ' ',
+        '`': ' ',
     }
 )
 
@@ -123,6 +124,10 @@ def _build_falkor_fulltext_query(
     # Remove stopwords and empty tokens
     query_words = sanitized_query.split()
     filtered_words = [word for word in query_words if word and word.lower() not in STOPWORDS]
+
+    if not filtered_words:
+        return ''
+
     sanitized_query = ' | '.join(filtered_words)
 
     if len(sanitized_query.split(' ')) + len(group_ids or '') >= max_query_length:
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -395,6 +395,7 @@ class FalkorDriver(GraphDriver):
                 '|': ' ',
                 '/': ' ',
                 '\\': ' ',
+                '`': ' ',
             }
         )
         sanitized = query.translate(separator_map)
@@ -435,6 +436,10 @@ class FalkorDriver(GraphDriver):
         # Remove stopwords and empty tokens from the sanitized query
         query_words = sanitized_query.split()
         filtered_words = [word for word in query_words if word and word.lower() not in STOPWORDS]
+
+        if not filtered_words:
+            return ''
+
         sanitized_query = ' | '.join(filtered_words)
 
         # If the query is too long return no query
--- a/tests/utils/search/test_search_security.py
+++ b/tests/utils/search/test_search_security.py
@@ -76,6 +76,55 @@ def test_falkordb_fulltext_query_rejects_invalid_group_ids():
         FalkorDriver.build_fulltext_query(driver, 'test', ['bad"group'])
 
 
+def test_falkordb_fulltext_query_strips_backticks():
+    """Backtick characters should be sanitized to prevent RediSearch syntax errors."""
+    from graphiti_core.driver.falkordb.operations.search_ops import _build_falkor_fulltext_query
+
+    result = _build_falkor_fulltext_query('`add_episode`', ['group1'])
+    # Backticks should be removed; remaining tokens joined with OR
+    assert '`' not in result
+    assert result != ''
+
+
+def test_falkordb_fulltext_query_returns_empty_on_stopwords_only():
+    """When all tokens are stopwords, return empty string instead of malformed query."""
+    from graphiti_core.driver.falkordb.operations.search_ops import _build_falkor_fulltext_query
+
+    result = _build_falkor_fulltext_query('the and is', ['group1'])
+    assert result == ''
+
+
+def test_falkordb_fulltext_query_returns_empty_on_punctuation_only():
+    """When input is all special characters, return empty string."""
+    from graphiti_core.driver.falkordb.operations.search_ops import _build_falkor_fulltext_query
+
+    result = _build_falkor_fulltext_query('!!!...???', ['group1'])
+    assert result == ''
+
+
+def test_falkordb_driver_build_fulltext_query_strips_backticks():
+    """FalkorDriver.build_fulltext_query should also strip backticks."""
+    from graphiti_core.driver.falkordb_driver import FalkorDriver
+
+    driver = MagicMock(spec=FalkorDriver)
+    driver.sanitize = FalkorDriver.sanitize.__get__(driver, FalkorDriver)
+
+    result = FalkorDriver.build_fulltext_query(driver, '`add_episode`', ['group1'])
+    assert '`' not in result
+    assert result != ''
+
+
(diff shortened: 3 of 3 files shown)


danielchalef:
Triaging the FalkorDB fulltext PRs: we're taking this one as the fix for **query-text sanitization** (backtick stripping + empty/stopword-only-query guard). It's the most complete in this group and is **complementary** to #1549, which handles group_id validity/escaping — different parts of the pipeline. We've closed the overlapping PRs (#1450/#1415/#1375/#1222) as duplicates of this one.

Two things before merge: (1) please **sign the CLA** (the check is currently failing), and (2) it'll need a **rebase** onto `main` (currently behind) — ideally after #1549 lands. Thanks for the solid fix and tests!

kramerica-inc-dev:
We hit exactly these failure classes in production (self-hosted memory service on FalkorDB, graphiti-core 0.28.2) — episodes dead-lettering on `RediSearch: Syntax error` — and independently implemented the same two fixes as build-time patches while waiting for this PR. Validation data, in case it helps prioritize:

- Stopword-only input: the builder emitted `(@group_id:"canary") ()`, which the engine rejects with `RediSearch: Syntax error at offset 22 near canary`. With the guard the builder returns `''` and every caller already treats that as "skip fulltext".
- Unescaped special characters inside terms fragment into the same class of errors; the backtick addition matches what we shipped.

One **additional variant** we found live that this PR does not cover yet: **punctuation-only tokens**. Sanitized snake_case filenames produced a query like

```
(@group_id:"trading") (backtest | diag_ | b | c | d | e | g | h | _ | py)
```

and the bare `_` term fails with `RediSearch: Syntax error at offset 68 near h`, deterministically dead-lettering the episode after retries. We fixed it with two small additions on top of the changes in this PR:

1. treat `_` as a separator in the sanitize map (RediSearch tokenizes on underscores anyway, cf. the analysis in #1549), and
2. after stopword filtering, drop tokens that contain no alphanumeric character at all (`[w for w in filtered_words if any(ch.isalnum() for ch in w)]`) — this also catches other punctuation-only leftovers.

Both verified in production: the failing document now lands cleanly.

@Maanik23 — since this PR has been marked as the canonical fix but needs a CLA + rebase: if you're short on time, we're happy to carry it across the finish line (rebase onto main, fold in the punctuation-only guard above, extend the tests in `tests/utils/search/test_search_security.py`) with credit to you for the original fix — whatever the maintainers prefer.

Naseem77:
@Maanik23 Reproduced both crashes (backticks and empty queries) and your patch handles them. Heads up that #1590 covers similar ground, maintainers should pick one. LGTM!

prasmussen15 (approved):

PR #1482, opened 2026-05-09
fix: pass configured Neo4j database to execute_query

### LINKED ISSUE #1481: [BUG] Database Parameter Not Honored in Neo4jDriver.execute_query()

**Reason:**

**Root Cause:**

```python
# Original code (with bug)
```

```python
```

**My workaround:**  

abhinavmathur-atlan:
## Summary
Fixes `Neo4jDriver.execute_query()` so the configured database is passed to the Neo4j driver as the top-level `database_` argument instead of being inserted into Cypher `parameters_`.

This keeps query parameters separate from driver configuration and allows read/search queries to run against the configured database instead of Neo4j's default database.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Documentation/Tests

## Objective
Bug fix for configured Neo4j databases: writes already use `session(database=self._database)`, but reads through `execute_query()` were not selecting the configured database.

## Testing
- [ ] All existing tests pass

Validation run:
- `uv run --extra dev pytest tests/driver/test_neo4j_driver.py`
- `uv run --extra dev pytest tests/driver`
- `uv run --extra dev ruff format graphiti_core/driver/neo4j_driver.py tests/driver/test_neo4j_driver.py`
- `uv run --extra dev ruff check graphiti_core/driver/neo4j_driver.py tests/driver/test_neo4j_driver.py`
- `uv run --extra dev pyright ./graphiti_core`

I also started `DISABLE_FALKORDB=1 DISABLE_KUZU=1 DISABLE_NEPTUNE=1 uv run --extra dev pytest -m "not integration"`; it progressed through the driver tests but then hit repeated errors in existing graph tests outside this driver change, so I stopped it and kept the validation scoped to the touched driver path.

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1481

### DIFF
--- a/graphiti_core/driver/neo4j_driver.py
+++ b/graphiti_core/driver/neo4j_driver.py
@@ -161,12 +161,10 @@ class Neo4jDriver(GraphDriver):
                 raise
 
     async def execute_query(self, cypher_query_: LiteralString, **kwargs: Any) -> EagerResult:
-        # Check if database_ is provided in kwargs.
-        # If not populated, set the value to retain backwards compatibility
         params = kwargs.pop('params', None)
         if params is None:
             params = {}
-        params.setdefault('database_', self._database)
+        kwargs.setdefault('database_', self._database)
 
         try:
             result = await self.client.execute_query(cypher_query_, parameters_=params, **kwargs)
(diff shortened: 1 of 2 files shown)


mehulp93:

Please note:

Issue #1489, opened 2026-05-15
[Bug + Feature] Historical backfill exposes two temporal-correctness gaps


brentkearney:
## Summary

While attempting to **backfill historical data** into Graphiti -— ingesting data from old session log files, so that the resulting facts are stamped with the time the underlying event actually happened, not the time ingestion ran -— two independent gaps were uncovered. The bi-temporal model is the differentiator that makes backfilling work: agents can reason about "what was true when," and contradictions get resolved via temporal supersedence. However, in the case of importing structured data, we know precisely when an event happened, so we need not rely on inference to discern the dates from the episode body, if we are able to pass it in as a parameter.

I propose modifying the MCP interface to add `reference_time`, and updating `delete_episode` so that it does not leave dangling references, pointing at no-longer existing nodes, in the graph.

## The use case

A "memory consolidation" script runs on old Claude session logs, to extract useful context information and import it into the graph using `add_memory` on the Graphiti MCP server for each one. Each save passes the session's actual date as the temporal reference.

For this to work end-to-end two things are needed to behave correctly:

1. The MCP `add_memory` tool must accept and propagate a caller-supplied `reference_time`.
2. In case of mistakes or corrections and we want to delete an entry, the MCP `delete_episode` tool must cascade to the edges and entities derived from the episode (so a bad ingest can be cleanly undone and re-tried).

Currently, Graphiti lacks both.

---

## Gap 1: MCP `add_memory` throws away `reference_time`

`graphiti_core.Graphiti.add_episode` already accepts a `reference_time` parameter and uses it correctly. The MCP wrapper at `mcp_server/src/graphiti_mcp_server.py` does not expose it — `add_memory` has no such parameter, and the internal queue service hardcodes `reference_time=datetime.now(timezone.utc)`.

Result: every MCP-driven ingest is stamped with the ingestion time, instead of the time when the underlying event happened. This blocks the backfill use case.

**Proposed fix:**
- Add optional `reference_time: str | None = None` (ISO-8601) to the MCP `add_memory` tool.
- Parse the string to `datetime` inside the tool (MCP can't carry `datetime` across the JSON boundary).
- Plumb through `QueueService.add_episode`, defaulting to `datetime.now(timezone.utc)` when the caller omits it — preserves current behavior for every existing caller.
- Unit test that an explicit `reference_time` reaches the underlying `add_episode` call and that `None` falls back to a fresh `now()`.

[...]

PR #1490, opened 2026-05-15
feat(mcp): add optional reference_time to add_memory

brentkearney:
Resolves part of #1489 (Gap 1).

## Summary

The MCP `add_memory` tool currently throws away the temporal-anchor capability that `graphiti_core.Graphiti.add_episode` already supports — the queue service hardcodes `reference_time=datetime.now(timezone.utc)`, so every MCP-driven ingest is stamped with ingestion time regardless of when the underlying event happened.

This PR exposes `reference_time` as an optional ISO-8601 parameter on the MCP tool and plumbs it through to `Graphiti.add_episode`, preserving the current default-to-now behavior when omitted.

## Why this matters

The bi-temporal model is the differentiator that makes Graphiti useful for historical backfill (memory consolidators, migration scripts, ingesting old logs/conversations). Without a way to pass the actual event time through MCP, every backfilled fact collapses to "true as of ingestion time" — which breaks temporal supersedence for any caller using the MCP transport.

`graphiti_core` already supports this via `Graphiti.add_episode(reference_time=...)`. This PR closes the matching gap in the MCP wrapper.

## Changes

| File | Change |
|---|---|
| `mcp_server/src/graphiti_mcp_server.py` | `add_memory` gains `reference_time: str | None = None`; parses ISO-8601 (including `Z` suffix) to `datetime` inside the tool; returns `ErrorResponse` on parse failure; docstring + example updated. |
| `mcp_server/src/services/queue_service.py` | `QueueService.add_episode` accepts `reference_time: datetime | None`, falling back to `datetime.now(timezone.utc)` when `None`. |
| `mcp_server/tests/test_queue_service.py` | New unit tests: (a) an explicit `reference_time` reaches the underlying `add_episode` call; (b) `None` defaults to a fresh `now()`. |

## Design notes

**Why a string on the MCP tool boundary?** MCP tool arguments cross a JSON boundary and a Python `datetime` doesn't deserialize directly. The tool accepts `str | None` (ISO-8601) and parses internally. The underlying `QueueService` method uses a proper `datetime | None` since it's all Python from there on.

**Backwards compatibility:** Required and preserved. Every existing caller (no `reference_time` argument) gets the same `datetime.now(timezone.utc)` behavior as before. Only callers that opt-in to passing the parameter see the new behavior.

**Error handling:** Invalid ISO-8601 strings return an `ErrorResponse` (consistent with the rest of the tool's error shape) rather than raising. Tested via the unit tests indirectly — bad strings never reach the queue.

## Verification

- `ruff check` + `ruff format --check` clean on all three files.
- `pytest tests/test_queue_service.py` passes (2/2).
- Manually verified on a self-hosted deployment: `add_memory(reference_time="2024-01-15T12:00:00Z", ...)` produces an Episodic node with `valid_at = 2024-01-15T12:00:00Z`, and the omit-the-arg path produces `valid_at ≈ now()`.

## Related PRs

This is PR 1 of 3 from #1489. The companion PRs:
- PR 2 — #1491
- PR 3 — #1492

### DIFF
--- a/mcp_server/src/graphiti_mcp_server.py
+++ b/mcp_server/src/graphiti_mcp_server.py
@@ -8,6 +8,7 @@ import asyncio
 import logging
 import os
 import sys
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Optional
 
@@ -326,6 +327,7 @@ async def add_memory(
     source: str = 'text',
     source_description: str = '',
     uuid: str | None = None,
+    reference_time: str | None = None,
 ) -> SuccessResponse | ErrorResponse:
     """Add an episode to memory. This is the primary way to add information to the graph.
 
@@ -345,6 +347,11 @@ async def add_memory(
                                - 'message': For conversation-style content
         source_description (str, optional): Description of the source
         uuid (str, optional): Optional UUID for the episode
+        reference_time (str, optional): ISO-8601 timestamp for when the episode actually occurred
+                                        (e.g. "2026-05-14T19:00:00Z"). Defaults to the ingestion time
+                                        when omitted. Use this to backfill historical events so the
+                                        knowledge graph's temporal ordering reflects reality rather
+                                        than when the episode was added.
 
     Examples:
         # Adding plain text content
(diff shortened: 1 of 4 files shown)


danielchalef:
Thanks for this — the bi-temporal `reference_time` on `add_memory` is a valuable addition, and your queue-service plumbing was on the right track.

It's now in `main` via **#1553** (the MCP core-parity work, merged): `add_memory` takes an optional `reference_time` (ISO-8601, naive→UTC-coerced) threaded through `QueueService.add_episode` → `Graphiti.add_episode`, replacing the hardcoded `datetime.now(timezone.utc)`. We implemented it there since this PR was blocked awaiting a CLA signature.

Closing as **superseded by #1553** — appreciate you surfacing it!

PR #1494, opened 2026-05-16
fix: avoid FalkorDB edge rematch scans

### LINKED ISSUE #1272: [BUG] [PERF] FalkorDB: edge_fulltext_search causes full graph scan due to re-MATCH pattern instead of using startNode/endNode
Description

The edge_fulltext_search method in FalkorSearchOperations uses a MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity) pattern to retrieve source and target nodes after a fulltext search on relationships. This causes FalkorDB to perform a full scan of all RELATES_TO edges for each fulltext result, leading to O(n×m) complexity and query timeouts on moderately-sized graphs.

The fulltext procedure already returns the relationship object — startNode(rel) and endNode(rel) can be used to directly access the connected nodes without any additional scan.

Environment
Graphiti version: latest (main branch)
Database: FalkorDB v4.x (Docker)
Graph size: ~5,169 RELATES_TO edges, ~1,500 Entity nodes
Reproduction
Run a fulltext edge search on any graph with >1,000 RELATES_TO edges:

python
results = await graphiti.search("API test system", group_ids=["my_group"])
The generated Cypher is:

cypher
CALL db.idx.fulltext.queryRelationships('RELATES_TO', '(@group_id:"my_group") (API | test | system)')
MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)  -- ← full scan here
WHERE e.group_id IN ["my_group"]
WITH e, score, n, m
RETURN ...
ORDER BY score DESC LIMIT 20
GRAPH.EXPLAIN confirms the scan:

Results
    Aggregate
        Filter
            Edge By Index Scan | [e:RELATES_TO]    ← scans ALL edges
                Node By Label Scan | (n:Entity)     ← scans ALL nodes
                    ProcedureCall
Benchmark Results
Query Pattern	Internal Time	Rows
Fulltext only (count)	2ms	1,492
Fulltext + MATCH {uuid: rel.uuid} (412 FT results)	26.7s	412
Full query (1,492 FT results + MATCH + WHERE + ORDER)	118.4s	20
Root Cause
In [
graphiti_core/driver/falkordb/operations/search_ops.py
](https://github.com/getzep/graphiti/blob/main/graphiti_core/driver/falkordb/operations/search_ops.py), 
the edge_fulltext_search
 method (and edge_bfs_search) 
 re-MATCHes the relationship by UUID instead of using the relationship object directly returned by the fulltext procedure:

python
# Current (slow) - line ~306
Suggested Fix
Replace the re-MATCH with direct endpoint access:

diff
- YIELD relationship AS rel, score
- MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)
+ YIELD relationship AS e, score
+ WITH e, score, startNode(e) AS n, endNode(e) AS m
This transforms the query from O(n×m) to O(n), eliminating the full graph scan entirely.

The same pattern also appears in 

edge_bfs_search
 (line ~418).

Impact
Queries timeout on graphs with ~5,000+ RELATES_TO edges
The default FalkorDB timeout (120s) is exceeded by the full query
Adding a Range index on RELATES_TO.uuid does not help — FalkorDB's query planner doesn't use it for this join pattern

mturac:
## Summary
Avoids the FalkorDB edge search re-MATCH pattern that scans `RELATES_TO` edges by `uuid` after the full-text procedure or BFS path traversal has already produced the relationship.

The full-text path now uses the returned relationship directly, and the BFS path filters path relationships to `RELATES_TO` before deriving source/target nodes with `startNode` / `endNode`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Performance improvement
- [ ] Documentation/Tests

## Objective
Fix the FalkorDB edge search query shape described in #1272 so edge endpoint lookup is linear in the returned relationships instead of forcing a relationship re-scan by UUID.

## Testing
- [ ] All existing tests pass

Validation run:
- `DISABLE_FALKORDB=1 DISABLE_KUZU=1 DISABLE_NEPTUNE=1 uv run pytest tests/driver/test_falkordb_search_ops.py`
- `uv run ruff format graphiti_core/driver/falkordb/operations/search_ops.py tests/driver/test_falkordb_search_ops.py --check`
- `uv run ruff check graphiti_core/driver/falkordb/operations/search_ops.py tests/driver/test_falkordb_search_ops.py`
- `git diff --check`

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1272

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -302,8 +302,8 @@ class FalkorSearchOperations(SearchOperations):
                 'edge_name_and_fact', limit=limit, provider=GraphProvider.FALKORDB
             )
             + """
-            YIELD relationship AS rel, score
-            MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)
+            YIELD relationship AS e, score
+            WITH e, score, startNode(e) AS n, endNode(e) AS m
             """
             + filter_query
             + """
@@ -415,7 +415,9 @@ class FalkorSearchOperations(SearchOperations):
             UNWIND $bfs_origin_node_uuids AS origin_uuid
             MATCH path = (origin {{uuid: origin_uuid}})-[:RELATES_TO|MENTIONS*1..{max_depth}]->(:Entity)
             UNWIND relationships(path) AS rel
-            MATCH (n:Entity)-[e:RELATES_TO {{uuid: rel.uuid}}]-(m:Entity)
+            WITH rel
+            WHERE type(rel) = 'RELATES_TO'
+            WITH rel AS e, startNode(rel) AS n, endNode(rel) AS m
             """
             + filter_query
             + """
--- a/tests/driver/test_falkordb_search_ops.py
+++ b/tests/driver/test_falkordb_search_ops.py
@@ -1,0 +1,52 @@
+from typing import Any
+
+import pytest
+
+from graphiti_core.driver.falkordb.operations.search_ops import FalkorSearchOperations
+from graphiti_core.search.search_filters import SearchFilters
+
+
+class RecordingExecutor:
+    def __init__(self):
+        self.cypher_query = ''
+        self.params: dict[str, Any] = {}
+
+    async def execute_query(self, cypher_query_: str, **kwargs: Any):
+        self.cypher_query = cypher_query_
+        self.params = kwargs
+        return [], None, None
+
+
+@pytest.mark.asyncio
+async def test_edge_fulltext_search_uses_returned_relationship_endpoints():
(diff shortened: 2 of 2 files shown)


danielchalef:
Thanks for this — your `startNode`/`endNode` approach is exactly right, and I've harvested your query-shape test (`tests/driver/test_falkordb_search_ops.py`) into #1500 with credit. Consolidating on #1500 because it also patches `graphiti_core/search/search_utils.py` — the path that actually runs on main (FalkorDriver doesn't set `search_interface`), so a driver-only fix wouldn't reach the live query. Closing as superseded by #1500.

PR #1497, opened 2026-05-20
fix: sanitize FalkorDB group ids in fulltext queries

### LINKED ISSUE #1483: [BUG] FalkorDriver.build_fulltext_query fails on hyphens in group_id (RediSearch syntax error)
## Summary

`FalkorDriver.build_fulltext_query` in graphiti-core 0.29 emits RediSearch queries that fail at parse time when `group_id` contains hyphens. The driver's comment claims to handle hyphens via quoting; the implementation only wraps in `"…"` without escaping.

This is related to but distinct from #1425 (hyphens in episode title) and #757 (general FalkorDB query errors). The root cause is the same RediSearch tokenizer behavior; this issue isolates it to the `group_id` field which is typically a UUID — so the bug fires 100% of the time for any user passing UUID-shaped group_ids.

## Reproduce

```python
import asyncio, json

async def main():
    driver = FalkorDriver(host="localhost", port=6379, database="repro")
    g = Graphiti(graph_driver=driver)
    await g.add_episode(
        name="repro",
        episode_body=json.dumps([{"subject": "a", "predicate": "X", "object": "b"}]),
        source=EpisodeType.json,
        source_description="repro",
        group_id="00000000-0000-0000-0000-000000000001",  # any hyphenated value
    )

asyncio.run(main())
```

## Expected

`add_episode` completes successfully.

## Actual

```
ERROR [graphiti_core.driver.falkordb_driver] Error executing FalkorDB query:
  RediSearch: Syntax error at offset 17 near 00000000
ResponseError: RediSearch: Syntax error at offset 17 near 00000000
```

The failing query (logged from a local patch) is:
```
    WHERE e.uuid in $edge_uuids AND e.group_id IN $group_ids
    ...
```

With params:
```
{'query': '(@group_id:"00000000-0000-0000-0000-000000000001") (Entity | has | …)', ...}
```

Offset 17 lands at the first hyphen inside `"00000000-0000-…"`.

## Root cause

`graphiti_core/driver/falkordb_driver.py`, around line 413:

```python
# Escape group_ids with quotes to prevent RediSearch syntax errors
# with reserved words like "main" or special characters like hyphens
escaped_group_ids = [f'"{gid}"' for gid in group_ids]
```

The comment is aspirational — wrapping in `"…"` doesn't escape hyphens for RediSearch's TEXT-field parser. Hyphens inside quoted phrases are still tokenized as separators at parse time and the parser raises a syntax error.

The fulltext index defined at `graph_queries.py:103` indexes `group_id` as a TEXT field via `db.idx.fulltext.createNodeIndex`. TEXT indexes tokenize hyphens as separators on STORE too, so the stored representation of `"foo-bar"` is `[foo, bar]`.

## Workaround (in our adapter layer)

We're shipping a private compat shim that overrides `build_fulltext_query` to scrub hyphens + underscores to spaces in `group_id` values before quoting. Multi-token phrase still matches the stored value because the index tokenizes the same way. Verified locally — full `add_episode` round-trip with hyphenated group_ids works after the patch.

Happy to send a PR if the maintainers would prefer either:
1. The scrub-to-space approach (lightest touch; matches index tokenization), or
2. Proper RediSearch backslash escaping (`\-`) inside the quoted phrase.

## Environment

- graphiti-core: 0.29.0
- Python: 3.13
- FalkorDB: latest (Docker)
- LLM/Embedder: OpenAI (defaults)

he-yufeng:
﻿## Summary
- sanitize FalkorDB group ids before inserting them into RediSearch fulltext filters
- keep the existing validation boundary: only already-valid group ids reach query construction
- cover a hyphenated UUID-style group id in the search security tests

## To verify
- `python -m pytest tests\utils\search\test_search_security.py -q`
- `python -m ruff check graphiti_core\driver\falkordb_driver.py graphiti_core\driver\falkordb\operations\search_ops.py tests\utils\search\test_search_security.py`
- `python -m ruff format --check graphiti_core\driver\falkordb_driver.py graphiti_core\driver\falkordb\operations\search_ops.py tests\utils\search\test_search_security.py`
- `python -m py_compile graphiti_core\driver\falkordb_driver.py graphiti_core\driver\falkordb\operations\search_ops.py tests\utils\search\test_search_security.py`

Fixes #1483

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -102,7 +102,7 @@ def _build_falkor_fulltext_query(
     if group_ids is None or len(group_ids) == 0:
         group_filter = ''
     else:
-        escaped_group_ids = [f'"{gid}"' for gid in group_ids]
+        escaped_group_ids = [f'"{_sanitize(gid)}"' for gid in group_ids]
         group_values = '|'.join(escaped_group_ids)
         group_filter = f'(@group_id:{group_values})'
 
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -405,7 +405,7 @@ class FalkorDriver(GraphDriver):
         else:
             # Escape group_ids with quotes to prevent RediSearch syntax errors
             # with reserved words like "main" or special characters like hyphens
-            escaped_group_ids = [f'"{gid}"' for gid in group_ids]
+            escaped_group_ids = [f'"{self.sanitize(gid)}"' for gid in group_ids]
             group_values = '|'.join(escaped_group_ids)
             group_filter = f'(@group_id:{group_values})'
 
--- a/tests/utils/search/test_search_security.py
+++ b/tests/utils/search/test_search_security.py
@@ -74,6 +74,19 @@ def test_falkordb_fulltext_query_rejects_invalid_group_ids():
         FalkorDriver.build_fulltext_query(driver, 'test', ['bad"group'])
 
 
+def test_falkordb_fulltext_query_sanitizes_group_ids_for_redisearch():
+    # Import inside the test so collection still works when FalkorDB extras are unavailable.
+    from graphiti_core.driver.falkordb.operations.search_ops import _build_falkor_fulltext_query
+
+    query = _build_falkor_fulltext_query(
+        'test',
+        ['00000000-0000-0000-0000-000000000001'],
+    )
+
+    assert '@group_id:"00000000 0000 0000 0000 000000000001"' in query
+    assert '00000000-0000-0000-0000-000000000001' not in query
+
+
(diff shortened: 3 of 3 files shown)


he-yufeng:
The CLAAssistant check appears to be failing before it can evaluate my CLA status:
 `graphql call to get the committers details failed: HttpError: Bad credentials`
This looks like a repository-side CLAAssistant/GitHub token configuration issue rather than a failed CLA signature on my side. Could a maintainer please re-run it or refresh the CLA Assistant credentials?

danielchalef:
Thanks for this. Closing as superseded by #1549, which escapes group_ids for RediSearch with backslash escaping (preserving exact-match semantics) and also fixes the default group_id validity. The space-tokenization approach here changes match semantics (e.g. `test-group` → two tokens), so we're going with #1549's escaping. Appreciate the contribution!

PR #1500, opened 2026-05-21
fix: avoid O(matches×graph) re-MATCH in edge search

jaredloman:
## Problem

Edge search re-MATCHes every relationship that a fulltext / path call already yielded, instead of using the relationship object directly:

```cypher
CALL db.idx.fulltext.queryRelationships('edge_name_and_fact', $query) YIELD relationship AS rel, score
```

The planner can't use the `RELATES_TO.uuid` index to anchor a relationship inside a `(node)-[rel {prop}]->(node)` pattern, so each yielded row triggers a graph-wide scan — O(results × graph).

On a moderately sized FalkorDB graph (**2,852 `Entity` / 18,849 `RELATES_TO` edges**) a single `search_memory_facts` reached **~25 minutes** (`GRAPH.SLOWLOG` latency 1,541,600 ms) and pinned multiple cores. The `db.idx.fulltext.queryRelationships` call itself returns in <1 ms; the re-MATCH is the entire cost.

## Affected code

This re-MATCH pattern appears in two places:

1. **`graphiti_core/search/search_utils.py`** — `edge_fulltext_search`, the **live path** used by `Graphiti.search()` when `driver.search_interface` is unset (the default).
2. **`graphiti_core/driver/falkordb/operations/search_ops.py`** — `edge_fulltext_search` + `edge_bfs_search` in the per-driver search operations.

## Fix

The fulltext / path call already returns the relationship object, so the endpoints are available directly via `startNode()` / `endNode()` — the re-MATCH is unnecessary:

```cypher
CALL db.idx.fulltext.queryRelationships('edge_name_and_fact', $query) YIELD relationship AS e, score
WITH e, score, startNode(e) AS n, endNode(e) AS m
```

`edge_bfs_search` gets the same treatment, with a `type(e) = 'RELATES_TO'` guard since `relationships(path)` over a `RELATES_TO|MENTIONS` path also yields `MENTIONS` edges that the old `[e:RELATES_TO {uuid: ...}]` pattern filtered out implicitly.

## Measurements

Verified end-to-end against a live FalkorDB instance (graphiti-core 0.28.x) — identical `search_memory_facts` calls:

| Search | Before | After |
|---|---:|---:|
| `edge_fulltext_search` (`search_utils.py` path) | ~25 min (timeout) | **<1 s** |
| `edge_fulltext_search` (driver `search_ops.py`) | ~25 min | **~12 ms** |
| `edge_bfs_search` (depth 2, 289-degree origin) | >45 s | **~1 s** |

## Correctness note (`edge_bfs_search`)

The old re-MATCH there was **undirected** — `(n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]-(m:Entity)` — which binds a pinned edge in both directions, returning each edge twice with `source`/`target` swapped (`RETURN DISTINCT` doesn't collapse them). `startNode(e)` / `endNode(e)` returns each edge once in its true orientation.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -314,8 +314,8 @@ class FalkorSearchOperations(SearchOperations):
                 'edge_name_and_fact', limit=limit, provider=GraphProvider.FALKORDB
             )
             + """
-            YIELD relationship AS rel, score
-            MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)
+            YIELD relationship AS e, score
+            WITH e, score, startNode(e) AS n, endNode(e) AS m
             """
             + filter_query
             + """
@@ -427,7 +427,9 @@ class FalkorSearchOperations(SearchOperations):
             UNWIND $bfs_origin_node_uuids AS origin_uuid
             MATCH path = (origin {{uuid: origin_uuid}})-[:RELATES_TO|MENTIONS*1..{max_depth}]->(:Entity)
             UNWIND relationships(path) AS rel
-            MATCH (n:Entity)-[e:RELATES_TO {{uuid: rel.uuid}}]-(m:Entity)
+            WITH rel AS e, startNode(rel) AS n, endNode(rel) AS m
+            WHERE type(e) = 'RELATES_TO'
+            WITH e, n, m
             """
             + filter_query
             + """
--- a/graphiti_core/search/search_utils.py
+++ b/graphiti_core/search/search_utils.py
@@ -201,8 +201,8 @@ async def edge_fulltext_search(
         return []
 
     match_query = """
-    YIELD relationship AS rel, score
-    MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)
+    YIELD relationship AS e, score
+    WITH e, score, startNode(e) AS n, endNode(e) AS m
     """
     if driver.provider == GraphProvider.KUZU:
         match_query = """
--- a/tests/test_falkordb_search_ops.py
+++ b/tests/test_falkordb_search_ops.py
@@ -1,0 +1,68 @@
+"""Query-shape regression tests for FalkorDB edge search (no live database).
+
+Guards the fix for the O(matches×graph) re-MATCH in edge search: the edge
+fulltext and BFS queries must derive endpoints from the relationship returned by
+the index (``startNode``/``endNode``) instead of re-matching every yielded edge
+by uuid against the whole graph (``MATCH (n)-[e {uuid: rel.uuid}]->(m)``), which
+caused multi-minute query times on large graphs.
+
+Harvested and adapted from #1494's driver test; runs as a unit test (a recording
+executor captures the emitted Cypher, so no FalkorDB connection is required).
+"""
+
+from typing import Any
+
+import pytest
+
+from graphiti_core.driver.falkordb.operations.search_ops import FalkorSearchOperations
+from graphiti_core.search.search_filters import SearchFilters
+
+
+class RecordingExecutor:
+    """Captures the Cypher and params a search method emits, returning no rows."""
+
+    def __init__(self):
+        self.cypher_query = ''
+        self.params: dict[str, Any] = {}
+
+    async def execute_query(self, cypher_query_: str, **kwargs: Any):
+        self.cypher_query = cypher_query_
+        self.params = kwargs
+        return [], None, None
+
+
+@pytest.mark.asyncio
+async def test_edge_fulltext_search_uses_returned_relationship_endpoints():
+    executor = RecordingExecutor()
+    operations = FalkorSearchOperations()
+
+    await operations.edge_fulltext_search(
+        executor,
+        'api test system',
+        SearchFilters(),
+        group_ids=['group-a'],
+    )
+
+    assert 'YIELD relationship AS e, score' in executor.cypher_query
+    assert 'WITH e, score, startNode(e) AS n, endNode(e) AS m' in executor.cypher_query
+    # The expensive per-row re-MATCH must be gone.
+    assert 'uuid: rel.uuid' not in executor.cypher_query
+
+
+@pytest.mark.asyncio
+async def test_edge_bfs_search_uses_path_relationship_endpoints():
+    executor = RecordingExecutor()
+    operations = FalkorSearchOperations()
+
+    await operations.edge_bfs_search(
+        executor,
+        ['origin-uuid'],
+        2,
+        SearchFilters(),
+        group_ids=['group-a'],
+    )
+
+    assert 'WITH rel AS e, startNode(rel) AS n, endNode(rel) AS m' in executor.cypher_query
+    # rel is rebound to e before the type guard, so the filter is on type(e).
+    assert "WHERE type(e) = 'RELATES_TO'" in executor.cypher_query
+    assert 'uuid: rel.uuid' not in executor.cypher_query


danielchalef:
Maintainer note: rebased onto current main and added a DB-free query-shape regression test (`tests/test_falkordb_search_ops.py`, harvested from #1494 — credit @mturac — and adapted to this PR's BFS `type(e)` guard). Selecting this as the canonical fix for the O(matches×graph) edge re-MATCH: it's the root-cause `startNode(e)`/`endNode(e)` rewrite, and crucially it patches **both** the FalkorDB driver `search_ops.py` **and** `graphiti_core/search/search_utils.py` — the latter is the path that actually executes today (FalkorDriver doesn't set `search_interface`). It also fixes `edge_bfs_search` and corrects the old undirected pattern that double-counted edges. Closing #1494 and #1507 as superseded. Thanks!

kramerica-inc-dev:
Production data in support of merging this: we run graphiti-core 0.28.2 on FalkorDB (self-hosted memory service, one graph per group_id) and the re-MATCH pattern this PR removes was our single biggest failure source — dense-document episodes took 15–25 minutes or dead-lettered on `Query timed out`, even after raising FalkorDB's query timeouts to 900s.

Evidence gathered on 2026-07-14 (graph with ~2.6k Entity nodes):

- `GRAPH.EXPLAIN` of the current query shape shows the pathology directly: per fulltext hit a `Node By Label Scan (n:Entity)` driving an `Edge By Index Scan` — O(hits × nodes). The rewritten shape (this PR's transform, which we backported as a build-time patch) has **no scan operators at all**: hits flow straight through Project/Filter.
- `GRAPH.PROFILE` of the rewritten query: **572 fulltext hits resolved in under 2 ms** end-to-end.
- End-to-end effect after deploying the backport: the same documents that previously ground for 15–25 minutes (or parked in our dead-letter queue after repeated timeouts) now land in 1–4 minutes; a replay of 11 previously-parked episodes completed with zero timeouts.

We can also confirm the maintainer note above from runtime observation: `search_utils.py` is the path that actually executes (`search_interface` is never assigned), so patching both locations, as this PR does, is exactly right.

One gap we found while validating: the **BFS** variant of the same antipattern in `search_utils.py` (`edge_bfs_search`, the generic branch shared by FalkorDB and Neo4j) is not covered by this PR — its BFS hunk lands in `falkordb/operations/search_ops.py` only, which is not reached at runtime. We've opened #1653 applying this PR's own BFS transform to that live path, framed explicitly as completing this one. Related: we also filed #1652 (empty graph name crash) and #1651 (MCP episode tools querying the wrong graph) from the same production debugging session.

Naseem77:
@jaredloman The fulltext half checks out with identical results. But the public edge_bfs_search generic branch still re-MATCHes, your BFS rewrite landed only in the unused FalkorSearchOperations path. Moving it to search_utils.py would complete it.

rschlek:
Independent production confirmation of both the bug and this fix, plus a data point on the BFS residual @Naseem77 flagged.

**Setup:** graphiti-core 0.29.2, FalkorDB, graph of ~10.4k entity nodes / ~7.7k `RELATES_TO` edges, ingest via `add_episode`.

**Before:** single `edge_fulltext_search` queries ran 10+ minutes; `add_episode` took ~45 min per episode with FalkorDB pinned at 300–450% CPU (single-threaded module saturated). `GRAPH.EXPLAIN` shows the `{uuid: rel.uuid}` re-MATCH scanning all edges per fulltext hit, as #1272 describes.

**After** applying this PR's `startNode()/endNode()` rewrite to `graphiti_core/search/search_utils.py` (the live path — confirming the consolidation rationale: patching only `FalkorSearchOperations` changed nothing for us, since `search_interface` is None for FalkorDB): the same queries return in seconds, FalkorDB idles at ~1% CPU, and ingest becomes purely LLM-bound.

**On the BFS residual:** confirmed — the generic `edge_bfs_search` in `search_utils.py` still re-MATCHes on the live path. We run the equivalent of this PR's rewrite there in production. One subtlety worth carrying into the fix: the BFS traversal is `[:RELATES_TO|MENTIONS*1..n]`, so once the re-MATCH (which implicitly type-filtered) is removed, a type guard is required:

```cypher
UNWIND relationships(path) AS rel
WITH DISTINCT rel AS e, startNode(rel) AS n, endNode(rel) AS m
WHERE type(e) = 'RELATES_TO'
```

(and any existing `filter_query` that begins with `WHERE` needs its first `WHERE` → `AND`.) With that guard the rewrite is behavior-identical for us; without it, `MENTIONS` edges leak into results. Happy to share exact timings or the full diff if useful.

prasmussen15 (changes requested):
Please rebase onto current main and ensure the live graphiti_core/search/search_utils.py path carries the complete edge-search fix. The driver-level FalkorSearchOperations path is not used by production; reconcile the BFS work explicitly with #1653 and retain regression coverage for the executable path.

prasmussen15 (changes requested):

seanperkins:
Independent reproduction, plus a controlled before/after, in case measured numbers help move this along.

Setup: `graphiti-core` 0.28.2, FalkorDB with the module loaded at its defaults (no `THREAD_COUNT` or other tuning).

**Graph shape:** 3,743 `Entity` nodes, 5,898 `RELATES_TO` edges — considerably smaller than the 18,849-edge graph in the description, which is why these are seconds rather than minutes. The effect shows up well before "large graph" scale.

### Isolating the cost

Decomposing the query supports the analysis in the description exactly:

```
db.idx.fulltext.queryRelationships(...) alone         1 ms   (20 rows)
single edge lookup by uuid, passed as a parameter     0 ms   (index used)
the two composed, as currently emitted             1945 ms
```

The index call is ~1ms and a uuid lookup is properly indexed on its own, but composing them costs 1.9s — consistent with the planner being unable to anchor on the index when the uuid arrives via a `YIELD` variable, and scanning per yielded row.

### The rewrite in this PR

```
current  (YIELD relationship AS rel + MATCH ... {uuid: rel.uuid})   1945.2 ms
this PR  (YIELD relationship AS e   + startNode(e)/endNode(e))         0.6 ms
```

Result sets are identical — same 20 rows, same `(edge, source, target)` triples, verified by set comparison rather than by eye.

### End-to-end effect on `add_episode`

Measured by wrapping the `llm_client` / `embedder` / `driver` boundaries and timing a single episode against the populated graph:

| | before | after |
|---|---|---|
| `add_episode` wall | 646.5s, 67.2s | 25.4s, 26.2s, 29.0s |
| `driver.execute_query` cumulative | 1920.2s, 71.1s | 5.1s, 3.1s, 4.0s |
| slowest single driver call | 338.8s | 0.53s |

Before the change the episode was graph-bound at roughly 48:1 over LLM time. After it, it is LLM-bound — 27–32s of LLM against 3–5s of driver.

### The variance may be the more interesting symptom

Two identical profiling runs 20 minutes apart, before the change, returned **646.5s and 67.2s** — a ~10x swing I could not account for and initially chased as a separate problem. After the rewrite, three consecutive runs sat within 7% of each other (25.4 / 26.2 / 29.0s).

So the re-MATCH appears to intermittently find a far worse plan, and the symptom isn't only "slow" but "unpredictably slow" — which is harder to attribute, and I suspect contributes to this being reported several times under different-sounding titles.

### Version note

Confirmed the pattern is still present on `main` and in 0.29.3 (`search_utils.py`, and the newer `driver/falkordb/operations/search_ops.py`), so upgrading isn't a workaround for anyone hitting this.

Happy to re-run any of the above if a different measurement would be more useful.

maui314159:
Independent confirmation on v0.29.3, in production: FalkorDB graph with 7,718 `Entity` nodes / 15,995 `RELATES_TO` edges — a single `search_memory_facts` went from sub-second to not completing within 2+ minutes once our group filter matched the full corpus (single core pinned; `GRAPH.INFO` showed the re-MATCH query as the only running work; the `queryRelationships` call itself returns in <1ms, matching your analysis exactly). A subtle aggravator: since the abandoned client request doesn't kill the server-side query, subsequent searches queue behind the still-grinding re-MATCH, so one slow search wedges the instance.

We applied this PR's `startNode()`/`endNode()` pattern to the same three sites and it fully resolves the edge paths: edge fulltext ~0.3s, full RRF edge hybrid ~1.4s, MCP `search_memory_facts` ~1s warm. Running it in production now.

One more occurrence of the same class that this PR doesn't cover — the episode path (`episode_fulltext_search` re-matching yielded nodes via unindexed `WHERE e.uuid = episode.uuid`): filed separately as #1819 since it needs its own one-line fix. Would love to see both land — as of v0.29.3 the failure mode is nasty: works fine on small graphs, then wedges permanently past a size threshold.

PR #1508, opened 2026-05-25
fix(kuzu): set _database default and create FTS indices in build_indices_and_constraints (#1258)

### LINKED ISSUE #1258: KuzuDriver missing _database attribute + FTS index not created (v0.28.1)
### Environment
- graphiti-core: v0.28.1
- kuzu: v0.11.3
- Python: 3.12.3
- OS: Ubuntu 24.04 (x86_64)

### Description

Two bugs in the new KuzuDriver introduced in v0.28.0:

#### Bug 1: Missing `_database` attribute

`KuzuDriver.__init__` does not set `self._database`, but `Graphiti.add_episode_bulk` references `self.driver._database` at line 1113:

```python
```

**Error:**
```
AttributeError: 'KuzuDriver' object has no attribute '_database'. Did you mean: 'with_database'?
```

**Workaround:** `driver._database = "default"` after initialization.

**Fix:** Add `self._database = db` (or a sensible default like `"default"`) to `KuzuDriver.__init__`, similar to how `Neo4jDriver` sets it.

#### Bug 2: FTS index `node_name_and_summary` not created

After calling `build_indices_and_constraints()`, fulltext search queries fail because the FTS index is not created:

```
RuntimeError: Binder exception: Table Entity doesn't have an index with name node_name_and_summary.
```

The query that fails:
```cypher
CALL QUERY_FTS_INDEX('Entity', 'node_name_and_summary', $query, TOP := $limit)
```

This happens during `add_episode_bulk` when Graphiti tries to deduplicate extracted nodes via fulltext search.

### Steps to Reproduce

```python
from graphiti_core.driver.kuzu_driver import KuzuDriver

driver = KuzuDriver(db="/tmp/test-kuzu")
driver._database = "default"  # workaround for bug 1

graphiti = Graphiti(graph_driver=driver, ...)
await graphiti.build_indices_and_constraints()
await graphiti.add_episode_bulk([episode], group_id="test")
# Bug 2 crashes here during node deduplication
```

### Expected Behavior

Both bugs should be fixed so the Kuzu driver works end-to-end like Neo4j and FalkorDB drivers.

### Additional Context

The Kuzu driver is a great addition (embedded, zero external dependencies). These appear to be oversights from the initial implementation in v0.28.0. Happy to submit a PR if helpful.

tobster1967:
Fixes #1258.

Two bugs in `KuzuDriver` that prevent `add_episode` and `search` from running end-to-end on v0.28+. Both surfaced independently while building a sibling `LadybugDriver` (LadybugDB is the embedded MIT-licensed Kuzu successor; that's a separate RFC discussion I'll open after this lands).

## Bug 1 — Missing `_database` attribute

`GraphDriver` declares `_database: str` as a typed-but-unbacked attribute. `Graphiti.add_episode` (graphiti.py:1032) and `Graphiti.add_episode_bulk` (graphiti.py:1113) both read `self.driver._database` for a group-id sanity check. `Neo4jDriver` sets it; `KuzuDriver.__init__` doesn't, so any call into those paths raises:

```
AttributeError: 'KuzuDriver' object has no attribute '_database'.
```

**Fix:** `self._database = ''` in `__init__`, mirroring `default_group_id: str = ''` on the base class.

## Bug 2 — `build_indices_and_constraints` was a no-op

The original comment claimed *"schema and indices are created during setup_schema()"*. That's a half-truth: `setup_schema` creates the *tables*, but FTS indices are emitted by `get_fulltext_indices(GraphProvider.KUZU)` and need a separate `CALL CREATE_FTS_INDEX` pass. `Graphiti.build_indices_and_constraints` calls into this driver method (not graph_ops directly), so the FTS pass never happened through the Graphiti entry point and subsequent `CALL QUERY_FTS_INDEX` raised:

```
Binder exception: Table Entity doesn't have an index with name node_name_and_summary.
```

**Fix:** iterate `get_range_indices` + `get_fulltext_indices` sequentially and swallow the specific `already exists` error class when `delete_existing=False`. The sequential + idempotent path is required because Kuzu does **not** support `CREATE_FTS_INDEX IF NOT EXISTS` — a second call to `build_indices_and_constraints` would otherwise raise `Binder exception: Index <name> already exists`. I verified this against `kuzu>=0.11.3` locally.

Note: I considered delegating to `self._graph_ops.build_indices_and_constraints` but it uses `semaphore_gather` with no idempotency, so the same duplicate-creation problem would surface there. Iterating in the driver keeps the failure mode predictable.

## Tests

New `tests/driver/test_kuzu_driver.py` — five tests, all skipped if `kuzu` is not installed (same `try/except ImportError` pattern as `test_falkordb_driver.py`):

| Test | Asserts |
|---|---|
| `test_provider_is_kuzu` | basic smoke |
| `test_database_attribute_is_set` | Bug 1 regression: `_database == ''` after construction |
| `test_build_indices_and_constraints_creates_fts_indices` | Bug 2 regression: `CALL QUERY_FTS_INDEX` runs without raising after `build_indices_and_constraints` |
| `test_build_indices_and_constraints_is_idempotent` | second call with `delete_existing=False` does not raise |
| `test_search_pipeline_after_build_indices_works` | end-to-end: insert + FTS query succeed |

All 5 pass locally with `kuzu>=0.11.3` installed. Existing FalkorDB tests untouched. No new ruff warnings.

## Verification

```
$ uv pip install -e ".[dev,kuzu]"
$ pytest tests/driver/ -v
...
======================== 28 passed, 1 skipped, 1 warning in 0.39s ========================
$ ruff check graphiti_core/driver/kuzu_driver.py tests/driver/test_kuzu_driver.py
All checks passed!
```

## Diff summary

- `graphiti_core/driver/kuzu_driver.py` — +30 / −4 (one new import, two edited methods)
- `tests/driver/test_kuzu_driver.py` — +127 (new file)

Total: 157 insertions, 4 deletions. Well under the 500-LOC RFC threshold from CONTRIBUTING.md.

## Note about the LadybugDriver

I'm working on adding a `LadybugDriver` for LadybugDB (Kuzu's embedded MIT-licensed successor). The two bugs above surfaced first there, and were then reproducible against `KuzuDriver`. The Ladybug work needs an RFC per CONTRIBUTING.md and will land in a separate issue+PR — this fix is the prerequisite, not the introduction.

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/graphiti_core/driver/kuzu_driver.py
+++ b/graphiti_core/driver/kuzu_driver.py
@@ -44,6 +44,7 @@ from graphiti_core.driver.operations.has_episode_edge_ops import HasEpisodeEdgeO
 from graphiti_core.driver.operations.next_episode_edge_ops import NextEpisodeEdgeOperations
 from graphiti_core.driver.operations.saga_node_ops import SagaNodeOperations
 from graphiti_core.driver.operations.search_ops import SearchOperations
+from graphiti_core.graph_queries import get_fulltext_indices, get_range_indices
 
 logger = logging.getLogger(__name__)
 
@@ -143,6 +144,11 @@ class KuzuDriver(GraphDriver):
         max_concurrent_queries: int = 1,
     ):
         super().__init__()
+        # GraphDriver declares ``_database: str`` typed but with no default;
+        # Graphiti.add_episode and add_episode_bulk both read
+        # ``self.driver._database``. Empty string mirrors
+        # ``default_group_id: str = ''`` on the base class. See #1258.
+        self._database = ''
         self.db = kuzu.Database(db)
 
         self.setup_schema()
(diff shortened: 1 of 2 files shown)


tobster1967:

danielchalef:
We've deprecated support for Kuzu as its authors have abandoned the project.

PR #1510, opened 2026-05-25
fix: optimize edge_fulltext_search LIMIT and remove group_id from Fal…

### LINKED ISSUE #1269: [BUG] FalkorDB: UUID group_ids with hyphens cause RediSearch syntax error (not fixed by PR #1175)
## Description

PR #1175 added double-quote escaping for `group_ids` in FalkorDB fulltext queries. However, UUID-formatted group_ids containing hyphens still cause `RediSearch: Syntax error`, because **RediSearch interprets hyphens as NOT operators even inside double quotes**.

This is a continuation of #841 and #757.

## Environment

- **graphiti-core**: 0.27.1 (includes PR #1175 fix)
- **FalkorDB**: latest Docker image
- **Python**: 3.12
- **OS**: macOS

## Reproduction

```python

driver = FalkorDriver(host="localhost", port=6379)
client = Graphiti(graph_driver=driver, llm_client=..., embedder=...)

# Use a UUID (with hyphens) as group_id — standard practice for multi-tenant apps
await client.add_episode(
    name="test",
    episode_body="user: my name is Michael",
    source="message",
    source_description="conversation",
    reference_time=datetime.now(UTC),
    group_id="f2c59d63-a7a7-b5e6-8000-000000000000",
)

# This triggers the error:
    query="Michael",
    group_ids=["f2c59d63-a7a7-b5e6-8000-000000000000"],
)
```

## Error

```
RediSearch: Syntax error at offset 20 near f2c59d63
```

## Root cause

`FalkorDriver.build_fulltext_query` generates:

```
(@group_id:"f2c59d63-a7a7-b5e6-8000-000000000000") (Michael)
```

RediSearch treats `-` as a NOT operator. Double quotes alone do not prevent this interpretation in FalkorDB's RediSearch implementation.

## Impact

**Complete blocker** for any application using UUIDs as `group_id` (which is the standard pattern for multi-tenant isolation).

## Suggested fix

The Cypher query already includes `WHERE n.group_id IN $group_ids` for filtering, making the `@group_id` prefix in the fulltext query **redundant**. The safest fix would be to **remove group_id filtering from the fulltext query** for FalkorDB and rely on the Cypher `WHERE` clause:

```python
def build_fulltext_query(self, query, group_ids=None, max_query_length=128):
    # Don't include group_id in fulltext query — the Cypher WHERE clause handles it
    sanitized_query = self.sanitize(query)
    query_words = sanitized_query.split()
    filtered_words = [w for w in query_words if w and w.lower() not in STOPWORDS]
    sanitized_query = ' | '.join(filtered_words)
    if len(sanitized_query.split(' ')) >= max_query_length:
        return ''
    return f'({sanitized_query})'
```

Alternatively, backslash-escape hyphens (`\-`) before constructing the fulltext query.

## Current workaround

Monkey-patch `build_fulltext_query` to omit `group_ids`:

```python
_original = falkor_driver.build_fulltext_query
def _patched(query, group_ids=None, max_query_length=128):
    return _original(query, group_ids=None, max_query_length=max_query_length)
falkor_driver.build_fulltext_query = _patched
```

## Related issues

- #841 — Regression: FalkorDB compatibility broken
- #757 — FalkorDB quickstart query error
- #1175 — Previous fix attempt (merged, but insufficient)

BaiYouQing:
## Description

This PR fixes two issues with FalkorDB search operations in `search_ops.py`:

### 1. Fix FalkorDB fulltext query when group_ids contain hyphens (Issue #1269)

FalkorDB treats hyphens as NOT operators in its fulltext search syntax, even inside
double-quoted strings. When an entity's `group_id` contains hyphens (e.g., "my-group-id"),
the `_build_falkor_fulltext_query` function would produce invalid queries. Since the
group_id filter is already applied at the Cypher `WHERE` level, this fulltext-level filter
is redundant and harmful.

**Change**: Removed the `group_id` filtering from `_build_falkor_fulltext_query`.

### 2. Optimize edge_fulltext_search LIMIT to prevent timeout (PR #1507)

In `edge_fulltext_search`, the Cypher query had `LIMIT $limit` after the per-row MATCH,
which means when FalkorDB fulltext search returns hundreds or thousands of candidates,
each candidate triggers a full MATCH traversal before the LIMIT is applied. This causes
"Max pending queries exceeded" and timeout errors as data grows.

**Change**: Hoisted `LIMIT $limit_with_buffer` before the `MATCH` clause. Added a 4x
buffer so post-MATCH WHERE filters still fill the requested limit.

## Related Issues
- Fixes #1269 (group_id hyphen handling)
- Addresses the root cause of #1507 (timeout during graph operations)

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -98,14 +98,14 @@ def _build_falkor_fulltext_query(
     group_ids: list[str] | None = None,
     max_query_length: int = MAX_QUERY_LENGTH,
 ) -> str:
-    """Build a fulltext query string for FalkorDB using RedisSearch syntax."""
-    if group_ids is None or len(group_ids) == 0:
-        group_filter = ''
-    else:
-        escaped_group_ids = [f'"{gid}"' for gid in group_ids]
-        group_values = '|'.join(escaped_group_ids)
-        group_filter = f'(@group_id:{group_values})'
+    """Build a fulltext query string for FalkorDB.
 
+    NOTE: group_id filtering is intentionally NOT included in the fulltext
+    query. RediSearch treats hyphens as NOT operators even inside double
+    quotes, which breaks group_ids with hyphens. This is safe because all
+    callers already apply a Cypher-level ``WHERE e.group_id IN $group_ids``
+    filter. See: https://github.com/getzep/graphiti/issues/1269
+    """
     sanitized_query = _sanitize(query)
 
     # Remove stopwords and empty tokens
@@ -113,11 +113,10 @@ def _build_falkor_fulltext_query(
     filtered_words = [word for word in query_words if word and word.lower() not in STOPWORDS]
     sanitized_query = ' | '.join(filtered_words)
 
-    if len(sanitized_query.split(' ')) + len(group_ids or '') >= max_query_length:
+    if len(sanitized_query.split(' ')) >= max_query_length:
         return ''
 
-    full_query = group_filter + ' (' + sanitized_query + ')'
-    return full_query
+    return sanitized_query
 
 
 class FalkorSearchOperations(SearchOperations):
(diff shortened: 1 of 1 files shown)


Naseem77:
@BaiYouQing The bug is real, I reproduced the 0 hits with hyphenated group_ids. But the patch lands in FalkorSearchOperations, which the pipeline never calls (search_interface is None for FalkorDB). The live path is search_utils.py. A rebase is needed too.

prasmussen15 (changes requested):
Please apply this fix to the live graphiti_core/search/search_utils.py path; FalkorSearchOperations is not called by the FalkorDB pipeline. Please also rebase onto current main.

Issue #1505, opened 2026-05-24
NaN/Inf values from embedder silently break entity deduplication and propagate to graph storage


Adelagric:
# NaN/Inf values from embedder silently break entity deduplication and propagate to graph storage

## Summary

There is no validation of embedding values anywhere in the embedder layer. A `NaN` or `Inf` returned by the provider (transient failure, batch endpoint corruption, partial response) propagates through `EntityNode.name_embedding` / `EntityEdge.fact_embedding`, breaks the entity deduplication path, and ends up persisted in the graph backend (`vecf32(...)` writes to Neo4j/FalkorDB, CSV writes to Neptune).

Observed on commit `34f56e6` (current `main`).

## Root cause

`graphiti_core/embedder/openai.py:54-66` returns the raw provider output without inspection:

```python
async def create(self, input_data) -> list[float]:
    result = await self.client.embeddings.create(
        input=input_data, model=self.config.embedding_model
    )
    return result.data[0].embedding[: self.config.embedding_dim]
```

Same shape in `voyage.py:54-76`. No `np.isfinite`, no length validation beyond truncation, no retry.

Downstream:

1. **Storage** — `models/nodes/node_db_queries.py:147,213,299` writes the vector directly via `SET n.name_embedding = vecf32(...)`. `NaN`/`Inf` is stored as-is.
2. **Dedup (Python path, bulk edges)** — `graphiti_core/utils/bulk_utils.py:534-538` computes `np.dot(normalize_l2(a), normalize_l2(b))`. With `NaN` in `a`, the result is `NaN`. The threshold check `similarity >= NODE_DEDUP_COSINE_MIN_SCORE` (0.6, defined at `graphiti_core/utils/maintenance/node_operations.py:65`) evaluates to `False`, so the edge is **treated as new**.
3. **Dedup (DB path, nodes)** — `node_operations.py:418-450` calls `node_similarity_search` which runs `WHERE score > $min_score` on the DB side. `NaN` scores are excluded, so the existing entity is **never matched**.

Net effect: a single transient `NaN` in the embedding output produces a duplicate entity / duplicate edge in the graph, with no signal to the caller.

## Reproduction

```python
import asyncio
import math
from graphiti_core.embedder.client import EmbedderClient

class NaNInjectingEmbedder(EmbedderClient):
    """Returns the same 1024-dim vector for any input, with a NaN at index 42."""
    def __init__(self):
        self.call_count = 0
    async def create(self, input_data):
[...]

PR #1511, opened 2026-05-25
fix: validate embeddings for NaN/Inf values with automatic retry

BaiYouQing:
## Description

Embedding providers can occasionally return corrupted responses containing NaN or Inf values in the embedding vector. When this happens, cosine similarity comparisons (used for entity deduplication) silently produce wrong results — `cosine_similarity(NaN, any_vector)` never exceeds 0.6, causing the dedup step to miss existing entities and create duplicate nodes.

This is related to Issue #1505.

### Changes
1. Added `_validate_embedding()` method that checks for NaN/Inf using numpy
2. On detection, logs a warning and raises ValueError to trigger a single retry
3. Both `create()` and `create_batch()` use this validation with automatic retry
4. Added `logging` and `numpy` imports

### Testing
- Tested with a local FalkorDB instance and DeepSeek/Tencent embedding APIs
- NaN detection correctly catches corrupt embeddings and retries
- After retry, the second call typically returns valid embeddings

### DIFF
--- a/graphiti_core/embedder/openai.py
+++ b/graphiti_core/embedder/openai.py
@@ -14,12 +14,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import logging
+
+import numpy as np
 from collections.abc import Iterable
 
 from openai import AsyncAzureOpenAI, AsyncOpenAI
 from openai.types import EmbeddingModel
 
 from .client import EmbedderClient, EmbedderConfig
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_EMBEDDING_MODEL = 'text-embedding-3-small'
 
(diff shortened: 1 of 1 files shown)


Adelagric:
Strong direction. Two extensions worth keeping in mind once this lands:

1. **Dimension validation.** `embedder/openai.py:60,66` silently truncates with `[: self.config.embedding_dim]` — if the provider returns fewer dims than expected, the resulting vector is shorter than `EMBEDDING_DIM`, mixing inconsistent-length vectors inside the same index. Worth a length check alongside the `isfinite` guard.

2. **L2 normalization invariant.** `helpers.py:116-119` applies `normalize_l2` inline in only two places (bulk edge dedup, MMR), but vectors persisted to the graph backend are the raw provider output. With FalkorDB's non-normalized cosine semantics (`graph_queries.py:158`), this drifts the meaning of the 0.6 dedup threshold.

Both could fit as separate sub-issues if the maintainers prefer to keep this PR focused on NaN/Inf. Out-of-process alternative for callers who want the full set of invariants enforced today: https://github.com/Adelagric/vector-router (Apache 2.0). Not a competitor to this PR — they compose.

Naseem77:
@BaiYouQing Nice guard, but only the OpenAI embedder validates. Azure/Gemini/Voyage can still return NaN. Extending it to all embedders would complete it.

prasmussen15 (changes requested):
Please extend the NaN/Inf embedding guard to every supported embedder, not only the OpenAI implementation, and add backend-agnostic regression coverage.

Adelagric:
This is the right call — the NaN/Inf guard only closes the gap for the embedder it lives in, and the same "trust the provider's return" assumption sits in every backend. Two concrete pieces for *extend-to-all + backend-agnostic*:

- **The surface is the per-embedder return points.** Each `embedder/*` `embed()`/`embed_batch()` hands back provider vectors unchecked, and several also silently reshape (e.g. the `[: embedding_dim]` truncation in the OpenAI embedder I flagged above) — which hides a wrong-dim return instead of surfacing it. Extending the guard means validating *finite + expected-dim* at each return, not only OpenAI's.
- **Backend-agnostic coverage** then falls out as a parametrized test over the embedder classes with a fake client returning a NaN / wrong-dim vector, asserting each raises (or retries) identically — so the contract is pinned uniformly rather than per-provider.

@BaiYouQing happy to pair if it helps — I can map the per-embedder return points and draft the parametrized regression, or just review as you extend it. Whichever's easier for you.

PR #1514, opened 2026-05-27
docs(mcp_server): fix Ollama example — use current nested provider config + working embedder

r266-tech:
The "Using Ollama for Local LLM" section in `mcp_server/README.md` doesn't work with the current configuration schema:

**1. Outdated config structure.** The example uses flat `api_base` / `api_key` keys, but `LLMConfig` / `EmbedderConfig` (`src/config/schema.py`) and the shipped `mcp_server/config/config.yaml` nest endpoint settings under `providers.<provider>.{api_key, api_url}`. There is no `api_base` field in the schema.

**2. Non-existent embedder provider.** The example sets `embedder.provider: "sentence_transformers"`, but `EmbedderFactory.create` (`src/services/factories.py`) only handles `openai`, `azure_openai`, `gemini`, and `voyage`, raising `ValueError: Unsupported Embedder provider` otherwise. There is no sentence_transformers embedder under `graphiti_core/embedder/` (it exists only as a cross-encoder/reranker), so following the current example fails at startup.

This PR updates the example to the nested provider format and points local embeddings at Ollama's OpenAI-compatible endpoint, which the code explicitly supports:

```python
# factories.py — EmbedderFactory.create, openai case
base_url=config.providers.openai.api_url,  # Support custom endpoints like Ollama
```

It also drops "Sentence Transformers" from the embedder list in the Features section, leaving the providers the factory actually implements.

### DIFF
--- a/mcp_server/README.md
+++ b/mcp_server/README.md
@@ -22,7 +22,7 @@ The Graphiti MCP server provides comprehensive knowledge graph capabilities:
 - **Graph Maintenance**: Clear the graph and rebuild indices
 - **Graph Database Support**: Multiple backend options including FalkorDB (default) and Neo4j
 - **Multiple LLM Providers**: Support for OpenAI, Anthropic, Gemini, Groq, and Azure OpenAI
-- **Multiple Embedding Providers**: Support for OpenAI, Voyage, Sentence Transformers, and Gemini embeddings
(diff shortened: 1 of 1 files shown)


r266-tech:
Closing this PR. The repo's `CLAAssistant` workflow is currently failing on a credentials error (`graphql call to get committers details failed: HttpError: Bad credentials`), so the CLA check can't go green regardless of signing. I've filed the same one-file docs fix as #1556 so a maintainer can apply it directly. Thanks!

PR #1524, opened 2026-06-01
fix: expose graphiti core version

### LINKED ISSUE #1104: [BUG] `graphiti_core` missing `__version__` attribute causes MCP server to log "unknown"
## Bug Description

The MCP server attempts to log the `graphiti_core` version at startup, but always displays "unknown" because:

1. `graphiti_core/__init__.py` doesn't expose a `__version__` attribute
2. The fallback logic in the MCP server never executes because the `except` block only catches import failures, not missing attributes

### Current behavior in `graphiti_core/__init__.py`

```python
from .graphiti import Graphiti
__all__ = ['Graphiti']
```

No `__version__` is defined.

### Current behavior in `mcp_server/src/graphiti_mcp_server.py` (around line 866)

```python
try:
    import graphiti_core
    graphiti_version = getattr(graphiti_core, '__version__', 'unknown')
    logger.info(f'  - Graphiti Core: {graphiti_version}')
except Exception:
    # Check for Docker-stored version file
    version_file = Path('/app/.graphiti-core-version')
    if version_file.exists():
        graphiti_version = version_file.read_text().strip()
        logger.info(f'  - Graphiti Core: {graphiti_version}')
```

Since `graphiti_core` imports successfully, `getattr()` returns `'unknown'` and the Docker fallback in the `except` block is never reached.

### 1. Add `__version__` to `graphiti_core/__init__.py`

Use `importlib.metadata` to pull the version from installed package metadata (populated from `pyproject.toml` at build time):

```python
from importlib.metadata import version, PackageNotFoundError

try:
    __version__ = version("graphiti-core")
except PackageNotFoundError:
    __version__ = "unknown"

from .graphiti import Graphiti

__all__ = ['Graphiti', '__version__']
```

This keeps `pyproject.toml` as the single source of truth and requires no additional maintenance on releases.

### 2. Fix the MCP server fallback logic

The current logic doesn't fall back to the Docker version file when `__version__` is missing. Here's a corrected version:

```python
graphiti_version = None
try:
    import graphiti_core
    graphiti_version = getattr(graphiti_core, '__version__', None)
except ModuleNotFoundError:
    logger.warning("Unable to import graphiti_core module")

if graphiti_version:
    logger.info(f'  - Graphiti Core: {graphiti_version}')
else:
    version_file = Path('/app/.graphiti-core-version')
    if version_file.exists():
        graphiti_version = version_file.read_text().strip()
        logger.info(f'  - Graphiti Core: {graphiti_version}')
    else:
        logger.info('  - Graphiti Core: version unavailable')
```

This ensures the Docker fallback file is actually checked when `__version__` is not available. There's one thing I'd like to mention. There might be reasons to allow the server to continue running if you can't import `graphiti_core`, but I would assume that if this module fails to be imported, the application won't function. In that case, I would recommend a `sys.exit(1)` at the end of the `except` clause, but that's outside the scope of the issue.

## Environment

- Graphiti Core Version: 0.24.3
- MCP Server Version: latest
- Running in Docker container

meaqua9420:
## Summary
- expose `graphiti_core.__version__` from package metadata
- make MCP startup version logging fall back to `/app/.graphiti-core-version` when the package version is unavailable
- add focused regression tests for both code paths

Fixes #1104

## Tests
- `uv run --extra dev pytest tests/test_version.py -q`
- `uv run --project mcp_server pytest mcp_server/tests/test_version_logging.py -q`
- `uv run --extra dev ruff format --check graphiti_core/__init__.py tests/test_version.py`
- `uv run --extra dev ruff check graphiti_core/__init__.py tests/test_version.py`
- `uv run --project mcp_server ruff format --check src/graphiti_mcp_server.py tests/test_version_logging.py`
- `uv run --project mcp_server ruff check src/graphiti_mcp_server.py tests/test_version_logging.py`
- `make lint`
- `uv run --project mcp_server pyright src/graphiti_mcp_server.py tests/test_version_logging.py`


meaqua9420:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: meaquamak@gmail.com

meaqua9420:
recheck

PR #1531, opened 2026-06-05
fix(falkor): strip nul bytes from parameters

### LINKED ISSUE #1525: [BUG] FalkorDB: NUL byte (\x00) in a string property fails the whole query with "Failed to parse query parameter '<name>' value"
### Summary

When any string value passed as a FalkorDB query parameter contains a NUL byte (`\x00`), FalkorDB rejects the **entire** query with:

```
Failed to parse query parameter '<name>' value
```

Because this happens at save time (e.g. the bulk Episodic save `UNWIND $episodes ... SET n.content = episode.content`), affected episodes are **silently dropped** from ingest — the error is per-episode and the run continues, so it's easy to miss.

### Environment
- `graphiti-core` 0.29.1, FalkorDB backend (`graphiti-core[falkordb]`), `falkordb` client 1.2.0.

### Root cause
FalkorDB query parameters are serialized into a `CYPHER <key>=<value> ...` header by the client. `falkordb/helpers.py::quote_string` escapes only backslash and double-quote:

```python
v = v.replace("\\", "\\\\")
v = v.replace('"', '\\"')
```

It does **not** handle `\x00`. A NUL embedded in a string value is emitted verbatim into the params header, and FalkorDB's parser fails on it. (Plain newlines/tabs are fine — only NUL trips it, which is why it shows up rarely.)

### How we hit it
Text extracted from PPTX/PDF source documents frequently contains stray NUL bytes (e.g. between glyphs). 12 of our ~2,000 episodes — all from board-meeting decks — failed ingest. Bisecting one failing episode's `content` pinpointed a single `\x00` at byte 864 (`"...Seamless Recruiting \x00 Onboarding..."`).

### Minimal reproduction
```python
g.query(
    "UNWIND $episodes AS e MERGE (n:Episodic {uuid:e.uuid}) SET n.content=e.content RETURN n.uuid",
    {"episodes": [{"uuid": "t", "content": "a \x00 b"}]},
)
# -> redis.exceptions.ResponseError: Failed to parse query parameter 'episodes' value
# Stripping the \x00 -> succeeds.
```

### Impact
Silent data loss during ingest for any content (commonly PPTX/PDF-extracted) containing NUL bytes.

### Suggested fix
NUL can never be stored in a FalkorDB string property, so it's safe to strip it. Either:
1. **graphiti-core (FalkorDB driver):** recursively strip `\x00` from string params in `FalkorDriver.execute_query` / `FalkorDriverSession.run` before sending (one-liner-ish, no behavior change for valid data), or
2. **falkordb-py:** have `quote_string` strip/reject NUL.

Happy to send a PR for option 1 if useful. (Distinct from the edge re-match perf issue #1272.)

he-yufeng:
## Summary

Fixes getzep/graphiti#1525.

FalkorDB rejects query parameters that contain embedded NUL bytes, and the failure happens before the query body runs. Ingested text from PDFs/PPTX can contain stray `\x00`, so a single bad string can make a whole bulk save query fail.

This strips NUL bytes from FalkorDB query parameters before calling the FalkorDB client. The cleanup is recursive for dict/list/tuple values and is applied in both driver paths that send params:

- `FalkorDriver.execute_query(...)`
- `FalkorDriverSession.run(...)`, including list-of-query batches

Valid strings are unchanged; only `\x00` is removed.

## To verify

- `$env:DISABLE_KUZU='True'; <pytest bootstrap>; pytest tests/driver/test_falkordb_driver.py::{new and adjacent tests} -q` -> `4 passed`
- `python -m py_compile graphiti_core\driver\falkordb_driver.py tests\driver\test_falkordb_driver.py`
- `python -m ruff check graphiti_core\driver\falkordb_driver.py tests\driver\test_falkordb_driver.py`
- `python -m ruff format --check graphiti_core\driver\falkordb_driver.py tests\driver\test_falkordb_driver.py`
- `git diff --check`

Note: this Windows environment has another editable checkout exposing a top-level `tests` package, so I used a small bootstrap to point `tests` at this repository before invoking pytest. `DISABLE_KUZU=True` avoids importing the optional Kuzu driver for these Falkor-only unit tests.

### DIFF
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -72,6 +72,18 @@ from graphiti_core.utils.datetime_utils import convert_datetimes_to_strings
 logger = logging.getLogger(__name__)
 
 
+def _strip_nul_bytes(value: Any) -> Any:
+    if isinstance(value, str):
+        return value.replace('\x00', '')
+    if isinstance(value, dict):
+        return {key: _strip_nul_bytes(item) for key, item in value.items()}
+    if isinstance(value, list):
+        return [_strip_nul_bytes(item) for item in value]
+    if isinstance(value, tuple):
+        return tuple(_strip_nul_bytes(item) for item in value)
+    return value
+
+
 class FalkorDriverSession(GraphDriverSession):
     provider = GraphProvider.FALKORDB
 
@@ -98,10 +110,12 @@ class FalkorDriverSession(GraphDriverSession):
         if isinstance(query, list):
             for cypher, params in query:
                 params = convert_datetimes_to_strings(params)
+                params = _strip_nul_bytes(params)
                 await self.graph.query(str(cypher), params)  # type: ignore[reportUnknownArgumentType]
         else:
             params = dict(kwargs)
             params = convert_datetimes_to_strings(params)
+            params = _strip_nul_bytes(params)
             await self.graph.query(str(query), params)  # type: ignore[reportUnknownArgumentType]
         # Assuming `graph.query` is async (ideal); otherwise, wrap in executor
         return None
@@ -225,6 +239,7 @@ class FalkorDriver(GraphDriver):
 
         # Convert datetime objects to ISO strings (FalkorDB does not support datetime objects directly)
         params = convert_datetimes_to_strings(dict(kwargs))
+        params = _strip_nul_bytes(params)
 
         try:
             result = await graph.query(cypher_query_, params)  # type: ignore[reportUnknownArgumentType]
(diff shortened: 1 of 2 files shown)


prasmussen15:
Hi @he-yufeng — thanks for this fix! 🙏
Before we can keep your contribution in the project, we need you to sign our Contributor License Agreement (CLA). You can read it here: 
 https://github.com/getzep/graphiti/blob/main/Zep-CLA.md

To sign, just reply to this PR with the following comment, exactly as written:
 Once you've signed, we'll restore your change. Thanks again for contributing to Graphiti!

PR #1175, opened 2026-01-25
fix(falkordb): escape group_ids in RediSearch fulltext queries

Milofax:
## Summary

- Group IDs like "main" are reserved words in RediSearch and cause syntax errors when used unquoted
- Group IDs with hyphens (e.g., "project-name") also need escaping
- This fix wraps all group_ids in double quotes in the fulltext query

**Before:** `(@group_id:main|project-name)`
**After:** `(@group_id:"main"|"project-name")`

## Problem

Intermittent errors like:
```
RediSearch: Syntax error at offset 20 near main
```

## Solution

Simple 4-line fix in `graphiti_core/driver/falkordb_driver.py`:
```python
escaped_group_ids = [f'"{gid}"' for gid in group_ids]
group_values = '|'.join(escaped_group_ids)
```

## Test plan

- [x] Tested with group_id "main" - no more syntax errors
- [x] Tested with group_id "Milofax-infrastructure" - works correctly
- [x] Tested with multiple group_ids - proper OR query generated

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1549, opened 2026-06-06
fix(falkordb): make default group_id valid and escape it for RediSearch

prasmussen15:
## Summary

Fixes the FalkorDB quickstart, which fails out of the box with the default group_id:
```
GroupIdValidationError: group_id "\_" must contain only alphanumeric characters, dashes, or underscores
```

### Root cause
Two earlier changes conflict:
- `get_default_group_id(FALKORDB)` / `FalkorDriver.default_group_id` returned `'\_'` (backslash-underscore) — the backslash escapes the underscore for RediSearch fulltext queries (#1175/#1183).
- `validate_group_id` (added in #1312 to harden search against injection) rejects anything outside `[a-zA-Z0-9_-]`, so it throws on the backslash.

When group_id is unspecified, graphiti substitutes the FalkorDB default `'\_'`; `add_episode`'s internal edge-resolution search then re-validates it and raises. The quickstart never completes.

### Fix
1. Use a **clean default group_id of `'_'`** (passes `validate_group_id`) in both `get_default_group_id` and `FalkorDriver.default_group_id`.
2. **Escape non-alphanumeric chars in group_ids when building RediSearch fulltext queries.** RediSearch treats `_` and `-` as token separators/operators, so a bare `@group_id:"_"` is a fulltext **syntax error** — it must be `@group_id:"\_"`. Escaping is applied in both `FalkorDriver.build_fulltext_query` (active path) and the driver-ops `_build_falkor_fulltext_query`. Since `validate_group_id` restricts group_ids to `[a-zA-Z0-9_-]`, escaping every non-alphanumeric char is sufficient and safe.

I probed RediSearch directly to confirm: with a node stored as group_id `_`, the query `@group_id:"\_"` parses **and matches**, whereas `@group_id:"_"` errors.

## Verification
Ran `examples/quickstart/quickstart_falkordb.py` end-to-end against a real FalkorDB container with the **default group_id** (no explicit group_id):
- 4 episodes added, basic search, center-node reranking, and node hybrid search all succeed
- **exit code 0, no exceptions**

Also confirmed the server path works under both redis-py 7.4.1 and 8.0.0.

Adds regression tests in `test_search_security.py`:
- the FalkorDB default group_id passes `validate_group_id`
- `_build_falkor_fulltext_query` escapes the default group_id to `@group_id:"\_"`

`ruff` + `pyright` clean on all changed files.

## Type of Change
- [x] Bug fix

## Testing
- [x] Unit tests added
- [x] Verified end-to-end locally (quickstart, default group_id)


danielchalef (approved):
lgtm

PR #1553, opened 2026-06-07
feat(mcp): core-parity — bi-temporal, sagas, communities, filters, triplets, custom types

danielchalef:
Closes the verified functionality gaps between the Graphiti MCP server and graphiti-core (audit window v0.24.0→v0.29.1). Built and adversarially reviewed by a multi-agent workflow; review verdict **SHIP-WITH-FIXES**, all must/should-fixes applied.

> **Stacked on #1552** (base = `modernize/mcp-server`) because these features need the v0.29 APIs (saga/community/`custom_extraction_instructions`) that #1552's `graphiti-core>=0.29.1` bump provides. Will retarget/rebase onto `main` once #1552 merges. Also **supersedes #1490's `reference_time`** (folded in here since #1490 lacks a contributor CLA).

## Bug fix
- **`delete_episode` was not orphan-safe** — it called raw `EpisodicNode.delete`, leaving entities/facts the episode created behind. Now uses `Graphiti.remove_episode`, which cascades solely-created edges/nodes. (docstring corrected.)

## New parameters on `add_memory` (threaded → queue → `Graphiti.add_episode`)
- `reference_time` (bi-temporal event time; ISO-8601, naive→UTC), `saga` + `saga_previous_episode_uuid`, `excluded_entity_types`, `custom_extraction_instructions`, `previous_episode_uuids`, `update_communities`. Configured `edge_types`/`edge_type_map` are auto-attached.

## New tools
- `summarize_saga(saga_name, group_id?)` — resolves the saga **name** to its UUID within the group (sagas are keyed by `(name, group_id)`) and summarizes.
- `build_communities(group_ids?)` — entity-community detection + summaries.
- `add_triplet(...)` — write a single source→fact→target directly (core dedups endpoints, generates embeddings).
- `get_episode_entities(episode_uuids)` — provenance: entities/facts created by given episodes.

## Search
- `search_memory_facts`: builds `SearchFilters` from new params — `edge_types` and `valid_at`/`invalid_at` date ranges.
- `search_nodes`: adds `center_node_uuid` (graph-centered recall).

## Custom types
- Configured entity types now register the **rich Pydantic models** (`models/entity_types.py`) instead of doc-only stubs, and a new **edge-type config path** (`models/edge_types.py` + `edge_types`/`edge_type_map` in `config/schema.py`). Added a `Person` entity type so the person↔org edge examples are coherent.
[...]

PR #1552, opened 2026-06-07
chore(mcp): modernize dependencies, MCP SDK, and default models

danielchalef:
Modernizes the Graphiti MCP server (deps, MCP SDK, default model, Docker pins). Produced and adversarially verified by a multi-agent workflow.

## Dependency bumps (`mcp_server/pyproject.toml`)
- `mcp` `>=1.9.4` → **`>=1.27.2,<2`**. Every SDK symbol the server uses is present and unchanged in 1.27.2 — `FastMCP(name, instructions=...)`, `@mcp.tool()`, `@mcp.custom_route(...)`, settable `mcp.settings.host/port`, and `run_stdio_async`/`run_sse_async`/`run_streamable_http_async`. **No code changes required** for the SDK bump. The `<2` ceiling guards against the planned v2 breaking release.
- `graphiti-core[falkordb]` `>=0.28.2` → **`>=0.29.1`** (latest PyPI stable).
- `sentence-transformers` `>=2.0.0` → **`>=3.2.1`** (aligns with graphiti-core's own floor).

## Default model → gpt-5.5, reasoning off
- Default LLM `gpt-4o-mini` → **`gpt-5.5`** in `src/config/schema.py` and the four `config/*.yaml` (interpolation default preserved). Embedder unchanged (`text-embedding-3-small`, 1536-dim).
- The OpenAI factory sends **`reasoning_effort: 'none'`** for `gpt-5.5` (reasoning off) — lower cost/latency with comparable extraction quality. Earlier reasoning models keep the historical `'minimal'` floor; non-reasoning models are unaffected.
- This works with the released `graphiti-core>=0.29.1`: the factory passes the effort explicitly and the core OpenAI client forwards it verbatim, so it does **not** depend on the unreleased core `'auto'` resolver (#1551).

## Docker pins
- `GRAPHITI_CORE_VERSION` `0.28.1` → `0.29.1` and `MCP_SERVER_VERSION` → `1.0.2` across `Dockerfile`, `Dockerfile.standalone`, `docker-compose.yml`.

## Code
- `src/services/factories.py`: graphiti-core 0.29.x narrowed `LLMConfig.temperature` to `float`; added `# type: ignore[arg-type]` (with comments) on the 5 client constructions where `None` is intentional for reasoning models (core stores it verbatim and downstream clients omit `temperature` when `None`). Plus the per-model reasoning-effort selection above. No behavior change for non-reasoning paths.

## Verification
[...]

PR #1557, opened 2026-06-07
test(mcp): live end-to-end FalkorDB suite + CI workflow (gpt-5.5)

danielchalef:
Live end-to-end test suite for the Graphiti MCP server + a CI workflow to run it.

> **Stacked on #1553** (base = `feat/mcp-core-parity`) so the suite exercises the full parity tool surface against the real default model.

## Suite — `mcp_server/tests/test_live_falkordb_int.py`
Spawns the MCP server over **stdio** → **FalkorDB** (default DB) + a real OpenAI model, and runs end to end:
`add_memory` → wait for async processing → **`search_nodes`** + **`search_memory_facts`** (asserts real extracted nodes/facts) → `get_status` → `delete_episode` → `clear_graph`. A second test asserts the **full parity tool set** is registered (add_memory, search_nodes, search_memory_facts, get_episodes, get_entity_edge, delete_entity_edge, delete_episode, clear_graph, get_status, summarize_saga, build_communities, add_triplet, get_episode_entities).

- **Self-skipping** unless `OPENAI_API_KEY` is set **and** FalkorDB is reachable — a no-op locally without setup and on fork PRs (no secrets).
- Local runs load the project-root `.env`; alphanumeric test `group_id` keeps the RediSearch fulltext filter valid (backend-portable).
- **Verified locally against FalkorDB + `gpt-5.5`: 2 passed in ~21s.**

## Workflow — `.github/workflows/mcp-server-tests.yml`
- Triggers on PRs/pushes to `main` filtered to `mcp_server/**` (+ the workflow file).
- FalkorDB on the host network (depot bridge-NAT workaround); `OPENAI_API_KEY` from the **`development`** environment.
- **`MODEL_NAME=gpt-5.5`** (the server default) and **`SEMAPHORE_LIMIT=20`** to keep the live episode processing fast.

## This suite already earned its keep
On the first `gpt-5.5` run it surfaced a real bug in the parity branch: the rich entity models (`Location`/`Event`/`Object`/`Topic`/`Person`/`Organization`) declared a `name` field, which graphiti-core reserves (`EntityNode.name`), so **every episode ingest failed silently** with the default config. Fixed on #1553 (drop the reserved field) plus a unit guard; this suite now passes.

### DIFF
--- a/.github/workflows/mcp-server-tests.yml
+++ b/.github/workflows/mcp-server-tests.yml
@@ -20,10 +20,18 @@ on:
 permissions:
   contents: read
 
+# Cancel superseded runs on the same ref so repeated pushes don't keep paying for
+# real OpenAI calls.
+concurrency:
+  group: mcp-server-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   live-mcp-tests:
     runs-on: depot-ubuntu-22.04
     environment: development
+    # Insurance against a hung stdio subprocess / uv sync.
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Set up Python
--- a/.github/workflows/mcp-server-tests.yml
+++ b/.github/workflows/mcp-server-tests.yml
@@ -64,8 +64,10 @@ jobs:
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           FALKORDB_URI: redis://localhost:6379
-          # Exercise the server's default model: gpt-5.5 (reasoning effort 'none').
-          MODEL_NAME: gpt-5.5
+          # gpt-5.5 is the server's runtime default, but it requires a key/account
+          # with fast gpt-5.5 access; CI pins a lighter, broadly-available model so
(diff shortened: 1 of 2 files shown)


danielchalef:
Addressed the latest review (`a943685`):

- **Surface errors:** `wait_for_episodes` / `search_until` now raise immediately on an `{'error': …}` tool response instead of masking it as empty (no more cause-hiding 180s timeouts).
- **Cleanup on failure:** wrapped the e2e in `try/finally` so `clear_graph` always runs.
- **Workflow hygiene:** added a `concurrency` group (cancel superseded runs → no wasted OpenAI calls) and a 20-min job `timeout-minutes`.
- **Docstring:** corrected (CI uses `gpt-5.5`, the server default — not a cost pin).
- Re-verified live against FalkorDB + gpt-5.5 (2 passed in ~19s).

**On the trigger warning:** kept `pull_request: branches: [main]` intentionally — this gates **main** per the original request. Because this PR is stacked on `feat/mcp-core-parity`, the workflow won't run here (it would on a PR into `main`), and the parity-tool assertions can't pass on `main` until #1553 lands anyway. The suite is verified locally; the CI gate activates once the stack reaches `main` (merge order: #1552 → #1553 → #1557). I didn't hardcode the ephemeral feature branch into the trigger to avoid leftover cruft.

PR #1560, opened 2026-06-08
ci(server): live end-to-end regression test + PR check

danielchalef:
## Summary

Adds a live end-to-end regression check for the `server/` REST API (`graph_service`), mirroring the existing `mcp-server-tests.yml` workflow.

**`server/tests/test_live_falkordb_int.py`** — spawns the FastAPI server as a uvicorn subprocess backed by FalkorDB + a real OpenAI model, and exercises the public API end to end:

- `POST /messages` (async ingest) → poll `GET /episodes/{group_id}` until the episode is persisted → `POST /search` and assert at least one fact was extracted → `DELETE /group/{group_id}` cleanup.
- Plus a `GET /healthcheck` smoke test.
- Self-skips (module-level) when `OPENAI_API_KEY` is unset **or** FalkorDB is unreachable, so it's a no-op on fork PRs and on local checkouts without setup.

**`.github/workflows/server-tests.yml`** — mirrors `mcp-server-tests.yml`:

- Triggers on PRs/pushes to `main` touching `server/**` or the workflow file.
- Depot runner, FalkorDB on the host network (bridge-NAT workaround), `development` environment for the OpenAI key.
- Pins `MODEL_NAME=gpt-4.1-mini` and `SEMAPHORE_LIMIT=20` for speed/reliability.
- Treats pytest exit code 5 (no tests collected — the self-skip case on fork PRs) as success; real failures still fail the job.

**`server/pyproject.toml`** — registers the `integration` pytest marker and adds `graphiti-core[falkordb]` to the **dev** dependencies. The server already branches on `db_backend='falkordb'` but never installed the driver; this is test-only so production installs stay lean.

## Validation

Run locally against a real FalkorDB container + live OpenAI:

- Live run: **2 passed in ~16s** (ingest → process → search → delete).
- Self-skip path (DB unreachable / no key): pytest exits **5**, which the workflow maps to success.

## Note for reviewers

The server advertises a FalkorDB backend (`db_backend='falkordb'`) but its **production** deps don't include the driver — a `db_backend=falkordb` deployment hits `ImportError: falkordb is required` today. This PR only adds the driver to dev deps (for the test). Whether to also add `graphiti-core[falkordb]` to the server's runtime dependencies is a separate call worth making.

### DIFF
--- a/.github/workflows/server-tests.yml
+++ b/.github/workflows/server-tests.yml
@@ -1,0 +1,89 @@
+name: Server Tests
+
+# Live end-to-end regression tests for the Graphiti graph_service (REST API)
+# against FalkorDB and a real OpenAI model. Spawns the FastAPI server as a
+# subprocess and exercises the public API end to end (ingest -> async process ->
+# search -> delete). Runs on PRs and pushes to main that touch the server. The
+# OpenAI key comes from the `development` GitHub environment; fork PRs receive no
+# secrets, so the suite skips itself there (the test module skips when
+# OPENAI_API_KEY is unset).
+on:
+  push:
+    branches: [main]
+    paths:
+      - "server/**"
+      - ".github/workflows/server-tests.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "server/**"
+      - ".github/workflows/server-tests.yml"
+
+permissions:
+  contents: read
+
+# Cancel superseded runs on the same ref so repeated pushes don't keep paying for
+# real OpenAI calls.
+concurrency:
+  group: server-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  live-server-tests:
+    runs-on: depot-ubuntu-22.04
+    environment: development
+    # Insurance against a hung server subprocess / uv sync.
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - name: Set up Python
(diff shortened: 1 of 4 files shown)


danielchalef:
Thanks for the review — addressed the actionable findings in 9efd21e:

- **Exit-5-can-mask-broken-selection (warning):** the exit-5→success path is now gated on `OPENAI_API_KEY` being *absent*. Since FalkorDB is gated by the prior "Wait for FalkorDB" step, a same-repo run with the key present can only hit exit 5 via a broken test selection (renamed marker/file) — that now **fails loudly**. Exit 5 is treated as success only in the fork self-skip case (no key).
- **`_unique_token()` 1-second granularity (suggestion):** switched to `uuid.uuid4().hex`, so per-run graph isolation holds even within the same second / under `pytest-xdist`.
- **`_wait_for_episodes` swallows non-200 (suggestion):** now fails fast on a 5xx (the server is already up past `/healthcheck`, so a 5xx is a real error) and reports the last non-200 body if it ultimately times out — no more opaque 180s generic failures.

Not actioned, by design:
- **Leftover empty FalkorDB graph:** would require coupling the test directly to FalkorDB (`GRAPH.DELETE`) in teardown just to drop *empty* graphs on a long-lived local instance — harmless in ephemeral CI, so left as-is.
- **README FalkorDB-backend caveat / runtime-deps:** this is the separate production-deps decision flagged in the PR body; tracking it outside this PR.

Re-validated locally against FalkorDB + live OpenAI: 2 passed in ~16s; self-skip path still exits 5.

PR #1567, opened 2026-06-08
ci: harden exit-5 handling in the live-test workflows (MCP backport + diagnostics)

danielchalef:
## Summary

Hardens the exit-code-5 handling in the live-test workflows.

1. **Backports the fail-loudly fix to `mcp-server-tests.yml`** (the server got it in #1560). The exit-5→success path is now gated on `OPENAI_API_KEY` being absent, so a fork PR's module-level self-skip stays green while a same-repo run that unexpectedly collects zero tests fails instead of reporting a green build that tested nothing.
2. **Clarifies the failure diagnostic in both workflows** (review follow-up).

## Exit-5 causes (corrected)

`pytest` exits 5 when no tests are collected. With the key present (same-repo run), that's unexpected and the job now fails loudly. It can be reached by either:

- **FalkorDB became unreachable** — the suite has a *second* module-level skip via a host-side TCP check to `localhost:6379`, distinct from the workflow's in-container `redis-cli ping` readiness gate. Normally equivalent under `--network host`, but not the same check.
- **The `integration` marker was renamed/removed** — `-m integration` then matches nothing.

(A renamed/removed **test file** is a pytest *usage error* → exit **4**, not 5 — still caught by `exit "$code"`. The earlier description conflated this with the marker case; only the marker case routes through the exit-5 branch.)

| Scenario | Before | After |
|----------|--------|-------|
| Fork PR (no key) → module self-skip → exit 5 | success | success |
| Same-repo, key present, exit 5 (FalkorDB down **or** marker renamed) | silent green | **fail** |
| Renamed test-file path (exit 4) / import error (exit 2) / real failure (exit 1) | fail | fail |

Workflow-only — no test or behavior change.

### DIFF
--- a/.github/workflows/mcp-server-tests.yml
+++ b/.github/workflows/mcp-server-tests.yml
@@ -72,11 +72,13 @@ jobs:
           SEMAPHORE_LIMIT: "20"
         # pytest exits 5 when no tests are collected. The legitimate case is a fork
         # PR with no OPENAI_API_KEY: the suite self-skips at module level. When the
-        # key IS present (same-repo run), FalkorDB is already up — the prior step
-        # gates on it — so exit 5 instead means a broken test selection (the
-        # `integration` marker or the test file was renamed) and must fail loudly,
-        # rather than reporting a green build that tested nothing. Import errors
-        # (exit 2) and real test failures (exit 1) always fail the job.
+        # key IS present (same-repo run) exit 5 is unexpected — the live suite should
+        # have run — so fail loudly rather than report a green build that tested
+        # nothing. Likely causes: FalkorDB became unreachable (the suite also skips at
+        # module level when it can't connect, even though the prior step gated it), or
+        # the `integration` marker was renamed/removed so the selection matches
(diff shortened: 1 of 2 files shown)


danielchalef:
Thanks — both findings are correct and addressed in 850225f (message/comment only, no behavior change):

- **Finding 1 (incomplete diagnostic):** right that the key-present exit-5 path is also reachable when FalkorDB goes unreachable (the suite's second module-level skip via the host-side TCP check, distinct from the in-container `redis-cli ping` gate). Broadened the failure message and comment to name both causes (FalkorDB-unreachable / renamed marker) instead of asserting "broken test selection."
- **Finding 2 (exit 4 vs 5):** correct — a renamed/removed *file* is a usage error (exit 4, separately caught), and only a renamed/removed `integration` *marker* produces exit 5. Stopped conflating the two in the comment, and corrected the PR description's "can only" claim + table.

Applied the same wording to `server-tests.yml` as well (it shipped with the original message in #1560) so the two workflows stay consistent. The *behavior* was already right — an unexpected key-present exit 5 fails loudly regardless of cause — this is purely diagnostic polish.

PR #1326, opened 2026-03-15
fix: route FalkorDB queries to correct graph for single group_id

spencer2211:
Fixes #1325

## Summary

The `handle_multiple_group_ids` decorator only switched the FalkorDB driver to the correct graph database when `len(group_ids) > 1`. For a single `group_id` (the most common case), queries fell through to the default execution path, hitting `default_db` which typically has no data. This caused silent empty results for all FalkorDB deployments using per-query group_ids.

Additionally, all `get_by_group_ids` classmethods on node and edge types bypassed the decorator entirely, querying the driver directly without routing to the correct FalkorDB graph.

### `decorators.py`
- Change `len(group_ids) > 1` to `len(group_ids) >= 1` so the decorator activates for any non-empty group_ids list

### `nodes.py` (4 methods)
- Add FalkorDB graph routing to `EpisodicNode.get_by_group_ids`, `EntityNode.get_by_group_ids`, `CommunityNode.get_by_group_ids`, `SagaNode.get_by_group_ids`

### `edges.py` (5 methods)
- Add FalkorDB graph routing to `EpisodicEdge.get_by_group_ids`, `EntityEdge.get_by_group_ids`, `CommunityEdge.get_by_group_ids`, `HasEpisodeEdge.get_by_group_ids`, `NextEpisodeEdge.get_by_group_ids`

## Approach

For each `get_by_group_ids` method:
- **Single group_id**: Clone the driver to target the group's graph database (`driver.clone(database=group_ids[0])`), then fall through to the existing query logic
- **Multiple group_ids**: Query each graph separately via recursive call with a single group_id, merge results, re-sort by UUID descending, and apply limit

This approach avoids code duplication — the single-group case reuses the existing query path, and the multi-group case decomposes cleanly into single-group calls.

## Context

FalkorDB uses separate graph databases per `group_id`, unlike Neo4j which stores all data in a single database and filters by `group_id` as a node/edge property. The decorator and classmethods need to explicitly switch the driver to the correct graph database before executing queries.

Discovered and verified against a production FalkorDB deployment running graphiti-core with the graphiti MCP server.

PR #1572, opened 2026-06-10
fix(falkordb): route single-group reads onto the group's graph

yimtheppariyapol:
## Summary
Single-group reads against FalkorDB return nothing even when the data is physically present in the group's graph. `handle_multiple_group_ids` only re-routed the driver onto the group graph when `len(group_ids) > 1`, so a single-group read fell through to the base driver (pinned at `default_db`) and queried the wrong graph.

## Root cause
FalkorDB stores each `group_id` in its own graph, so a read must clone the driver onto that graph. The decorator gated that routing on `len(group_ids) > 1`. The write path mutates `self.driver` onto the group graph, which is why same-process write-then-read tests masked the bug — the reader inherited the mutated driver. A fresh reader process with a single `group_id` queried `default_db` and got 0 results.

Reproduced: `GRAPH.QUERY <group>` showed episodes + facts present; a single-group `get_episodes([group])` returned 0; a two-group `[group, other]` search returned everything (because `len > 1` routed).

## What changed
- **`decorators.py`** — replace the `len > 1` gate with a blank-filtered `routable_group_ids` (mirrors `search()`'s `!= ['']` convention). A single real group now routes; a `['']`/`[None]` group no longer routes onto an empty graph name.
- **`falkordb_driver.py`** — memoize `clone()` per database. `__init__` schedules `build_indices_and_constraints()` on every construction — load-bearing exactly once per group, but pure churn on every subsequent read now that single-group reads clone. Caching keeps the once-per-group index build while removing the per-call constructor cost.

## Verification
- Single-group `get_episodes([group])` now returns the group's facts against a live FalkorDB (previously 0).
- Memoized `clone()`: repeated clone of the same db returns the same object (no churn); `default_group_id` maps to `default_db`; self-clone returns self; a new group is constructed once then cached, and the new group's graph still gets its indices built.

## Follow-up (not in this PR)
The write-path `self.driver` mutation in `graphiti.py` could be made non-mutating so reads and writes share one routing mechanism, removing the asymmetry that masked this bug.

### DIFF
--- a/graphiti_core/decorators.py
+++ b/graphiti_core/decorators.py
@@ -44,13 +44,18 @@ def handle_multiple_group_ids(func: F) -> F:
         if group_ids is None and group_ids_pos is not None and len(args) > group_ids_pos:
             group_ids = args[group_ids_pos]
 
-        # Only handle FalkorDB with multiple group_ids
+        # FalkorDB stores each group_id in its own graph, so reads must clone the
+        # driver onto the group's graph. This applies to a single group_id too: the
+        # base driver points at default_db, so an unrouted single-group read finds
+        # nothing (writes mutate self.driver and masked this in same-process tests).
+        # Blank/None entries mean "unspecified" (mirrors search()'s `!= ['']` rule),
+        # so filter them out rather than routing onto an empty graph name.
+        routable_group_ids = [gid for gid in group_ids if gid] if group_ids else []
         if (
             hasattr(self, 'clients')
             and hasattr(self.clients, 'driver')
             and self.clients.driver.provider == GraphProvider.FALKORDB
-            and group_ids
-            and len(group_ids) > 1
+            and routable_group_ids
         ):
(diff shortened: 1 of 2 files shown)


yimtheppariyapol:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: [tiramitsu.88@gmail.com](mailto:tiramitsu.88@gmail.com)

yimtheppariyapol:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: tiramitsu.88@gmail.com

Naseem77:
@yimtheppariyapol I tested this and single-group get_by_group_ids still returns 0 after the patch. #1326 solves more of this, might be worth joining efforts there and closing this one.

prasmussen15:
Closed in favor of: https://github.com/getzep/graphiti/pull/1670

PR #1575, opened 2026-06-11
fix(mcp): keep strong reference to episode queue worker task to prevent silent GC

### LINKED ISSUE #1574: MCP server: queue worker silently garbage-collected under streamable-http — add_memory queues but never processes
# Issue: Queue worker silently garbage-collected under streamable-http transport — `add_memory` queues but never processes

## Summary

`add_memory` returns `"Episode 'X' queued for processing"` (200 OK) but the queue worker never starts processing under the **streamable-http transport**. No LLM calls happen, no episodes appear in the graph database, no error logs are produced. `search_nodes`, `get_episodes` and other read tools work correctly because they read pre-existing data.

Root cause: `services/queue_service.py:add_episode_task()` calls `asyncio.create_task(...)` without keeping a reference to the returned task. Per Python asyncio documentation:

> *"Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps **weak references** to tasks. A task that isn't referenced elsewhere may be garbage collected at any time, even before it's done."*
> — [docs.python.org/3/library/asyncio-task.html#asyncio.create_task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)

Under streamable-http transport, GC pressure during request handling is high enough that the orphaned worker task gets collected before its first `await` on the queue completes.

## Reproduction

> Re-verified against `main` on 2026-06-11: `services/queue_service.py` still discards the `asyncio.create_task` result (only a `_queue_workers: dict[str, bool]` flag is kept).

**Environment:**
- `zepai/knowledge-graph-mcp:1.0.2-standalone` (graphiti-core 0.28.2, mcp 1.26.0)
- Transport: `streamable-http`
- Backend: FalkorDB (also reproducible with Neo4j per fork reports)
- LLM gateway: any OpenAI-compatible endpoint
- Kubernetes Deployment, single replica, 256Mi memory limit

**Steps:**
1. Start MCP server with `transport: http` (streamable-http)
2. Connect MCP client (e.g., `curl` or Claude Code via mcp-remote)
3. Send `tools/call` for `add_memory` with valid `name` + `episode_body`
4. Observe response: `"Episode '<name>' queued for processing"` (success, 200 OK)
5. Wait 60 seconds and observe pod logs

**Expected:**
- `INFO services.queue_service - Starting episode queue worker for group_id: <gid>`
- `INFO services.queue_service - Processing episode <uuid> for group <gid>`
- HTTP requests to LLM/embedder
- `INFO graphiti_core.graphiti - Completed add_episode in N ms`
- New `Episodic` node in graph database

**Actual:**
- No log output between request and a hypothetical 60-minute mark
- No HTTP requests to LLM/embedder
- No new `Episodic` node in graph database
- No error in logs (silent failure)

## Diagnosis

The worker task created at `queue_service.py:45`:

```python
if not self._queue_workers.get(group_id, False):
    asyncio.create_task(self._process_episode_queue(group_id))  # ← result discarded
```

…is discarded. asyncio's event loop only keeps **weak** references to tasks. Under streamable-http transport, when a request handler completes and triggers garbage collection, the orphaned task is collected before its first `await` on the empty queue can suspend it.

Verification: we instrumented the running pod by replacing `queue_service.py` with a patched version that stores the task reference in `self._worker_tasks`. After the patch:

```
12:45:09 - Starting episode queue worker for group_id: kiroot_v2
12:45:09 - Processing episode None for group kiroot_v2
12:45:13 - HTTP Request: POST .../v1/responses "HTTP/1.1 200 OK"
12:45:14 - HTTP Request: POST .../v1/embeddings "HTTP/1.1 200 OK" (×6)
12:46:01 - Completed add_episode in 51522.60 ms
12:46:01 - Successfully processed episode None for group kiroot_v2
```

FalkorDB Episodic count went from 569 → 570 with the test episode `"GC-Anchor-Patch Verification 2026-05-10"`.

## Proposed fix

A 6-line, behaviorally-neutral change to `services/queue_service.py`: store the worker task reference in a dict so it's strongly held until it exits.

A PR with the 6-line fix follows.

## References

- [asyncio.create_task documentation warning](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)
- Related FastMCP issue with similar root cause (different framework but same anti-pattern): [PrefectHQ/fastmcp#1115](https://github.com/PrefectHQ/fastmcp/issues/1115)
- Several community forks worked around this by switching to Redis-backed persistent queues:
  - [michabbb/graphiti-mcp-but-working](https://github.com/michabbb/graphiti-mcp-but-working) — BRPOPLPUSH pattern
  - [itcook/graphiti-mcp-pro](https://github.com/itcook/graphiti-mcp-pro) — async parallel processing rewrite

The fix proposed here is much smaller (6 lines) and preserves the existing architecture while eliminating the GC race.

rneissl:
# PR: Fix queue worker garbage-collection under streamable-http transport

## What

Stores a strong reference to the asyncio.Task created by `QueueService.add_episode_task` so the event loop can't garbage-collect the worker mid-execution.

## Why

Per [Python asyncio docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task):

> *"Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks."*

Without anchoring, under high-GC-pressure conditions (streamable-http request handling) the worker task can be collected before its first `await self._episode_queues[group_id].get()` suspends it. Result: `add_memory` queues episodes but no worker ever processes them — silent failure, no error logs.

Reported in #1574. Fixes #1574.

## How

Three small changes in `mcp_server/src/services/queue_service.py`:

1. **`__init__`**: add `self._worker_tasks: dict[str, asyncio.Task] = {}` — strong-ref storage
2. **`add_episode_task`**: store the task: `self._worker_tasks[group_id] = asyncio.create_task(...)`
3. **`_process_episode_queue` finally-block**: `self._worker_tasks.pop(group_id, None)` — clean up reference when worker exits

## Verification

Tested on `zepai/knowledge-graph-mcp:1.0.2-standalone` with FalkorDB backend, streamable-http transport, in a Kubernetes Deployment.

**Before patch:** `add_memory` returns "queued" but no log activity, no LLM calls, no new graph nodes, no error.

**After patch:**

```
12:45:09 INFO services.queue_service - Starting episode queue worker for group_id: kiroot_v2
12:45:09 INFO services.queue_service - Processing episode None for group kiroot_v2
12:45:13 INFO httpx - HTTP Request: POST .../v1/responses "HTTP/1.1 200 OK"
12:45:14 INFO httpx - HTTP Request: POST .../v1/embeddings "HTTP/1.1 200 OK" (×6)
[entity-extraction, dedup, embedding, edge-resolution loop continues for ~51s]
12:46:01 INFO graphiti_core.graphiti - Completed add_episode in 51522.60 ms
12:46:01 INFO services.queue_service - Successfully processed episode None for group kiroot_v2
```

Graph: Episodic node count went from 569 → 570, new entities and edges extracted from the test episode.

## Backwards compatibility

None broken. Public API of `QueueService` unchanged. Only adds an internal `_worker_tasks` dict and stores/cleans up references inside existing methods. Behavior under stdio transport is identical to before (the GC race didn't manifest there because handler completion didn't trigger the same GC pattern).

## Tests

The MCP server doesn't ship asyncio-stress tests for the queue worker today. Adding a regression test would require simulating GC pressure, which is non-trivial. Manual reproduction steps are in the linked issue. Happy to add a test if maintainers point at a similar testing pattern in the repo.

---

Patch developed and verified by Roland Neissl (BAB IKT, Austria) while migrating the graphiti-mcp service from the ToolHive operator to a plain GitOps deployment. Originally verified on `zepai/knowledge-graph-mcp:1.0.2-standalone` (2026-05-10), re-based and re-verified against `main` on 2026-06-11.


Naseem77:
@rneissl Checked this against the official asyncio docs: create_task results must stay referenced or the task can be GC'd mid-run, which is exactly what the queue worker risked. Your change anchors and releases the refs correctly. Zero-risk hardening.

prasmussen15 (changes requested):

PR #1578, opened 2026-06-12
fix(extraction): plumb custom_extraction_instructions into entity summary prompts

maxziro:
## Problem

`custom_extraction_instructions` is honored by the node and edge extraction prompts, but it never reaches the entity **summary** stage: `extract_attributes_from_nodes` and its summary helpers have no way to receive it, so the `extract_summaries_batch` / `extract_entity_summaries_from_episodes` prompts always run unconstrained.

Practical effect (how we hit this): with instructions like *"Perform all extraction in Italian"*, entity names and edge facts come back in Italian as expected, while freshly **composed** entity summaries come back in English — even when the source episode is entirely Italian. Summaries that are produced by the fact-appending shortcut look fine (the facts are already constrained), which makes the gap surface only intermittently, on summaries the LLM actually composes.

## Change

Thread the existing parameter through the summary path, mirroring how the extraction prompts already handle it:

- `extract_attributes_from_nodes(..., custom_extraction_instructions=None)` → `_extract_entity_summaries_batch` → `_process_summary_flight` → prompt context (`''` when unset).
- Interpolate `{context['custom_extraction_instructions']}` in both summary prompt templates (`extract_summaries_batch` and `extract_entity_summaries_from_episodes`), same bare-line pattern as `extract_message` / `extract_json` / `extract_text`.
- Pass the parameter at both call sites: `add_episode` and the bulk path (`_resolve_nodes_and_edges_bulk`, called by `add_episode_bulk`).
- Docstrings updated accordingly.

No behavior change when the parameter is unset (empty string interpolation, like the other prompts). The Go-mirrored system prompt (`_entity_episode_summary_system_prompt`) is intentionally untouched — the injection lands in the user message.

## Tests

Three unit tests in `tests/utils/maintenance/test_node_operations.py`, mirroring the existing custom-instructions plumbing tests in `test_bulk_utils.py`:

- batch summary prompt receives the instructions (long-summary LLM route),
- episode-prompt path (`skip_fact_appending=True`) receives them too,
- `extract_attributes_from_nodes` passes them through to the batch helper.

`tests/utils/maintenance/test_node_operations.py`: 36 passed (33 existing + 3 new).

### DIFF
--- a/graphiti_core/graphiti.py
+++ b/graphiti_core/graphiti.py
@@ -821,6 +821,7 @@ class Graphiti:
         edge_types: dict[str, type[BaseModel]] | None,
         edge_type_map: dict[tuple[str, str], list[str]],
         episodes: list[EpisodicNode],
+        custom_extraction_instructions: str | None = None,
     ) -> tuple[list[EntityNode], list[EntityEdge], list[EntityEdge], dict[str, str]]:
         """Resolve nodes and edges against the existing graph."""
         nodes_by_uuid: dict[str, EntityNode] = {
@@ -880,6 +881,7 @@ class Graphiti:
                     episode,
                     previous_episodes,
                     entity_types,
+                    custom_extraction_instructions=custom_extraction_instructions,
                 )
                 for episode, previous_episodes in episode_context
             ]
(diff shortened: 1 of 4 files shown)


maxziro:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 12826347+maxziro@users.noreply.github.com

PR #1581, opened 2026-06-14
fix(search): parenthesize group_id OR-chain in multi-group fulltext queries

Saltasm:
## Problem

The Neo4j fulltext query builders join group filters with `OR`, then append the query terms with `AND`, without parenthesizing the OR-chain:

```
group_id:"a" OR group_id:"b" OR group_id:"c" AND (terms)
```

Lucene's classic query parser treats `AND` as promoting its adjacent clauses to `MUST`, so this parses as:

| clause | occur |
|---|---|
| `group_id:"a"` | SHOULD |
| `group_id:"b"` | SHOULD |
| `group_id:"c"` | **MUST** (promoted by the following `AND`) |
| `(terms)` | **MUST** |

A `BooleanQuery` with any `MUST` clause requires only the `MUST` clauses, so this effectively becomes `(group_id:"c" AND terms)`: the leading groups become scoring-only and the query terms are scoped to just the **last** `group_id`. A multi-group fulltext search therefore silently returns matches from only the last group passed (and zero rows when that group is empty), while the semantic leg of hybrid search masks the regression.

## Fix

Wrap the OR-chain in parens so the terms apply across all groups:

```
(group_id:"a" OR group_id:"b" OR group_id:"c") AND (terms)
```

Applied to both Neo4j builders that have the issue:
- `graphiti_core/search/search_utils.py` — `fulltext_query` (neo4j branch)
- `graphiti_core/driver/neo4j/operations/search_ops.py` — `_build_neo4j_fulltext_query`

Single-group queries are unaffected (`group_id:"a" AND (terms)` already parses correctly). FalkorDB and Kuzu builders are unaffected — they already parenthesize / use different syntax.

## Verification

With the parenthesized query, an identical multi-group search returns the same cross-group result set regardless of the order group_ids are passed; without it, results collapse to whichever group_id is passed last.


Saltasm:
@danielchalef gentle ping when you have a moment. Small fix (+10/-2): in multi-group fulltext search the `group_id` OR-chain isn't parenthesized, so Lucene's AND > OR precedence binds the query terms to only the *last* `group_id` — a multi-group search silently collapses to that single group (and zeroes out bm25 episode search when the last group is empty). `ruff` is green; the red `triage` check is the fork-secrets automation, not the diff. Happy to rebase onto latest `main` if that's useful. Thanks!

Saltasm:
Updated to current `main`. The fix is unchanged (+10/−2) and sits cleanly under the Kuzu/FalkorDB refactor of `fulltext_query` — the refactor didn't touch the group-filter join, so the bug is still present at HEAD: the OR-chain is joined unparenthesized before ` AND ` in both `search_utils.fulltext_query` and `neo4j/operations/search_ops._build_neo4j_fulltext_query`, so a multi-group fulltext search still collapses to the last group passed.

CI on the updated head: CLA / ruff / check-fork green. The remaining `triage` failure is the labeling workflow's token permissions on fork PRs, not this change.

PR #1598, opened 2026-06-18
fix(cross-encoder): don't crash OpenAI reranker when a response lacks logprobs

ly-wang19:
## Summary
`OpenAIRerankerClient.rank()` crashes with `ValueError: zip() argument 2 is shorter than argument 1` when any passage's chat-completion response has no usable `top_logprobs`. It collected scores in a separate list, `continue`-ing past empty-logprob responses, then re-paired them with `zip(passages, scores, strict=True)` — so a single skipped response desyncs the two lists and the strict zip raises, failing the **entire** rerank instead of just dropping the one unscoreable passage.

## Type of Change
- [x] Bug fix

## Objective
A single empty/truncated reranker response (possible with `max_tokens=1` + `logit_bias`) shouldn't take down the whole reranking. Pair each passage with its own response while iterating so the lists can't desync — keeping the existing "skip the unscoreable passage" behaviour, without the crash.

## Testing
- [x] All existing tests pass

Added `tests/cross_encoder/test_openai_reranker_client.py` (the OpenAI reranker was previously untested): one test feeds an empty-logprobs response in the middle and asserts the other passages still rank (fails before the change with the strict-zip `ValueError`, passes after); one covers the normal all-scored path. `ruff` clean.

## Breaking Changes

## Checklist
- [x] Code follows project style guidelines (`ruff` passes)
- [x] Self-review completed

### DIFF
--- a/graphiti_core/cross_encoder/openai_reranker_client.py
+++ b/graphiti_core/cross_encoder/openai_reranker_client.py
@@ -103,17 +103,22 @@ class OpenAIRerankerClient(CrossEncoderClient):
                 else []
                 for response in responses
             ]
-            scores: list[float] = []
-            for top_logprobs in responses_top_logprobs:
+            results: list[tuple[str, float]] = []
+            for passage, top_logprobs in zip(passages, responses_top_logprobs, strict=True):
                 if len(top_logprobs) == 0:
+                    # No usable logprobs for this passage (e.g. an empty or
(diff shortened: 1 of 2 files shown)


ly-wang19:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: wly999666@gmail.com

ly-wang19:

ly-wang19:
recheck

PR #764, opened 2025-07-24
Fix: Respect LLMConfig.max_tokens in OpenAI-based clients

paveljakov:
## Summary
Fixes issue where LLMConfig.max_tokens was ignored, causing 'Output length exceeded max tokens 8192' errors even when higher token limits were configured.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
- Fix BaseOpenAIClient to respect config.max_tokens when explicitly set
- Prevents max_tokens from being overridden by constructor defaults
- Affects OpenAIClient and AzureOpenAILLMClient
- Add comprehensive unit tests to verify the fix
- Maintains backward compatibility for existing code

## Testing
- [x] Unit tests added/updated
- [ ] Integration tests added/updated
- [ ] All existing tests pass

## Breaking Changes
- [ ] This PR contains breaking changes

## Checklist
- [x] Code follows project style guidelines (`make lint` passes)
- [x] Self-review completed
- [ ] Documentation updated where necessary
- [x] No secrets or sensitive information committed

## Related Issues
Closes #763

----

> [!IMPORTANT]
> Fixes `BaseOpenAIClient` to respect `LLMConfig.max_tokens`, adding unit tests to verify behavior.
> 
>   - **Behavior**:
>     - Fixes `BaseOpenAIClient` to respect `config.max_tokens` when set, preventing it from being overridden by constructor defaults.
>     - Affects `OpenAIClient` and `AzureOpenAILLMClient`.
>   - **Testing**:
>     - Adds unit tests in `test_openai_client.py` to verify `max_tokens` precedence and behavior.
>     - Tests include scenarios for custom `max_tokens`, default values, and precedence over constructor parameters.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=getzep%2Fgraphiti&utm_source=github&utm_medium=referral)<sup> for 3e4be37b6fb06f50569202f06dcf2646851c5ca4. You can [customize](https://app.ellipsis.dev/getzep/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #1604, opened 2026-06-21
fix: respect LLMConfig.max_tokens in OpenAIGenericClient (#763)

### LINKED ISSUE #763: [BUG] LLMConfig.max_tokens not respected by OpenAI-based clients
# LLMConfig.max_tokens not respected by OpenAI-based clients

## Bug Description

When configuring `max_tokens` in `LLMConfig`, the setting is ignored by OpenAI-based clients (`OpenAIClient` and `AzureOpenAILLMClient`), causing token limit errors even when a higher limit is configured.

**Additionally, there are currently two confusing ways to set max_tokens for OpenAI clients:**
1. Through `LLMConfig(max_tokens=...)`  
2. Through the client constructor parameter `OpenAIClient(max_tokens=...)`

This dual configuration is confusing because:
- It's unclear which one takes precedence
- Users naturally expect the config object to be the authoritative source
- The constructor parameter silently overrides the config setting
- Other clients (Gemini, Anthropic) don't have this confusing dual interface

## What I was trying to do

Configure a higher token limit for LLM responses to avoid truncation:

```python
from graphiti_core.llm_client.config import LLMConfig
from graphiti_core.llm_client.openai_client import OpenAIClient

config = LLMConfig(
    api_key="...",
    model="gpt-4",
    max_tokens=16000  # Set higher limit
)

client = OpenAIClient(config=config)
```

## What I expected to happen

The client should respect the `max_tokens=16000` setting and allow responses up to 16,000 tokens, similar to how other configuration options work.

## What actually happened

The client still uses the default 8192 token limit, resulting in errors like: Output length exceeded max tokens 8192: Could not parse response content as the length limit was reached

## Root Cause

In `BaseOpenAIClient.__init__()`, the constructor parameter `max_tokens=DEFAULT_MAX_TOKENS` always overrides `config.max_tokens`, even when explicitly set by the user:

```python
# Current problematic code:
def __init__(self, config: LLMConfig | None = None, cache: bool = False, max_tokens: int = DEFAULT_MAX_TOKENS):
    # ...
    super().__init__(config, cache)
    self.max_tokens = max_tokens  # This overwrites config.max_tokens!
```

## Code sample that demonstrates the issue

```python
from graphiti_core.llm_client.config import LLMConfig
from graphiti_core.llm_client.openai_client import OpenAIClient

config = LLMConfig(max_tokens=16000)
client = OpenAIClient(config=config)

print(f"Config max_tokens: {config.max_tokens}")  # Shows: 16000
print(f"Client max_tokens: {client.max_tokens}")  # Shows: 8192 (wrong!)

# User expects: client.max_tokens == 16000
# Actual result: client.max_tokens == 8192
```

## Affected Clients

- `OpenAIClient`
- `AzureOpenAILLMClient`

**Other clients handle this correctly:**
- `GeminiClient` - respects config and has clear precedence rules
- `AnthropicClient` - respects config settings
- `GroqClient` - respects config settings

## Proposed Solution

The fix should:
1. **Respect `config.max_tokens` when explicitly set** (different from default)
2. **Use constructor parameter as fallback** when config has default value
3. **Maintain backward compatibility** for existing code
4. **Clear precedence rules**: Config takes priority over constructor parameter

```python
# Proposed fix:
if config.max_tokens != DEFAULT_MAX_TOKENS:
    self.max_tokens = config.max_tokens  # Use config when explicitly set
else:
    self.max_tokens = max_tokens  # Use constructor parameter as fallback
```

## Environment

- Graphiti version: [current version]
- Python version: 3.12+
- Affected clients: OpenAIClient, AzureOpenAILLMClient

## Labels

- `bug`
- `help wanted`
- `good first issue`
- `llm-client`

Hasnaathussain:
PR #764 fixed `BaseOpenAIClient`, but `OpenAIGenericClient` is a separate implementation that doesn't inherit from it — the fix doesn't reach it.

In `openai_generic_client.py`, the constructor parameter `max_tokens: int = 16384` was unconditionally assigned to `self.max_tokens`, so `LLMConfig.max_tokens` was always ignored. Users of Ollama, vLLM, LM Studio, or any other OpenAI-compatible backend were affected.

Closes #763

---

**Changes:**
- `graphiti_core/llm_client/openai_generic_client.py`: change `max_tokens` parameter to `int | None = None`; resolve as `constructor arg > config.max_tokens`
- `tests/llm_client/test_openai_generic_client.py`: three regression tests covering config respected, constructor override, and default case

**To verify:**
```python
from graphiti_core.llm_client.config import LLMConfig
from graphiti_core.llm_client.openai_generic_client import OpenAIGenericClient

client = OpenAIGenericClient(config=LLMConfig(max_tokens=32000))
assert client.max_tokens == 32000  # was 16384 before this fix
```

### DIFF
--- a/graphiti_core/llm_client/openai_generic_client.py
+++ b/graphiti_core/llm_client/openai_generic_client.py
@@ -62,7 +62,7 @@ class OpenAIGenericClient(LLMClient):
         config: LLMConfig | None = None,
         cache: bool = False,
         client: typing.Any = None,
-        max_tokens: int = 16384,
+        max_tokens: int | None = None,
         structured_output_mode: StructuredOutputMode = 'json_schema',
     ):
         """
@@ -72,7 +72,9 @@ class OpenAIGenericClient(LLMClient):
             config (LLMConfig | None): The configuration for the LLM client, including API key, model, base URL, temperature, and max tokens.
             cache (bool): Whether to use caching for responses. Defaults to False.
             client (Any | None): An optional async client instance to use. If not provided, a new AsyncOpenAI client is created.
-            max_tokens (int): The maximum number of tokens to generate. Defaults to 16384 (16K) for better compatibility with local models.
+            max_tokens (int | None): The maximum number of tokens to generate. When None (the default),
+                the value from ``config.max_tokens`` is used. Pass an explicit integer to override the
+                config value for this instance.
             structured_output_mode (StructuredOutputMode): Whether to request structured
                 output via native ``json_schema`` (the default, uses constrained decoding)
                 or to fall back to ``json_object``. Set to ``'json_object'`` for providers
@@ -90,8 +92,8 @@ class OpenAIGenericClient(LLMClient):
 
         super().__init__(config, cache)
 
-        # Override max_tokens to support higher limits for local models
-        self.max_tokens = max_tokens
+        # Resolve max_tokens: constructor arg takes precedence, then config, then default.
+        self.max_tokens = max_tokens if max_tokens is not None else config.max_tokens
         self.structured_output_mode: StructuredOutputMode = structured_output_mode
 
         if client is None:
--- a/tests/llm_client/test_openai_generic_client.py
+++ b/tests/llm_client/test_openai_generic_client.py
@@ -47,10 +47,15 @@ def _messages() -> list[Message]:
     ]
 
 
-def _make_client(content: str = '{"foo": "bar"}', error: Exception | None = None, **kwargs):
+def _make_client(
+    content: str = '{"foo": "bar"}',
+    error: Exception | None = None,
+    config: 'LLMConfig | None' = None,
+    **kwargs,
+):
     completions = DummyChatCompletions(content=content, error=error)
     client = OpenAIGenericClient(
-        config=LLMConfig(api_key='test', model='test-model'),
+        config=config or LLMConfig(api_key='test', model='test-model'),
         client=DummyClient(completions),
         **kwargs,
     )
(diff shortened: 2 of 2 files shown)


Hasnaathussain:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: hasnaat.hussain.2@gmail.com

Hasnaathussain:
Hi @dchalef / getzep team! Friendly request to review and approve this PR. It resolves the max_tokens constraint issue for OpenAIGenericClient. Let me know if any updates are needed! Thank you!

Hasnaathussain:
Rebased onto current main at signed head 4e8b027; the focused max_tokens regression suite passes 3 tests and Ruff is clean. The triage failure remains external.

Hasnaathussain:
I rebased the PR onto current `upstream/main` and added DCO trailers to both contributor commits. The new head is `4125ad6`; the implementation and tests are unchanged. The branch is now current with the base.

Hasnaathussain:
Upstream advanced after the previous rebase, so I rebased once more onto the latest `upstream/main`. The new signed head is `06a12d3`; the two-file diff is unchanged, the focused max_tokens selection still passes 3 tests, and Ruff is clean.

linhongyu510 (approved):
Verified the current head against `main`. Changing the constructor default to `None` restores the intended precedence (explicit constructor override, then `LLMConfig.max_tokens`) while preserving the existing 16384 effective default through `LLMConfig`. The resolved value continues to flow through `generate_response()` into the OpenAI-compatible request, and the focused tests cover config, explicit override, and default behavior. Local validation: `python3 -m pytest tests/llm_client/test_openai_generic_client.py -q` — **12 passed**; Ruff check and format-check passed on both changed files; `git diff --check HEAD^..HEAD` passed. No blocking issue found.

Hasnaathussain:
@linhongyu510 The approved head is behind main again. Would you like me to rebase it once more, or should I leave the approved commit untouched until the workflow gate is cleared? I have not changed the branch since your approval.

PR #1605, opened 2026-06-21
fix(edges): return reference_time in entity edge read query

ly-wang19:
## Summary
`get_entity_edge_return_query()` omitted `reference_time` from its SELECT, so an `EntityEdge`'s `reference_time` was silently dropped on every read-back even though it is persisted on save.

## Type of Change
- [x] Bug fix

## Objective
`EntityEdge.reference_time` is written by `EntityEdge.save()` on all providers (Neo4j/FalkorDB via `SET e = $edge_data`, Neptune props-preserved, Kuzu explicit column), and the record parsers read it from the top-level record (`record.get('reference_time')`, after popping it out of `attributes`). But `get_entity_edge_return_query()` aliased `created_at`/`expired_at`/`valid_at`/`invalid_at` and not `reference_time`, so every read path (`get_by_uuid`/`get_by_uuids`/`get_between_nodes`/`get_by_node_uuid`/`get_by_group_ids`) returned `reference_time=None`. This adds `e.reference_time AS reference_time` to both branches, matching the sibling temporal aliases.

## Testing
- [x] All existing tests pass

Added `tests/test_edge_db_queries.py`: a parametrized unit test asserting the entity-edge return query selects `reference_time` for every provider. It fails before this change and passes after. `ruff` is clean on the changed files.

## Breaking Changes

## Checklist
- [x] Code follows project style guidelines (ruff passes on changed files)
- [x] Self-review completed


Naseem77:
@ly-wang19 Round-tripped this live on FalkorDB: saving an edge with reference_time and reading it back gave None. With your patch the read query returns the correct value. All-provider tests pass.

prasmussen15 (approved):

prasmussen15 (approved):

PR #1610, opened 2026-06-22
docs: fix incorrect LLM provider in gliner2 example comments and Pyth…

KlossKarl:
While working through the gliner2 example I noticed the inline comments describe the reasoning stage as being handled by "an OpenAI client" in three places (gliner2_neo4j.py:87, :119, :164), but the example actually wires up Gemini (GeminiClient / GeminiEmbedder), and the example's README already says Gemini. GLiNER2Client requires an explicit llm_client and has no OpenAI default, so the comments are just stale — likely copied from an OpenAI-based example. I changed the two comments describing this example's concrete setup to say "Gemini" and left the general architecture note as "the LLM client" since GLiNER2Client accepts any provider.

While in the examples I also caught a small mismatch in the quickstart README, which lists Python 3.9+ even though pyproject.toml requires >=3.10 — so I bumped it to 3.10+. Both are doc-only fixes with no behavior change; nothing more than keeping the comments and prerequisites honest about what the code does.

### DIFF
--- a/examples/gliner2/gliner2_neo4j.py
+++ b/examples/gliner2/gliner2_neo4j.py
@@ -83,8 +83,8 @@ entity_types: dict[str, type[BaseModel]] = {
 #################################################
 # GLiNER2 is a lightweight extraction model
 # (205M-340M params) that runs locally on CPU.
-# It handles entity extraction (NER), while an
-# OpenAI client handles edge/fact extraction,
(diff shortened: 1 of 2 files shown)


KlossKarl:
I have read the CLA Document and I hereby sign the CLA on behalf of myself, e-mail: klosskyle92@gmail.com

PR #1613, opened 2026-06-24
fix: handle datetime serialization in to_prompt_json

### LINKED ISSUE #1526: to_prompt_json crashes on datetime attribute values (Object of type datetime is not JSON serializable)
### What I was doing

Ingesting episodes with custom entity types whose attributes include `datetime`/`date` fields (a date extracted into an entity attribute, e.g. a contract `signed_at` or a person `date_of_birth`). The **first** ingest of a brand-new entity succeeds. The **second** ingest that touches the same entity triggers entity-summary regeneration, which serializes the entity's `attributes` dict into the summary prompt and crashes.

### Call path

```
extract_attributes_from_nodes
  → _extract_entity_summaries_batch
  → extract_summaries_batch
  → to_prompt_json(context['attributes'])
  → json.dumps(...)   # graphiti_core/prompts/prompt_helpers.py — no default= handler
```

### Expected

`to_prompt_json` serializes attribute dicts containing `datetime`/`date` values (rendering them as ISO-8601), the way attribute data realistically looks.

### Actual

```
TypeError: Object of type datetime is not JSON serializable
```

`graphiti_core/prompts/prompt_helpers.py::to_prompt_json` calls `json.dumps(data, ...)` with no `default=` handler, so any `datetime`/`date` value raises.

### Minimal repro

```python
from graphiti_core.prompts.prompt_helpers import to_prompt_json

to_prompt_json({"signed_at": datetime(2024, 3, 15, tzinfo=timezone.utc)})
# TypeError: Object of type datetime is not JSON serializable
```

### Version

graphiti-core 0.29.x (also present on `main` / v0.29.1).

### Fix

PR incoming — adds a `default=` fallback to `to_prompt_json` (ISO-8601 for `date`/`datetime`, `str()` otherwise) plus unit tests. Output is unchanged for already-serializable data.

neto007:
## Problem

`to_prompt_json()` in `graphiti_core/prompts/prompt_helpers.py` calls `json.dumps(data, ...)` without a `default=` handler, so any `datetime` or `date` attribute value raises:

```
TypeError: Object of type datetime is not JSON serializable
```

This crash occurs during entity-summary regeneration on the second ingest of an entity that has datetime attributes (e.g., `signed_at`, `date_of_birth`).

**Call path:**
```
extract_attributes_from_nodes
  → _extract_entity_summaries_batch
  → extract_summaries_batch
  → to_prompt_json(context['attributes'])
  → json.dumps(...)   # no default= handler
```

## Fix

Added a `default` handler to `json.dumps()` in `to_prompt_json()` that:
1. Serializes `datetime` and `date` objects to ISO-8601 format via `.isoformat()`
2. Falls back to `str()` for any other non-serializable objects

This is backward-compatible — already-serializable data is unchanged.

## Files Changed

- **`graphiti_core/prompts/prompt_helpers.py`** — Added `default_serializer` inner function with `datetime`/`date` handling, imported `datetime` and `date` from the `datetime` module
- **`tests/helpers_test.py`** — Added `test_to_prompt_json_datetime` covering datetime with timezone, date objects, mixed data, already-serializable data, and non-serializable fallback

## Testing

```bash
PYTHONPATH=. DISABLE_NEO4J=1 DISABLE_FALKORDB=1 DISABLE_NEPTUNE=1 pytest tests/helpers_test.py -v
# tests/helpers_test.py::test_lucene_sanitize PASSED
# tests/helpers_test.py::test_to_prompt_json_datetime PASSED
```

Closes #1526


Naseem77:
@neto007 The datetime fix is correct (verified the TypeError and the ISO output with your patch), but 346 of the 348 files are permission-only changes. A clean 2-file diff would make this easy to merge.

prasmussen15 (changes requested):
The datetime fix is useful, but it is buried in a 348-file change. Please split it into a focused PR containing only the serialization fix and its tests, then rebase it onto current main.

Issue #912, opened 2025-09-17
[BUG] Validation error on pydantic ExtractedEntities


did-82:
## Bug Description
I get a validation error when running the quickstart_neo4j on the `ExtractedEntities` pydantic model.

I am using `deepseek-r1:7b`.

LLM response is the following:
`'{"$defs": {"ExtractedEntity": {"properties": {"name": {"description": "Name of the extracted entity", "title": "Name", "type": "string"}, "entity_type_id": {"description": "ID of the classified entity type. Must be one of the provided entity_type_id integers.", "title": "Entity Type Id", "type": "integer"}}, "required": ["name", "entity_type_id"], "title": "ExtractedEntity", "type": "object"}}, "properties": {"extracted_entities": {"description": "List of extracted entities", "items": {"$ref": "#/$defs/ExtractedEntity"}, "title": "Extracted Entities", "type": "array"}}, "required": ["extracted_entities"], "title": "ExtractedEntities", "type": "object"}'`

that throws the following error:
`Field required [type=missing, input_value={'$def': {'ExtractedEntit... 'entity_type_id': 0}]}}, input_type=dict]`

## Steps to Reproduce
Provide a minimal code example that reproduces the issue:

```python
from datetime import datetime
from graphiti_core import Graphiti
from graphiti_core.nodes import EpisodeType
from graphiti_core.search.search_config_recipes import NODE_HYBRID_SEARCH_RRF
from graphiti_core.llm_client import LLMConfig
from graphiti_core.embedder.openai import OpenAIEmbedder, OpenAIEmbedderConfig
from graphiti_core.cross_encoder.openai_reranker_client import OpenAIRerankerClient

from graphiti_core.llm_client.openai_generic_client import OpenAIGenericClient


# Configure Ollama LLM client
llm_config = LLMConfig(
    api_key="ollama",  # Ollama doesn't require a real API key, but some placeholder is needed
    model="deepseek-r1:7b",
    small_model="deepseek-r1:7b",
    base_url="http://localhost:11434/v1",  # Ollama's OpenAI-compatible endpoint
)

llm_client = OpenAIGenericClient(config=llm_config)

# Initialize Graphiti with Ollama clients
graphiti = Graphiti(
    "bolt://localhost:7687",
    "neo4j",
    "your_password",
    llm_client=llm_client,
    embedder=OpenAIEmbedder(
        config=OpenAIEmbedderConfig(
            api_key="ollama",  # Placeholder API key
[...]

Issue #868, opened 2025-08-27
[BUG] Cannot get minimal example to work with Ollama


Powerkrieger:
## Bug Description
I tried setting up the quickstart example with the Ollama configuration from the README. I get a pydantic Field required error.

## Steps to Reproduce
Provide a minimal code example that reproduces the issue:

```python
import asyncio
import json
import logging
from datetime import datetime, timezone
from logging import INFO

from dotenv import load_dotenv
from graphiti_core import Graphiti
from graphiti_core.cross_encoder.openai_reranker_client import OpenAIRerankerClient
from graphiti_core.embedder.openai import OpenAIEmbedder, OpenAIEmbedderConfig
from graphiti_core.llm_client.config import LLMConfig
from graphiti_core.llm_client.openai_generic_client import OpenAIGenericClient
from graphiti_core.nodes import EpisodeType


async def main():
    # Configure logging
    logging.basicConfig(
        level=INFO,
        format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
        datefmt='%Y-%m-%d %H:%M:%S',
    )
    logger = logging.getLogger(__name__)
    load_dotenv()

    # Configure Ollama LLM client
    llm_config = LLMConfig(
        api_key="abc",  # Ollama doesn't require a real API key
        model="deepseek-r1:7b",
        small_model="deepseek-r1:7b",
        base_url="http://xxx.xxx.xxx.xxx:11434/v1",  # Ollama provides this port
    )

    llm_client = OpenAIGenericClient(config=llm_config)
    try:
        # Initialize Graphiti with Ollama clients
        graphiti = Graphiti(
            "bolt://localhost:7687",
            "neo4j",
            "password",
            llm_client=llm_client,
            embedder=OpenAIEmbedder(
                config=OpenAIEmbedderConfig(
                    api_key="abc",
                    embedding_model="nomic-embed-text",
                    embedding_dim=768,
                    base_url="http://xxx.xxx.xxx.xxx:11434/v1",
                )
            ),
            cross_encoder=OpenAIRerankerClient(client=llm_client, config=llm_config),
        )

        await graphiti.build_indices_and_constraints()

        # Episodes list containing both text and JSON episodes
        episodes = [
            {
                'content': 'Kamala Harris is the Attorney General of California. She was previously '
                           'the district attorney for San Francisco.',
                'type': EpisodeType.text,
                'description': 'podcast transcript',
            },
            {
[...]

PR #1619, opened 2026-06-27
fix(mcp): expose generic structured output mode

### LINKED ISSUE #1618: MCP server should expose OpenAIGenericClient structured_output_mode
## Problem

The MCP server currently detects OpenAI-compatible endpoints (Ollama, LM Studio, vLLM, routers/proxies, etc.) and uses `OpenAIGenericClient`, but the MCP config does not expose the core client's `structured_output_mode` option. As a result, MCP users are effectively locked into the `json_schema` default for generic OpenAI-compatible endpoints.

Some OpenAI-compatible endpoints accept `response_format={"type":"json_schema"}` but do not actually enforce the nested schema. In those cases Graphiti ingestion can fail during extraction because the LLM returns a JSON object with the wrong shape.

Example failure shapes observed:

```text
ExtractedEntities() argument after ** must be a mapping, not list
```

or objects missing the expected field, e.g. returning:

```json
{"entities":[{"text":"Mac Graphiti MCP","type":1}]}
```

where Graphiti expects:

```json
{"extracted_entities":[{"name":"Mac Graphiti MCP","entity_type_id":1}]}
```

## Reproduction context

This reproduced with the standalone MCP image:

```text
image: zepai/knowledge-graph-mcp:standalone
org.opencontainers.image.version: 1.0.2
graphiti.core.version: 0.28.2
```

Configuration used an OpenAI-compatible local router as the LLM endpoint:

```yaml
llm:
  provider: "openai"
  model: "auto"
  providers:
    openai:
      api_url: "http://host.docker.internal:2099/v1"
```

The router accepted `json_schema` requests, but routed to a backend model that returned a different extraction shape (`entities` instead of `extracted_entities`). Switching the same prompt to JSON-object mode with the Pydantic schema injected into the prompt made the output match Graphiti's expected `ExtractedEntities` shape, and a real `add_episode` canary then completed successfully.

## Why this is not just a model-quality issue

The core `OpenAIGenericClient` already has the intended escape hatch:

```python
OpenAIGenericClient(..., structured_output_mode="json_object")
```

The README also documents the `json_schema` vs `json_object` trade-off. The missing piece is that the MCP server config/factory does not pass this option through, so MCP deployments cannot select the fallback without patching `mcp_server/src/services/factories.py` locally.

## Proposed fix

Expose an MCP config option such as:

```yaml
llm:
  structured_output_mode: "json_schema"  # default; or "json_object"
```

and pass it to `OpenAIGenericClient` when the MCP server chooses the generic OpenAI-compatible client.

I am opening a small PR for this so users can set `structured_output_mode: "json_object"` for routers/proxies/local endpoints that accept but do not enforce `json_schema`.

## Related

This is closely related to existing structured-output validation failures such as #912 and #868, but the focus here is specifically the MCP server missing the config pass-through for an option that already exists in `graphiti_core`.

xiaoyaner0201:
## Summary

Expose `llm.structured_output_mode` in the MCP server config and pass it through to `OpenAIGenericClient` for OpenAI-compatible non-OpenAI endpoints.

## Why

`graphiti_core.llm_client.openai_generic_client.OpenAIGenericClient` already supports:

- `json_schema` (default): request provider-side schema enforcement
- `json_object`: request JSON-object mode and inject the Pydantic schema into the prompt

The MCP server currently chooses `OpenAIGenericClient` for OpenAI-compatible endpoints such as Ollama, LM Studio, vLLM, and routers/proxies, but it does not expose this option. Some compatible endpoints accept `json_schema` but do not actually enforce the nested schema, causing ingestion failures such as `ExtractedEntities` receiving a list or an object missing `extracted_entities`.

This lets MCP users set:

```yaml
llm:
  structured_output_mode: "json_object"
```

instead of patching `mcp_server/src/services/factories.py` locally.

Fixes #1618. Related to #912 and #868.

## Changes

- Add `LLMConfig.structured_output_mode` with default `json_schema`.
- Pass the configured value to `OpenAIGenericClient` for generic OpenAI-compatible endpoints.
- Add `LLM_STRUCTURED_OUTPUT_MODE` to the default MCP config.
- Document when to use `json_object` in the MCP README.
- Add factory tests covering default and configured structured-output mode.

## Testing

- ✅ `uv run --project mcp_server pytest mcp_server/tests/test_factories.py -q`
  - `26 passed`
- ✅ `uv run --with '.[falkordb]' --with pytest --with pytest-asyncio pytest tests/llm_client/test_openai_generic_client.py -q`
  - `9 passed, 1 warning`
- ✅ `uv run --project mcp_server ruff format --check mcp_server/src/config/schema.py mcp_server/src/services/factories.py mcp_server/tests/test_factories.py`
- ✅ `uv run --project mcp_server ruff check mcp_server/src/config/schema.py mcp_server/src/services/factories.py mcp_server/tests/test_factories.py`
- ✅ `git diff --check`

## Notes

This keeps the current default behavior (`json_schema`) unchanged. Users only opt into `json_object` for endpoints where provider-side schema enforcement is unreliable.


xiaoyaner0201:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: dongsjoa@gmail.com

Naseem77:
@xiaoyaner0201 Tested it: structured_output_mode sat in the config but never reached OpenAIGenericClient. With your patch it flows from config through the factory to the client, both modes verified, invalid values rejected. 26 tests pass.

prasmussen15 (approved):

prasmussen15 (approved):

PR #1622, opened 2026-07-01
fix(server): use dedicated client for /messages background ingest

Algorhythmicss:
## Problem

`POST /messages` returns `202` but never persists anything — `MATCH (n) RETURN count(n)` stays `0` and `/search` always returns `{"facts": []}`.

The endpoint queues `add_messages_task` onto the module-level `AsyncWorker`, but the task closes over the **request-scoped** `graphiti` client from the `get_graphiti` FastAPI dependency. That dependency's `finally: await client.close()` runs as soon as the HTTP request returns (`202`) — before the queued job executes. When the worker later runs the job, the Neo4j driver is already closed → `DriverError('Driver closed')`.

Because the worker loop only catches `asyncio.CancelledError`, that exception propagates out of `while True` and silently kills the worker, so all subsequent ingests are dropped too with no log output.

## Fix

- Extract client construction into `build_graphiti(settings)` in `zep_graphiti.py`.
- `add_messages_task` now builds and closes its own client (independent of request scope).
- The worker loop catches and logs generic exceptions so one failing job no longer kills ingestion.

## Testing

Verified on the Neo4j 5.26.2 backend (`docker compose up`): after the fix, ingesting messages creates `Entity` nodes and bi-temporal `RELATES_TO` edges, and `POST /search` returns the extracted facts with `valid_at` / `invalid_at` metadata.

> Note: verified against the Neo4j path only; I did not exercise the FalkorDB profile or run the full server test suite.


Naseem77:
@Algorhythmicss Reproduced both issues: queued /messages jobs ran on a client FastAPI had already closed ('Driver closed'), and one failing job killed the worker loop permanently. With your change ingestion gets its own client and the worker survives failing jobs. Verified with live requests.

prasmussen15 (changes requested):

PR #1626, opened 2026-07-02
fix(falkordb): respect reference_time in retrieve_episodes

### LINKED ISSUE #1625: [BUG] FalkorDB retrieve_episodes can return episodes after reference_time
## Bug Description

With the FalkorDB driver, `retrieve_episodes(reference_time=...)` can return episodic nodes whose `valid_at` is after the requested `reference_time`.

The implementation intends to filter by `e.valid_at <= $reference_time`, but FalkorDB can still return rows for which the same expression evaluates to `False` when projected. This makes point-in-time episode retrieval include future episodes.

## Steps to Reproduce

Provide a minimal code example that reproduces the issue:

```python
from uuid import uuid4

async def main():
    driver = FalkorDriver(host='localhost', port=6379, database=f'repro_{uuid4().hex}')
    await driver.build_indices_and_constraints()

    group_id = 'repro-reference-time'
    created_at = datetime(2024, 1, 1, tzinfo=timezone.utc)

    await driver.episode_node_ops.save(
        driver,
        EpisodicNode(
            uuid=str(uuid4()),
            name='past episode',
            group_id=group_id,
            source=EpisodeType.text,
            source_description='test',
            content='past',
            created_at=created_at,
            valid_at=datetime(2024, 1, 1, tzinfo=timezone.utc),
        ),
    )
    await driver.episode_node_ops.save(
        driver,
        EpisodicNode(
            uuid=str(uuid4()),
            name='future episode',
            group_id=group_id,
            source=EpisodeType.text,
            source_description='test',
            content='future',
            created_at=created_at,
            valid_at=datetime(2024, 3, 1, tzinfo=timezone.utc),
        ),
    )

    episodes = await driver.episode_node_ops.retrieve_episodes(
        driver,
        datetime(2024, 2, 1, tzinfo=timezone.utc),
        last_n=10,
        group_ids=[group_id],
    )
    print([episode.name for episode in episodes])
```

A lower-level FalkorDB check shows the same predicate behaving inconsistently:

```cypher
MATCH (e:Episodic)
WHERE e.valid_at <= $ref
RETURN e.name, e.valid_at, e.valid_at <= $ref AS valid_at_ok
```

Rows can pass the `WHERE` clause while `valid_at_ok` is `False` for the same `$ref`.

## Expected Behavior

`retrieve_episodes` should only return episodes whose `valid_at` is less than or equal to `reference_time`. In the example above, only `past episode` should be returned.

## Actual Behavior

FalkorDB can return episodes after the requested reference time. In the example above, `future episode` can be returned even though its `valid_at` is `2024-03-01` and the requested reference time is `2024-02-01`.

## Environment

- **Graphiti Version**: current `main` / 0.29.2
- **Python Version**: 3.12.13
- **Operating System**: macOS
- **Database Backend**: FalkorDB via local Docker, Python client `falkordb` 1.4.0
- **LLM Provider & Model**: not applicable; this reproduces at the driver operation layer without LLM calls

## Installation Method

- [ ] pip install
- [ ] uv add

## Error Messages/Traceback

```text
No exception is raised. The query returns rows outside the requested temporal boundary.
```

## Configuration

```python
driver = FalkorDriver(host='localhost', port=6379, database='...')
```

## Additional Context

This appears to be specific to FalkorDB's direct `WHERE e.valid_at <= $reference_time` evaluation path for ISO datetime strings. Computing the predicate first and filtering on the projected boolean behaves as expected:

```cypher
MATCH (e:Episodic)
WITH e, e.valid_at <= $ref AS valid_at_ok
WHERE valid_at_ok
RETURN e
```

This matters for Graphiti because `retrieve_episodes` is documented as a point-in-time query, and future episodes leaking into the result can affect episode context selection.

## Possible Solution

In the FalkorDB `retrieve_episodes` implementation, project the temporal comparison first and then filter on that boolean:

```cypher
MATCH (e:Episodic)
WITH e, e.valid_at <= $reference_time AS valid_at_ok
WHERE valid_at_ok
...
```

The same pattern should be used in the saga-specific branch.

pcy06:
## Summary
Fixes FalkorDB episode retrieval so `retrieve_episodes(reference_time=...)` does not return episodes whose `valid_at` is after the requested reference time.

The FalkorDB query now projects `e.valid_at <= $reference_time` into a boolean with `WITH` and filters on that value. This avoids a FalkorDB direct `WHERE e.valid_at <= $reference_time` path that can return rows even when the same predicate projects as `False`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Documentation/Tests

## Objective
This is a bug fix for the FalkorDB driver. Point-in-time episode retrieval is documented to return only episodes at or before `reference_time`; future episodes leaking into the result can affect the context Graphiti uses for subsequent episode ingestion.

The change is intentionally limited to `FalkorEpisodeNodeOperations.retrieve_episodes` and the corresponding driver tests.

## Testing
- [ ] All existing tests pass

Commands run:

```bash
uv run pytest tests/driver/test_falkordb_driver.py -q
DISABLE_FALKORDB=1 DISABLE_KUZU=1 DISABLE_NEPTUNE=1 uv run pytest tests/driver/test_falkordb_driver.py -q -m 'not integration'
make lint
```

## Breaking Changes
- [x] No breaking changes

If this is a breaking change, describe:
- Not applicable

## Checklist
- [x] Self-review completed
- [x] Documentation updated where necessary - no docs change needed for this bug fix

## Related Issues
Closes #1625


pcy06:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: pcy06@kakao.com

pcy06:
recheck

Naseem77:
@pcy06 Good catch on reference_time! The legacy retrieve_episodes path in utils/maintenance has the same bug and isn't covered. Patching that too would make this complete.

prasmussen15 (changes requested):
Please patch the legacy retrieve_episodes path in utils/maintenance as well, and add coverage so reference_time is respected on both paths.

pcy06:
@prasmussen15 Addressed the review feedback:

- Applied the same `valid_at` boolean projection to the legacy `retrieve_episodes` in `utils/maintenance/graph_data_operations.py`, gated to FalkorDB so other providers keep the direct predicate (and index usage).
- Added coverage for both paths: query-shape unit tests plus an integration test verifying `reference_time` is respected against a real FalkorDB.
- Also restructured the integration tests so assertion failures are reported instead of being swallowed by the connection skip.

Ready for another look.

PR #1631, opened 2026-07-04
fix(community): bound label propagation iterations to guarantee termi…

MrNirmata:
…nation

label_propagation() ran a `while True` loop that only exited on full convergence. Synchronous (Jacobi-style) label propagation is not guaranteed to converge: on symmetric or tie-heavy graphs the assignment can settle into a period-two oscillation and flip between two states forever, so the loop never terminates. build_communities() then hangs, pinning a CPU with no progress and no error.

Cap the propagation at MAX_LABEL_PROPAGATION_ITERATIONS (1000) and return the latest assignment when it has not converged, logging a warning. Graphs that do converge are unaffected: the `no_change` early-exit still fires exactly as before, so results are identical for every converging input.

Add unit tests for a small non-convergent graph that previously hung, the non-convergence warning, and a separable graph that still converges to the expected communities.

## Summary
Brief description of the changes in this PR.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
**For new features and performance improvements:** Clearly describe the objective and rationale for this change.

## Testing
- [ ] Unit tests added/updated
- [ ] All existing tests pass

## Breaking Changes

If this is a breaking change, describe:
- What functionality is affected
- Migration path for existing users

## Checklist
- [ ] Self-review completed
- [ ] No secrets or sensitive information committed

## Related Issues
Closes #[issue number]

### DIFF
--- a/graphiti_core/utils/maintenance/community_operations.py
+++ b/graphiti_core/utils/maintenance/community_operations.py
@@ -18,6 +18,7 @@ from graphiti_core.utils.maintenance.edge_operations import build_community_edge
 from graphiti_core.utils.text_utils import MAX_SUMMARY_CHARS, truncate_at_sentence
 
 MAX_COMMUNITY_BUILD_CONCURRENCY = 10
+MAX_LABEL_PROPAGATION_ITERATIONS = 1000
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


MrNirmata:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: mrnirmata@gmail.com

PR #1633, opened 2026-07-06
fix(search): apply property_filters and validate group_id on write path

danielchalef:
## Summary
Fixes two related correctness/security issues in search-filter and partition-key (`group_id`) handling.

**1. `SearchFilters.property_filters` was silently ignored.**
The field was declared on `SearchFilters` but never read by `node_search_filter_query_constructor` or `edge_search_filter_query_constructor`, so searches always executed without the intended property filter and returned data callers believed was restricted. Both constructors now build parameterized Cypher fragments from `property_filters` via a shared `property_filter_query_constructor` helper (`n.<prop>`/`e.<prop>` with the value passed as a query parameter; `IS NULL`/`IS NOT NULL` emit no parameter). Because Cypher property keys cannot be parameterized, `property_name` is validated against a safe-identifier pattern (`validate_property_name`) both at the Pydantic layer and defensively inside the constructor, preventing query injection.

**2. Write/read asymmetry on `group_id`.**
The write path accepted arbitrary `group_id` values while the read/delete path rejects anything outside `[a-zA-Z0-9_-]`, allowing records to be written that are permanently unreachable and undeletable via any API path. `group_id` is now validated on the core `Node` and `Edge` models and on the `AddMessagesRequest` / `AddEntityNodeRequest` server DTOs. `GroupIdValidationError` now also subclasses `ValueError` so Pydantic surfaces it as a normal validation error (HTTP 422) instead of a 500.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
Restore the documented behavior of `property_filters` (operators rely on it to scope results) and close the write-path gap that let unreachable/undeletable records be persisted.

## Testing
- [x] All existing tests pass

New regression tests in `tests/utils/search/test_search_security.py` cover:
- node & edge property-filter query construction
- `IS NULL`/`IS NOT NULL` operators emitting no params
- rejection of unsafe property names (Pydantic + constructor bypass via `model_construct`)
- write-path `group_id` rejection on `EntityNode` / `EntityEdge` and acceptance of valid/default values

Existing search-security, node, edge, and attribute unit tests pass; `make format`, `ruff check`, and `pyright` are clean on the changed files.

## Breaking Changes

Behavioral note (not a code-contract break): callers that were previously writing `group_id` values containing characters outside `[a-zA-Z0-9_-]` will now receive a validation error at write time instead of silently creating unreachable records. This aligns the write path with the already-enforced read/delete path.

## Checklist
- [x] Self-review completed

## Related Issues
N/A

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gj4mbjvGJ4oEtqdzTN1UqP)_


danielchalef:
Thanks for the review — addressed in d38e1e9.

**Hydration hard-fail (main finding):** moved `group_id` validation off the `Node`/`Edge` field validators (which fired on every construction, including `get_*_from_record` hydration) to the write boundary, so reads of legacy records with non-conforming `group_id`s are tolerant again:
- Validate in `EntityNode.save()` and `EntityEdge.save()`.
- Validate every element in `add_nodes_and_edges_bulk_tx` (the bulk path bypasses `save()`; this also covers `add_triplet`, which persists through it).
- `add_episode`/`add_episode_bulk` already validate at entry.
- Server DTOs now re-raise as `ValueError` so they surface a 422 regardless of whether the installed `graphiti-core`'s `GroupIdValidationError` subclasses `ValueError` (the server depends on the published core, where it doesn't yet).

**Suggestions:**
- Added server DTO rejection/acceptance tests (`server/tests/test_dto_validation.py`), an `is_not_null` case, and a shared node/edge filter-semantics test; documented that a single `property_filters` list is intentionally applied to `n` (node search) and `e` (edge search).
- `cypher_to_opensearch_operator` is indeed dead and pre-existing; confirmed there's no OpenSearch/aoss driver — every search backend goes through the two Cypher constructors, so `property_filters` isn't silently bypassed. Left the dead helper alone to keep this diff scoped.

---
_Generated by [Claude Code](https://claude.ai/code)_

danielchalef:
Addressed the remaining completeness note in 7e95161: lifted the write-boundary check into `Node._validate_for_write()` / `Edge._validate_for_write()` and call it from **every** concrete `save()` override (episodic/community/saga nodes, episodic/community/has-episode/next-episode edges), so a direct `EpisodicNode(...).save()` / `CommunityNode(...).save()` with a non-conforming `group_id` no longer slips through. Hydration still constructs models directly and stays tolerant. Added a regression test that a non-entity save (`EpisodicNode`) also rejects an unsafe `group_id`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Naseem77:
@danielchalef Verified this: property_filters passed into search were being silently ignored, and invalid group_ids were accepted on write. With your patch the filters generate real clauses and bad group_ids get rejected with a clear error. The 20 validation tests pass.

prasmussen15 (changes requested):
Required update: CLA Assistant is failing because this branch includes an unallowlisted `claude` committer. Please ensure every non-exempt committer has a valid CLA signature (or rewrite the commits under a signer identity), then rerun the CLA check.

PR #1635, opened 2026-07-08
fix: align Neptune community AOSS index

### LINKED ISSUE #1632: Mismatch between AOSS initialization and registration indices
## Bug Description
In the current code, the initialization process (`build_indices_and_constraints`) creates four indices (`community_name`, `edge_name_and_fact`, `episode_content`, and `node_name_and_summary`), whereas `node.py` attempts to save data to the `communities` index.

Consequently, while four indices are indeed created in AOSS, data is not actually saved because the index names do not match during the data ingestion phase.

iroiro147:
## Summary
- align the shared `COMMUNITY_INDEX_NAME` default with Neptune's registered AOSS community index (`community_name`)
- use that same constant for AOSS registration, Neptune community saves, fallback `CommunityNode.save`, and Neptune community fulltext search
- remove the invalid `await` around the sync Neptune `save_to_aoss` fallback call
- add focused regression coverage for the community AOSS index contract

Closes #1632

## Testing
- `make format`
- `make lint`
- `uv run pytest tests/driver/test_neptune_aoss_indices.py tests/driver/test_falkordb_driver.py -q` -> 28 passed, 1 skipped, 1 warning
- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/driver/test_neptune_aoss_indices.py -q` -> 3 passed, 1 warning

## Notes
- I also attempted `make test`; it reached 168 passed, 2 skipped, 10 deselected, then errored on unavailable local Neo4j at `localhost:7687` in `tests/test_add_triplet.py`. I interrupted the repeated Neo4j retries after confirming the failure was external-service setup, not this candidate change.

### DIFF
--- a/graphiti_core/driver/driver.py
+++ b/graphiti_core/driver/driver.py
@@ -52,7 +52,7 @@ load_dotenv()
 
 ENTITY_INDEX_NAME = os.environ.get('ENTITY_INDEX_NAME', 'entities')
 EPISODE_INDEX_NAME = os.environ.get('EPISODE_INDEX_NAME', 'episodes')
-COMMUNITY_INDEX_NAME = os.environ.get('COMMUNITY_INDEX_NAME', 'communities')
+COMMUNITY_INDEX_NAME = os.environ.get('COMMUNITY_INDEX_NAME', 'community_name')
 ENTITY_EDGE_INDEX_NAME = os.environ.get('ENTITY_EDGE_INDEX_NAME', 'entity_edges')
 
 
--- a/graphiti_core/driver/neptune/operations/community_node_ops.py
+++ b/graphiti_core/driver/neptune/operations/community_node_ops.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING, Any
 
-from graphiti_core.driver.driver import GraphProvider
+from graphiti_core.driver.driver import COMMUNITY_INDEX_NAME, GraphProvider
 from graphiti_core.driver.operations.community_node_ops import CommunityNodeOperations
 from graphiti_core.driver.query_executor import QueryExecutor, Transaction
 from graphiti_core.driver.record_parsers import community_node_from_record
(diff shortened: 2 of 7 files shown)


iroiro147:
Hi maintainers — a gentle follow-up on this bug fix for #1632. It aligns the Neptune/AOSS community index name across registration, save, fallback, and search paths, removes the invalid await in the fallback, and includes focused regression coverage. GitHub reports the branch mergeable; CLA, Ruff, and fork checks pass, and the remaining triage failure is the repository automation’s external-contributor permission check. Happy to adjust the implementation or tests if you would prefer a different approach.

iroiro147:
Rebased onto current `main` (head `43b07a0`) — the branch was 100 commits behind,
which mattered because the repo requires branches to be up to date. The old
`triage` red is also gone: that check came from `pr-triage.yml`, which was
removed from `main` in #1788, and it was never a code failure — it was
`anthropics/claude-code-action` refusing to run for an actor with only read
permission, which is every outside contributor by construction.

Re-confirming the bug is still live on `main`, because the title undersells it:

```
graphiti_core/nodes.py:700                   await driver.save_to_aoss('communities', …)
graphiti_core/driver/neptune_driver.py:356   def save_to_aoss(self, name, data) -> int   # sync
graphiti_core/driver/neptune/operations/search_ops.py:472   driver.run_aoss_query('community_name', …)
```

Two separate defects. The AOSS index is written as `communities` but registered
and searched as `community_name`, which is the split-brain #1632 reports. And
`save_to_aoss` is synchronous, so that `await` raises
`TypeError: object int can't be used in 'await' expression` the moment the
Neptune community path runs at all.

On the one change that looks risky — `COMMUNITY_INDEX_NAME`'s default moving from
`'communities'` to `'community_name'` — that constant is defined in
`driver.py:55` and referenced nowhere else in `graphiti_core`, so it is currently
dead config and the change is inert for non-Neptune drivers. The `'community_name'`
string Neo4j/FalkorDB/Kuzu use is a graph fulltext index name and is untouched.

`ruff` passes. `pyright` and `unit-tests` cannot run at all until a maintainer
approves the fork workflow runs — the previous ones sat in `action_required` and
expired after thirty days without ever starting a job, so their red is an
artefact of the queue, not of this branch.

PR #1636, opened 2026-07-08
fix: capture init task ref + cancel on close for Neo4jDriver and FalkorDriver

### LINKED ISSUE #1513: Neo4jDriver.__init__:57 schedules orphan create_task without cancel/await pair
## Issue

`graphiti_core/driver/neo4j_driver.py:51-60` schedules an asyncio task via `loop.create_task(self.build_indices_and_constraints())` but discards the Task reference. Without `_tasks: set[asyncio.Task]` membership or an `await` pair in `close()`, this is a fire-and-forget task whose lifecycle is not managed.

## Reproducer

In a pytest-asyncio function-scoped loop:

```python
import pytest

@pytest.mark.asyncio
async def test_a():
    g = Graphiti(uri, user, password)
    # asyncio.get_event_loop().create_task fires at line 57.
    # Task starts CREATE FULLTEXT INDEX Cypher.
    # Test function returns — pytest-asyncio closes the loop.
    # Background task is cancelled mid-transaction.
```

The `IF NOT EXISTS` guard + race-swallow at `neo4j_driver.py:100-105` makes this mostly silent in the happy case, but:

1. Connection in the driver pool may be left in mid-tx state.
2. `RuntimeWarning: coroutine was never awaited` noise (depending on Python version).
3. Consumers that call `await driver.build_indices_and_constraints()` explicitly (the documented API) race with the ctor task, producing `EquivalentSchemaRuleAlreadyExists` (swallowed) and wasted index-creation work.

## Suggested fix

Either:

(a) **Capture the task reference** and add it to a `self._tasks: set[asyncio.Task]` (or similar). Provide `close()` that drains/cancels remaining tasks before closing the driver client.

(b) **Make the side-effect explicit** — remove the `create_task` from `__init__`, and require callers to `await driver.build_indices_and_constraints()`. Document this in the migration notes for the next release.

## Consumer impact

The evo-memory project (Knowledge Graph harness for AI agents) mitigates this at the wrapper level by removing the duplicate explicit `await` call and trusting the ctor-fired task. Other consumers may not realize their explicit awaits are racing with the orphan task.

## References

- `graphiti_core/driver/neo4j_driver.py:51-60` (orphan task schedule)
- `graphiti_core/driver/neo4j_driver.py:100-105` (race swallow)
- Similar pattern in `graphiti_core/driver/falkordb_driver.py:157` (also orphan)

## Note

Filing as upstream tracking only — not requesting urgent action. Our consumer wrapper mitigation works defensively. Posting in case other consumers hit the same race.

nandanadileep:
## Summary

Both `Neo4jDriver.__init__` and `FalkorDriver.__init__` schedule `build_indices_and_constraints()` via `loop.create_task()` without capturing the task reference. This creates a fire-and-forget background task whose lifecycle is not managed.

## Problem

1. **Orphan task**: The task reference is discarded, so there's no way to cancel or await it during shutdown.
2. **Race condition**: Callers that explicitly `await driver.build_indices_and_constraints()` race with the constructor-fired task, producing `EquivalentSchemaRuleAlreadyExists` errors.
3. **Unhandled exceptions**: If the task fails (e.g., connection error), "Task exception was never retrieved" warnings appear.
4. **Mid-tx state**: Under pytest-asyncio, when the test function returns and the event loop closes, the background task can be cancelled mid-transaction.

## Fix

- Capture the task in `self._init_task: asyncio.Task | None`
- In `close()\`, cancel the task if still running and await it with `suppress(asyncio.CancelledError)`
- Removed redundant comments and moved the `import asyncio` to the top of `neo4j_driver.py`

## Testing

- `make lint` passes
- `make typecheck` passes
- `DISABLE_FALKORDB=1 DISABLE_KUZU=1 DISABLE_NEPTUNE=1 uv run pytest -m 'not integration'` — 165 passed, 2 skipped (FalkorDB integration), 1 pre-existing Neo4j auth error (no local Neo4j)

Closes #1513


nandanadileep:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: nandanadileep29@gmail.com

Naseem77:
@nandanadileep Thanks for catching this! One gap: close() skips done() tasks, so a failed init task still leaks its exception. Calling task.exception() on done tasks too would complete it.

prasmussen15 (changes requested):
Please update close() to retrieve exceptions from completed init tasks too; skipping done() tasks still leaves a failed init task exception unobserved.

nandanadileep:
@prasmussen15 Thanks for the review! The requested change was already landed in d3fc190 (pushed Jul 20) — `close()` in both drivers now calls `self._init_task.exception()` on already-done tasks so a failed init task's exception is observed:

```python
if self._init_task is not None:
    if not self._init_task.done():
        self._init_task.cancel()
        with suppress(asyncio.CancelledError):
            await self._init_task
    elif not self._init_task.cancelled():
        self._init_task.exception()
```

The review may have been against a stale diff. That said, it surfaced a related edge case: `Task.exception()` raises `CancelledError` if the done task was cancelled externally, which would escape `close()`. Commit 91124f5 guards against that with the `cancelled()` check above. Ready for re-review 🙏

prasmussen15 (approved):

Issue #1164, opened 2026-01-20
[BUG] MCP Episodes Fail Silently When Content Contains "attributes" Keyword


AlbertoDMendoza:
## Bug Description
Episodes added via the `add_memory` MCP tool return a success message but fail silently during background processing when the episode body text contains the word "attributes". The tool returns `{"message": "Episode 'X' queued for processing in group 'default'"}`, but the episode is never saved to the knowledge graph. The failure only appears in server logs with the error: `attributes cannot be used as an attribute for DataObject as it is a protected attribute name`.

This results in silent data loss where users believe their data was successfully saved when it was not.

The `EntityEdge` Pydantic BaseModel has a field literally named `attributes`. When episode content contains the word "attributes", the entity/edge extraction and serialization process encounters a naming conflict:

1. Episode text containing "attributes" is submitted
2. Graphiti's entity/edge extraction parses the text
3. During `EntityEdge` instantiation or serialization, Pydantic encounters the word "attributes" in the content/metadata
4. Conflict occurs between the Pydantic field name and content reference
5. Pydantic raises: "attributes cannot be used as an attribute for DataObject as it is a protected attribute name"

## Steps to Reproduce

```python
# Using the Graphiti MCP add_memory tool
add_memory(
    name="Test Episode with Attributes",
    episode_body="This system has several attributes including age, status, and priority. The DOM attributes control rendering behavior.",
    source="text",
    source_description="Technical documentation"
)
```

**Expected**: Episode is saved to knowledge graph  
**Actual**: Episode fails silently during background processing

## Expected Behavior
One of the following should occur:
1. The episode should be processed successfully despite containing the word "attributes" in the content
2. OR the `add_memory` tool should return an immediate error indicating that "attributes" is a reserved keyword
3. OR the background processing error should be propagated back to the caller via a callback/webhook

## Actual Behavior
1. `add_memory` tool returns success: `{"message": "Episode 'Test Episode with Attributes' queued for processing in group 'default'"}`
2. Background processing fails with error (only visible in server logs)
3. Episode is NOT saved to the knowledge graph
4. User receives NO notification of the failure
5. Subsequent searches for the episode return no results

## Environment
- **Graphiti Version**: 1.0.0 (MCP server)
- **Python Version**: 3.10
- **Operating System**: Windows 11
- **Database Backend**: Neo4j 5.x
- **LLM Provider & Model**: OpenAI gpt-4o-mini

## Installation Method
- [x] Development installation (git clone)

[...]

PR #1638, opened 2026-07-08
fix(mcp): validate episode types before queueing

vismaytiwari:
Addresses #1164

## Summary

This validates MCP episode type configuration before an episode is added to the background queue.

Previously, `add_memory` could return a queued-success response even when `Graphiti.add_episode` would immediately reject the configured entity types. The queue worker then logged the failure later, so callers only saw success while the episode was never processed.

This reuses Graphiti's existing entity-type and excluded-entity-type validators in `QueueService.add_episode`, before creating the background task. Invalid reserved fields such as `DataObject.attributes` now fail before queueing, so the MCP tool returns an error instead of a false success message.

## Validation

- Reproduced the current behavior locally: `QueueService.add_episode` returned a queue position, then the worker logged `attributes cannot be used as an attribute for DataObject as it is a protected attribute name.`
- Verified after the change that the same case raises before queue creation and the Graphiti client is not called.
- Verified `add_memory` now returns an error response for that case and does not create a queue.
- `uv run pytest tests/test_core_parity.py -q` from `mcp_server`: 28 passed.
- `uv run ruff check src/services/queue_service.py tests/test_core_parity.py`: passed.
- `uv run ruff format --check src/services/queue_service.py tests/test_core_parity.py`: passed.
- `uv run pyright src/services/queue_service.py`: 0 errors.

Note: I also tried MCP-wide `uv run pyright src tests`, but it currently reports existing unrelated MCP test/import typing issues, so I kept the pyright check scoped to the changed service file.

### DIFF
--- a/mcp_server/src/services/queue_service.py
+++ b/mcp_server/src/services/queue_service.py
@@ -6,6 +6,9 @@ from collections.abc import Awaitable, Callable
 from datetime import datetime, timezone
 from typing import Any
 
+from graphiti_core.helpers import validate_excluded_entity_types
+from graphiti_core.utils.ontology_utils.entity_types_utils import validate_entity_types
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


vismaytiwari:
I have read the CLA Document and I hereby sign the CLA behalf of myself , e-mail: vismay.t@gmail.com

vismaytiwari:
recheck

PR #1644, opened 2026-07-11
fix(search): make FalkorDB tie-breaking deterministic

jaimehgb:
## Summary
- add UUID ascending as the secondary order for all seven score-ranked FalkorDB search primitives
- make reciprocal rank fusion break equal scores by UUID
- mirror query ordering in the FalkorDB SearchOperations implementation

This prevents database iteration order and input insertion order from changing capped search results when scores tie.

## Tests
- added focused query-rendering coverage for the legacy and SearchOperations paths
- added an equal-score RRF regression test
- `pytest --confcutdir tests/utils/search tests/utils/search/deterministic_tie_breaks_test.py tests/utils/search/search_utils_test.py -q` (20 passed)

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -218,7 +218,7 @@ class FalkorSearchOperations(SearchOperations):
             """
             + get_entity_node_return_query(GraphProvider.FALKORDB)
             + """
-            ORDER BY score DESC, uuid ASC
+            ORDER BY score DESC, n.uuid ASC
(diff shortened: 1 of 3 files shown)


Naseem77:
@jaimehgb Reproduced the nondeterminism: the same 6 tied-score nodes inserted in different orders came back in different orders, and with your patch both come back in identical UUID order. 15 tests pass. Only thing left is a ruff format run on the test file (that's the failing check) and the CLA.

prasmussen15 (changes requested):
Can you sign the CLA and fix listing issues

Issue #1642, opened 2026-07-09
[BUG] Cross-encoder reranking drops most candidates in edge_search


pleb631:
## Description

In `edge_search`, the retrieval stage collects candidates from multiple search methods (`bm25`, `cosine_similarity`, `bfs`) with `2 * limit` candidates each.

However, when using `EdgeReranker.cross_encoder`, only the first `limit` candidates from the merged candidate set are passed to the reranker:

```python
fact_to_uuid_map = {
    edge.fact: edge.uuid for edge in list(edge_uuid_map.values())[:limit]
}
```

This causes most retrieved candidates to be discarded before reranking.

## Example

Assume:

```python
limit = 20
```

With three enabled search methods:

* BM25: 40 candidates
* Vector similarity: 40 candidates
* BFS: 40 candidates

The search stage can produce up to 120 candidates.

But cross-encoder only evaluates:

```python
edge_uuid_map.values()[:20]
```

The remaining candidates are never scored.

Because `edge_uuid_map` preserves insertion order, candidates from later search methods may be completely excluded depending on the order of `search_results`.

## Impact

* Hybrid search recall is reduced.
* Candidates retrieved by vector search or BFS may never reach the reranker.
* Cross-encoder ranking quality becomes dependent on retrieval method ordering.

## Suggested Fix

Separate the reranker candidate limit from the final result limit.

For example:

candidate_limit = config.cross_encoder_candidate_limit

candidate_edges = list(edge_uuid_map.values())[:candidate_limit]

fact_to_uuid_map = {
    edge.fact: edge.uuid
    for edge in candidate_edges
}

However, simply increasing candidate_limit may still introduce bias because edge_uuid_map preserves the insertion order of search_results. Candidates from earlier search methods can occupy the candidate pool before results from other retrieval methods are considered.

A better approach would be to use a balanced merge strategy before reranking. For example, if multiple search methods return results:

[
    bm25_results,
    vector_results,
    bfs_results,
]

select candidates in a round-robin manner:

bm25[0], vector[0], bfs[0],
bm25[1], vector[1], bfs[1],
...

until reaching cross_encoder_candidate_limit.

This ensures that candidates from each retrieval method have an opportunity to be evaluated by the cross-encoder, improving recall and avoiding search-method ordering bias.

LH-kevin:
I’m interested in working on this issue.

[...]

PR #1637, opened 2026-07-08
fix(mcp): configure cross-encoder from providers instead of hardcoded OpenAI

winklemad:
## Summary

The MCP server did not pass a `cross_encoder` to `Graphiti()`, so Graphiti defaulted to `OpenAIRerankerClient()` and required `OPENAI_API_KEY` at startup even when the configured LLM and embedder both used other providers. This fixes #1393 without allowing a failed local fallback to silently restore that OpenAI default.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective

Select an available reranker from the configured providers and keep non-OpenAI MCP configurations independent of an OpenAI key.

## Change

- Add `CrossEncoderFactory` in `services/factories.py`. It prefers the LLM provider, then the embedder provider, and uses Graphiti's local `BGERerankerClient` when neither provider has a native reranker.
- Pass the selected cross-encoder into both FalkorDB and Neo4j `Graphiti()` construction.
- Warn that the local `BAAI/bge-reranker-v2-m3` fallback may download about 2.3 GB on first run.
- If the base MCP install lacks `sentence-transformers`, fail startup with actionable `providers`-extra / Graphiti-extra / provider guidance. The error is intentionally not swallowed into `cross_encoder=None`, which would silently restore the OpenAI reranker.

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests added/updated (not required for factory selection/error propagation)
- [x] All relevant existing tests pass

Validation run locally:

- `pytest tests/test_cross_encoder_factory.py -q` — 4 passed
- `pytest tests/test_factories.py tests/test_cross_encoder_factory.py -q` — 28 passed
- `ruff check .` — passed
- `ruff format --check .` — passed
- Focused Pyright on the changed files — passed

## Breaking Changes

- [ ] This PR contains breaking changes

## Checklist

- [x] Code follows project style guidelines (`make lint` equivalent checks pass for the changed files)
- [x] Self-review completed
- [x] Documentation updated where necessary (no public configuration schema change)
- [x] No secrets or sensitive information committed

## Related Issues

Closes #1393

Issue #1393, opened 2026-04-10
[BUG] MCP Server: cross_encoder/reranker not configurable, hardcoded to OpenAI — requires OPENAI_API_KEY even with non-OpenAI providers


marians:
## Bug Description

The MCP server does not pass a `cross_encoder` argument when constructing the `Graphiti` client. This causes `Graphiti.__init__` to always fall back to `OpenAIRerankerClient()`, which requires `OPENAI_API_KEY`. This makes it impossible to run the MCP server without an OpenAI API key, even when using Anthropic/Gemini/Groq for both LLM and embeddings.

## Steps to Reproduce

Configure the MCP server to use non-OpenAI providers for both LLM and embeddings:

```yaml
# config.yaml
llm:
  provider: "anthropic"
  model: "claude-haiku-4-5-20251001"
  temperature: 0.0
  providers:
    anthropic:
      api_key: "${ANTHROPIC_API_KEY}"

embedder:
  provider: "gemini"
  model: "gemini-embedding-001"
  providers:
    gemini:
      api_key: "${GOOGLE_API_KEY}"

database:
  provider: "falkordb"
```

Start the MCP server. Both LLM and embedder clients are created successfully, but initialization crashes when `Graphiti.__init__` tries to create the default `OpenAIRerankerClient`.

## Expected Behavior

The MCP server should allow configuring the cross-encoder/reranker provider, or at minimum select a matching reranker based on the configured LLM provider.

## Actual Behavior

The server crashes with:

```
openai.OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable
```

Traceback:

```
File "/app/mcp/src/graphiti_mcp_server.py", line 225, in initialize
    self.client = Graphiti(
File "/app/mcp/.venv/lib/python3.11/site-packages/graphiti_core/graphiti.py", line 224, in __init__
    self.cross_encoder = OpenAIRerankerClient()
File "/app/mcp/.venv/lib/python3.11/site-packages/graphiti_core/cross_encoder/openai_reranker_client.py", line 55, in __init__
    self.client = AsyncOpenAI(api_key=config.api_key, base_url=config.base_url)
openai.OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable
```

## Environment
[...]

Issue #777, opened 2025-07-28
[BUG] RuntimeWarning errors in MMR reranker, search does not return results


paveljakov:
## Bug Description
The MMR (Maximal Marginal Relevance) reranker in `graphiti_core/search/search_utils.py` produces numpy RuntimeWarning errors during dot product calculations when processing embedding vectors, causing search operations to return empty results. The warnings indicate mathematical issues including divide by zero, overflow, and invalid value encounters during similarity calculations and MMR score computation, which break the MMR algorithm's functionality.

## Steps to Reproduce
Provide a minimal code example that reproduces the issue:

```python
# Example search operation that triggers the MMR reranker warnings
from graphiti_core.search.search_config import SearchConfig, NodeSearchConfig, NodeSearchMethod, NodeReranker
from graphiti_core.search.search_filters import SearchFilters

# Configure search with MMR reranker
search_config = SearchConfig(
    node_config=NodeSearchConfig(
        search_methods=[NodeSearchMethod.cosine_similarity, NodeSearchMethod.bm25],
        reranker=NodeReranker.mmr,
        sim_min_score=0.6,
        mmr_lambda=0.7,
        bfs_max_depth=3
    ),
    limit=20
)

# Search filters targeting specific node types
search_filters = SearchFilters(
    node_labels=['Product', 'SparePart']
)

# Perform search that triggers MMR reranker
await graphiti.search_with_config(
    query="product features and available options",
    config=search_config,
    search_filters=search_filters
)
```

## Expected Behavior
The search operation should complete successfully without any runtime warnings and return relevant search results. The MMR reranker should properly calculate similarity scores and MMR values for all candidate embeddings, returning ranked results based on both relevance to the query and diversity.

## Actual Behavior
The search operation produces multiple numpy RuntimeWarning messages and fails to return any results:

- `RuntimeWarning: divide by zero encountered in dot` 
- `RuntimeWarning: overflow encountered in dot`
- `RuntimeWarning: invalid value encountered in dot`

These warnings occur during the MMR calculation process, specifically in the similarity matrix computation and MMR score calculation steps. When these warnings are triggered, the search operation completes but returns an empty result set instead of the expected relevant nodes.

## Environment
- **Graphiti Version**: 0.18.0
- **Python Version**: 3.13
- **Operating System**: macOS (darwin 24.5.0)
- **Database Backend**: neo4j
- **LLM Provider & Model**: OpenAI GPT 4.1-mini

## Installation Method
- [ ] pip install
[...]

PR #1647, opened 2026-07-13
feat(mcp): configurable search tuning — reranker choice, stale-fact filtering, result counts, relevance scores

### LINKED ISSUE #1645: MCP server: search behavior is hardcoded — proposal: configurable reranker, stale-fact filtering, result counts, relevance scores
## Summary

The MCP server's search tools (`search_memory_facts` and `search_nodes`) run one fixed recipe: results are ordered by reciprocal-rank fusion (RRF), capped at 10, returned without relevance scores, and mixed together with facts that have since been superseded by newer information. None of this is configurable.

That's a fine starting point, but once a graph has accumulated months of real memory, deployments need a few knobs. After running Graphiti as long-lived agent memory, these are the four we found ourselves needing:

**1. Choice of ranking strategy.** Different workloads want different orderings. RRF is a solid default; MMR trades a little relevance for more variety; a cross-encoder gives the most accurate relevance ordering at the cost of an extra model call per search. Today the choice is hardcoded — a deployment can't pick.

**2. Hide superseded facts by default.** Graphiti's temporal model marks old facts as invalid when new information contradicts them, but the search tools still return those stale facts ranked equally with current ones. An agent asking "where does X work?" shouldn't have to guess which of three answers is the live one. We propose excluding invalidated facts by default, with a per-call `include_invalidated=true` escape hatch for callers who want history.

**3. Result counts as deployment config.** The default of 10 results is baked into the tool signatures. Deployments feeding results into an LLM context window want to set their own default (we use 6) and still allow per-call overrides.

**4. Relevance scores in results.** Search results come back as a bare list — a client can't tell a strong match from a barely-related one, so it can't apply its own cutoff or weight facts by confidence. Surfacing the reranker's score per result fixes that.

## How this relates to existing issues and PRs

There's an active cluster of work about which **client** performs cross-encoder reranking — #1393 and #1441, with PRs #1637, #1442, #1152, #1174, and #1227. Those all solve "the reranker client is hardcoded to OpenAI and crashes without an OpenAI key."

This proposal is about a different layer: which ranking **strategy** the search uses, and how results are filtered, capped, and presented. The two are complementary — and because the proposed default stays `rrf`, which needs no cross-encoder client at all, nothing here depends on that cluster landing first.

Related observations from other issues:

- #777 reported MMR searches returning empty results. We hit the same failure mode on a live graph: applying a score cutoff to MMR output discards nearly everything, because MMR scores aren't calibrated the way cross-encoder scores are. Our implementation therefore never score-filters MMR results, and the config documents that.
- #1642 documents a candidate-truncation bug in the core cross-encoder path. It's one of the reasons `cross_encoder` is offered as an option here but **not** made the default.
- #1571 asks whether reranking accounts for a fact's temporal validity. The default-on liveness filter in point 2 is a practical partial answer: expired facts don't compete with live ones in the first place.

## Proposal

A `search` config block in the MCP server's `config.yaml` (with matching `SEARCH__*` environment variables):

```yaml
search:
  reranker: rrf            # rrf | mmr | cross_encoder
  mmr_lambda: 0.5          # only used by mmr
  reranker_min_score: 0.0  # only meaningful for cross_encoder
  max_facts: 6             # default result count for search_memory_facts
  max_nodes: 6             # default result count for search_nodes
  exclude_invalidated: true
```

The default reranker stays `rrf`, so ordering behavior is unchanged out of the box. Three defaults are deliberate changes, called out for review: stale facts are excluded by default (with the per-call opt-back-in), results carry a `score` field, and the default result count is 6 rather than 10 (sized for feeding an LLM context window — easy to keep at 10 if preferred, since it's now just config).

I have a working implementation with unit and integration tests, running in production on our deployment. PR to follow shortly.

brentkearney:
## Summary

Implements #1645: makes the MCP server's search behavior configurable instead of hardcoded. Adds a `search` config block (with `SEARCH__*` env vars) that controls which reranker orders results, whether superseded facts are returned, and the default result counts — and surfaces each result's relevance score so clients can judge match quality.

Everything lives in `mcp_server/`; `graphiti_core` is untouched.

## What changes for users

```yaml
search:
  reranker: rrf            # rrf | mmr | cross_encoder
  mmr_lambda: 0.5          # only used by mmr
  reranker_min_score: 0.0  # only meaningful for cross_encoder
  max_facts: 6             # default result count for search_memory_facts
  max_nodes: 6             # default result count for search_nodes
  exclude_invalidated: true
```

- `search_memory_facts` and `search_nodes` order results with the configured reranker. The default stays `rrf`, so out-of-the-box ordering is unchanged. When a `center_node_uuid` is given, the node-distance recipe still wins, as today.
- Facts that Graphiti's temporal model has marked invalid (superseded or expired) are excluded by default. Callers who want history back pass `include_invalidated=true` on the call.
- `max_facts` / `max_nodes` become deployment-level defaults with per-call overrides. The shipped default is 6, sized for feeding results into an LLM context window — happy to keep it at 10 if you'd rather not change the out-of-the-box count.
- Each returned fact/node now carries a `score` field from the reranker, so clients can apply their own cutoffs or weight results by confidence.

## Design notes (learned on a live graph)

- **MMR results are never score-filtered.** MMR's scores aren't calibrated like cross-encoder scores; applying a cutoff to them collapses recall to near zero (same failure mode reported in #777). The config help text documents this, and `reranker_min_score` only applies to `cross_encoder`.
- **`cross_encoder` is offered but not the default.** It needs a configured cross-encoder client, adds a model call per search, and the core edge-search path currently truncates its candidate pool (#1642). The option is there for deployments that want it; the default doesn't depend on it.
- This PR is complementary to the cross-encoder **client** configuration work (#1393/#1441 and PRs #1637/#1442/#1152/#1174/#1227): those choose which client reranks; this chooses which strategy the search uses and how results are filtered and presented. With the default `rrf`, nothing here requires a cross-encoder client at all.

## Files

- `mcp_server/src/config/schema.py` — new `SearchTuningConfig` block
- `mcp_server/src/utils/search_tuning.py` — builds the core `SearchConfig` for the chosen reranker; liveness filter for invalidated facts
- `mcp_server/src/graphiti_mcp_server.py` — search tools wired to the config, per-call overrides
- `mcp_server/src/models/response_types.py`, `mcp_server/src/utils/formatting.py` — `score` field on results
- `mcp_server/README.md` — env-var reference and cross-encoder caveat

## Tests

- `test_search_tuning.py`, `test_search_config_schema.py`, `test_formatting_score.py`, `test_search_tools_wiring.py` — unit tests for the config block, search-config construction, liveness filtering, score formatting, and tool wiring (22 tests)
- `test_search_relevance_int.py` — integration test validating result relevance end-to-end (requires a live backend, skipped otherwise)

This configuration has been running in production on our deployment (Neo4j backend, OpenAI provider).

Closes #1645

### DIFF
--- a/mcp_server/README.md
+++ b/mcp_server/README.md
@@ -258,9 +258,9 @@ nested config, so the env prefix is `SEARCH__` (double underscore).
 
 | Var | Default | Purpose |
 |---|---|---|
-| `SEARCH__RERANKER` | `mmr` | `mmr` (diversity-aware, suppresses recurring hub facts), `rrf` (plain rank fusion), or `cross_encoder`. |
-| `SEARCH__MMR_LAMBDA` | `0.5` | MMR relevance/diversity tradeoff; 1.0 = no diversity. |
-| `SEARCH__RERANKER_MIN_SCORE` | `0.0` | Min score to keep a result. Only meaningful for `cross_encoder`. |
+| `SEARCH__RERANKER` | `rrf` | `rrf` (reciprocal-rank fusion; relevance-ordered and diverse — the reliable default), `mmr` (diversity-aware but miscalibrated for low-magnitude embedding similarities — validate before use), or `cross_encoder`. |
+| `SEARCH__MMR_LAMBDA` | `0.5` | MMR relevance/diversity tradeoff; 1.0 = no diversity. Only used when reranker is `mmr`. |
+| `SEARCH__RERANKER_MIN_SCORE` | `0.0` | Min score to keep a result. Only meaningful for `cross_encoder`; ignored for `mmr` (which must not filter) and left at 0 for `rrf`. |
 | `SEARCH__MAX_FACTS` | `6` | Default result count for `search_memory_facts`. |
 | `SEARCH__MAX_NODES` | `6` | Default result count for `search_nodes`. |
 | `SEARCH__EXCLUDE_INVALIDATED` | `true` | Exclude superseded/expired facts by default. Per-call override: `include_invalidated=true`. |
--- a/mcp_server/README.md
+++ b/mcp_server/README.md
@@ -251,6 +251,25 @@ The `config.yaml` file supports environment variable expansion using `${VAR_NAME
 
 You can set these variables in a `.env` file in the project directory.
 
+### Search tuning (env vars, prefix `SEARCH__`)
+
+These knobs tune the `search_memory_facts` and `search_nodes` tools. They are
+nested config, so the env prefix is `SEARCH__` (double underscore).
+
+| Var | Default | Purpose |
+|---|---|---|
+| `SEARCH__RERANKER` | `mmr` | `mmr` (diversity-aware, suppresses recurring hub facts), `rrf` (plain rank fusion), or `cross_encoder`. |
+| `SEARCH__MMR_LAMBDA` | `0.5` | MMR relevance/diversity tradeoff; 1.0 = no diversity. |
+| `SEARCH__RERANKER_MIN_SCORE` | `0.0` | Min score to keep a result. Only meaningful for `cross_encoder`. |
+| `SEARCH__MAX_FACTS` | `6` | Default result count for `search_memory_facts`. |
+| `SEARCH__MAX_NODES` | `6` | Default result count for `search_nodes`. |
+| `SEARCH__EXCLUDE_INVALIDATED` | `true` | Exclude superseded/expired facts by default. Per-call override: `include_invalidated=true`. |
+
+**Cross-encoder caveat:** `SEARCH__RERANKER=cross_encoder` gives a true relevance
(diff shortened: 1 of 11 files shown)


brentkearney:

brentkearney:
Closing this — not pursuing it.

Two reasons, laid out in full at https://github.com/getzep/graphiti/issues/1645#issuecomment-5153289013: the reranker-choice half was net-zero (RRF was already the effective default, and #1637 has since covered this properly), and `exclude_invalidated: true` as a default discards the bi-temporal information that makes superseded facts useful.

The observability half of #1645 — a `suppressed_invalidated` count and emitting `ranker` alongside `score` — is the part worth keeping, and @Palo-Alto-AI-Research-Lab has offered to PR it. Closing this so there's no conflict on the same tool signatures.

Issue #1283, opened 2026-02-28
Freature: Save episode ids BEFORE completion for previous_episode_uuids


Justin99b:
# The Problem
Lets say we have different knowledge events that happen in quick succession which are all related to each other becasue they are happening in a knowledge gathering process.

Right now the episodes that are supposed to be related might not have been indexed yet and are therefore not being able to get picked up to give related context even when available.

# The Solution
pre save the minimal episodes info like the id, timestamp in the database and have it in a "waiting" state so later episode injestion can reference these in the previous_episode_uuids. This way, we are able to get strong linking between the episodes even when not all of it is processed yet.

This means, We can have a detection that if those relations have been all indexed, we are able to one more processing of it and have the full relations loaded and can therefore have the best understanding between these without needing to rely on embedding after the fact.

# Disclaimer
I am not 100% if i understand the structure fully to have gotten this right, please let me know if my reqeust is BS or if it could actually help. I certainly need something simelar in my current project

Issue #961, opened 2025-10-01
[Feature Request] Improve Episodes API: return UUID, support GET by ID, allow custom metadata


fredpottier:
### 📝 Description

**Context**
While integrating zepai/graphiti:latest into a Qdrant-based ingestion pipeline, we identified several limitations in the current Episodes API that make it difficult to implement reliable synchronization and entity backfill.

**Problem**

POST /messages successfully creates an episode, but only returns { "success": true }.
The generated episode_uuid is not exposed in the response.

No endpoint exists to fetch a specific episode by its UUID (e.g. GET /episode/{uuid} returns 405 Method Not Allowed).
Episodes contain an empty name field and do not allow custom metadata.

As a result:

- It is impossible to map a Qdrant chunk episode_id to its corresponding Graphiti episode.
- Entity backfill from Graphiti to Qdrant cannot be implemented.
- Only /search can be used to access entities, which is inefficient for structured enrichment.

**Impact**

Limits bidirectional sync between Qdrant and Graphiti.
Prevents retroactive enrichment of chunks with canonicalized entities.
Reduces Graphiti’s usefulness as a reliable source of truth for entities/facts.

**Feature Request**

> **Return the episode UUID on creation**

POST /messages should return the generated episode_uuid.

Example:

```
{
  "success": true,
  "episode_uuid": "8ac0480a-fe54-425b-b81b-14b8d1e9bbdf"
}
```

> Add a GET endpoint for episodes by UUID

GET /episode/{uuid} should return the full episode, including entities and facts.

Example response:

```
{
  "uuid": "...",
  "group_id": "...",
  "content": "...",
  "entities": [...],
  "facts": [...]
}
```

> Support custom metadata or user-provided IDs

Allow clients to specify a name or external_id when creating an episode.
This would enable mapping between external stores (e.g. Qdrant chunk IDs) and Graphiti episodes.

**Benefits**

- Reliable sync between Qdrant and Graphiti
- Enables structured backfill of entities into Qdrant
- Improves traceability and interoperability in multi-system architectures

Makes Graphiti usable for production pipelines

**✅ Acceptance Criteria**

1.  POST /messages returns episode_uuid
2.  GET /episode/{uuid} implemented and returns entities/facts
3.  Support for name or external_id field on episode creation

PR #1648, opened 2026-07-13
fix: add_episode creates the episode when a supplied uuid is not found

### LINKED ISSUE #1646: add_episode raises NodeNotFoundError for a new uuid instead of creating the episode — blocks pre-assigned episode ids
## Summary

`Graphiti.add_episode()` accepts a `uuid` parameter, but it only works for episodes that already exist in the graph (it reprocesses them). Pass a uuid that isn't in the graph yet and the call fails with `NodeNotFoundError` instead of creating the episode under that uuid.

This makes it impossible for a caller to choose an episode's id at save time — even though the save path already uses MERGE-on-uuid semantics that would handle a caller-supplied uuid perfectly well.

## Why callers want to pre-assign uuids

Any system that integrates Graphiti with an external store, queue, or pipeline needs to know an episode's id **before or at** save time, not after ingestion finishes:

- #1283 asks for episode ids to be available before processing completes, so that episodes saved in quick succession can reference each other via `previous_episode_uuids`. With pre-assigned uuids this just works: the caller generates the ids, saves episode A with uuid₁, and saves episode B referencing uuid₁ — no new "waiting state" machinery needed.
- #961 asks (among other things) for user-provided ids on episode creation, so external records (e.g. vector-store chunks) can be mapped to their Graphiti episodes. Letting `add_episode` honor a supplied uuid covers that part of the request at the core level.

Our own use case is the same shape: a memory pipeline that writes an episode to Graphiti and records the same id in its own bookkeeping, so a bad ingest can later be deleted and retried by id.

## Current behavior

```python
await graphiti.add_episode(
    name='test',
    episode_body='...',
    source_description='...',
    reference_time=now,
    uuid=str(uuid4()),   # a fresh uuid, not yet in the graph
)
# graphiti_core.errors.NodeNotFoundError: node <uuid> not found
```

## Proposed behavior

- `uuid` exists in the graph → reprocess that episode (unchanged).
- `uuid` supplied but not found → create the episode under the supplied uuid, matching the MERGE-on-uuid semantics the save path already uses.
- `uuid` omitted → generate one (unchanged).

No existing caller is affected: the only path that changes is the one that currently raises.

I have a small fix with tests; PR to follow shortly.

brentkearney:
## Summary

Fixes #1646: `add_episode` currently raises `NodeNotFoundError` when given a `uuid` that isn't in the graph yet. This change makes it create the episode under the supplied uuid instead, so callers can pre-assign episode ids.

Behavior after the change:

- `uuid` exists in the graph → reprocess that episode (unchanged)
- `uuid` supplied but not found → create the episode with the supplied uuid (was: crash)
- `uuid` omitted → generate one (unchanged)

The only code path that changes is the one that currently raises, so no existing caller is affected. Creating under a supplied uuid matches the MERGE-on-uuid semantics the save path already uses.

This also gives callers the primitive requested in #1283 (reference episodes in `previous_episode_uuids` before their processing completes — just pre-assign the ids) and covers the user-provided-id part of #961.

## Changes

- `graphiti_core/graphiti.py` — `add_episode` catches `NodeNotFoundError` on the uuid lookup and falls through to creating the episode with the supplied uuid; docstring updated
- `mcp_server/src/graphiti_mcp_server.py` — `add_memory` docstring updated to document the new uuid semantics
- `tests/test_add_episode_uuid.py` — tests for all three uuid paths (existing uuid reprocesses, new uuid creates with that uuid, omitted uuid generates)

Closes #1646

### DIFF
--- a/graphiti_core/graphiti.py
+++ b/graphiti_core/graphiti.py
@@ -1017,7 +1017,9 @@ class Graphiti:
         group_id : str | None
             An id for the graph partition the episode is a part of.
         uuid : str | None
-            Optional uuid of the episode.
+            Optional uuid of the episode. If an episode with this uuid already
+            exists it is reprocessed; otherwise the episode is created with
+            this uuid.
         update_communities : bool
             Optional. Whether to update communities with new node information
         entity_types : dict[str, BaseModel] | None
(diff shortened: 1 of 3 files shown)


brentkearney:

jaumemir:
This is a critical fix for the library, I am using a monkeypatch, which I don't like. I urge to merge. It seems blocked by an administrative issue (CLA acceptance, when it has been already accepted)

PR #1436, opened 2026-04-24
fix: use directed match in edge_bfs_search to prevent duplicate edges

mrunmayee17:
fix: use directed Cypher match in edge_bfs_search to prevent duplicate edges
                                                                                  
Undirected -(m:Entity) in the BFS Cypher query matches each RELATES_TO edge twice (once as source→target, once as target→source). Because DISTINCT is applied on n.uuid, e, m.uuid, the swapped pair (target, edge, source) passes through as a distinct row, doubling the result count and swapping source/target on every returned edge.

Changing to directed ->(m:Entity) ensures each edge is matched exactly once in its original orientation.
                                                                                  
Fixes #789   

## Summary
edge_bfs_search returned duplicate edges with swapped source_node_uuid/target_node_uuid due to an undirected Cypher match pattern. This fix applies the directed match across all three driver implementations (Neo4j, FalkorDB, Neptune) and both fallback paths in search_utils.py.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Testing
- [x] Unit tests added/updated - test_graphiti_mock.py now asserts exact edge count of 2 and verifies source_node_uuid/target_node_uuid orientation on returned edges
- [ ] Integration tests added/updated
- [x] All existing tests pass - Neo4j and FalkorDB require live DB and are validated in CI.

## Breaking Changes
- [ ] This PR contains breaking changes

If this is a breaking change, describe:
- What functionality is affected
- Migration path for existing users

## Checklist
- [x] Code follows project style guidelines (`make lint` passes)
- [x] Self-review completed
- [x] Documentation updated where necessary
- [x] No secrets or sensitive information committed

## Related Issues
Closes #789

Issue #789, opened 2025-07-30
[BUG] BFS Returns Duplicate Edges with Swapped Source/Target Nodes


paveljakov:
## Bug Description
The BFS edge search returns duplicate relationships where the same edge UUID appears twice but with the source and target node UUIDs swapped. This occurs because the BFS implementation uses directed traversal but undirected relationship matching, causing each relationship to be found in both directions.

## Steps to Reproduce
Provide a minimal code example that reproduces the issue:

```python
from graphiti_core.search.search import search
from graphiti_core.search.search_config import SearchConfig, EdgeSearchConfig, EdgeSearchMethod
from graphiti_core.search.search_filters import SearchFilters

# Configure search to use BFS
config = SearchConfig(
    edge_config=EdgeSearchConfig(
        search_methods=[EdgeSearchMethod.bfs],
        bfs_max_depth=1
    )
)

# Perform search using Graphiti's native search method
results = await graphiti.search_(
    query="spare parts",
    group_ids=["inventory"],
    config=config,
    search_filter=SearchFilters(),
    bfs_origin_node_uuids=["766fb29e-7868-4430-bc0c-007572e0db57"]
)

# Check for duplicates with same UUID but swapped nodes
edge_uuids = {}
for edge in results.edges:
    key = f"{edge.source_node_uuid}->{edge.target_node_uuid}"
    
    if edge.uuid in edge_uuids:
        print(f"Duplicate found for UUID {edge.uuid}:")
        print(f"  Original: {edge_uuids[edge.uuid]}")
        print(f"  Duplicate: {key}")
    
    edge_uuids[edge.uuid] = key

# You'll see output like:
# Duplicate found for UUID 209e9060-530f-4e79-a5f2-d92691b711f3:
#   Original: 766fb29e-7868-4430-bc0c-007572e0db57->28db29b8-f08a-4061-8511-38903846caf4
#   Duplicate: 28db29b8-f08a-4061-8511-38903846caf4->766fb29e-7868-4430-bc0c-007572e0db57
```

## Expected Behavior
Each relationship should appear only once in the BFS results, maintaining the directionality established during the path traversal. A relationship with UUID `X` should only appear once, either as `A->B` or `B->A`, not both.

## Actual Behavior
The same relationship appears twice with the same UUID but with swapped source and target nodes:

```json
[
  {
    "uuid": "209e9060-530f-4e79-a5f2-d92691b711f3",
[...]

PR #1653, opened 2026-07-15
fix(search): consume bfs relationship hits directly in edge_bfs_search

kramerica-inc-dev:
# fix(search): consume bfs relationship hits directly in edge_bfs_search

## Summary
Completes #1500 on the path that executes today. #1500 rewrites the edge fulltext and BFS queries to consume the relationships the index/path already returned (`startNode`/`endNode`) instead of re-matching every hit by uuid against the whole graph — but its BFS hunk only lands in `graphiti_core/driver/falkordb/operations/search_ops.py`. That module is not reached at runtime: `driver.search_interface` is declared (`graphiti_core/driver/driver.py`) and never assigned anywhere in `graphiti_core`, so `edge_bfs_search` always falls through to the generic query in `graphiti_core/search/search_utils.py`. This PR applies the exact same transform there:

```diff
             UNWIND relationships(path) AS rel
-            MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]-(m:Entity)
+            WITH rel AS e, startNode(rel) AS n, endNode(rel) AS m
+            WHERE type(e) = 'RELATES_TO'
+            WITH e, n, m
```

Notes on the transform:
- **MENTIONS filter kept explicit.** The BFS pattern traverses `RELATES_TO|MENTIONS`, so `rel` can be a MENTIONS edge; the old re-MATCH filtered those out implicitly via the `e:RELATES_TO` label. The `WHERE type(e) = 'RELATES_TO'` guard preserves that (same as #1500's BFS hunk). The trailing `WITH e, n, m` keeps the appended filter clause (`WHERE …`) valid and the `e`/`n`/`m` bindings, filter constructor, and `RETURN` untouched.
- **Direction/duplicates.** The old undirected re-MATCH matched each edge twice (once per orientation), and `RETURN DISTINCT` could not collapse the swapped source/target rows. `startNode`/`endNode` yields one row per relationship in its original orientation, so this also removes the duplicated/swapped-edge symptom reported in #789 and targeted by PR #1436 — on this path. #789 additionally flags the same undirected pattern in `get_embeddings_for_edges`, which is out of scope here, so this PR does not claim to close it.
- **Scope.** Only the generic (non-Kuzu, non-Neptune) BFS branch is changed. Kuzu builds its own query earlier in the function and Neptune has its own branch; both are untouched, as is the fulltext query in this file (that hunk belongs to #1500). `startNode()`/`endNode()` are native on both engines that share this branch — Neo4j and FalkorDB (the Neptune branch in this same function already uses them in its `RETURN`).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Performance improvement
- [ ] Documentation/Tests

## Objective
The re-MATCH forces an O(matches × graph) label scan per BFS hit instead of consuming the relationship the path traversal already produced. On our production FalkorDB graph (~2.6k Entity nodes), removing this same antipattern on the fulltext path took dense-document episode processing from 15–25 min down to 1–4 min; `GRAPH.PROFILE` of the rewritten query resolves 572 hits in <2 ms with no scan operator (measured 2026-07-14). The BFS variant fixed here is the same shape, made worse by the undirected match doubling every row.

## Testing
- [x] All existing tests pass

Added `tests/utils/search/test_edge_bfs_query_shape.py`, a DB-free query-shape regression test mirroring the RecordingExecutor approach of #1500's `tests/test_falkordb_search_ops.py`: a recording driver captures the emitted Cypher. Asserts: no `uuid: rel.uuid` re-MATCH in the BFS query; `startNode`/`endNode` consumption present; explicit `type(e) = 'RELATES_TO'` guard present; `e`/`n`/`m` bindings still reach filters and `RETURN`; and the fulltext query is not rewritten by this change (formulated so it stays green after #1500 merges).

```
DISABLE_FALKORDB=1 DISABLE_KUZU=1 DISABLE_NEPTUNE=1 python -m pytest \
  tests/utils/search/test_edge_bfs_query_shape.py \
  tests/utils/search/test_search_security.py -q -m "not integration"
12 passed, 1 warning in 0.09s
```

## Breaking Changes

No. Query results are identical except that the duplicated/swapped-orientation rows produced by the old undirected re-MATCH are gone; each edge is now returned once, in its stored direction.

## Checklist
- [x] Code follows project style guidelines (`make lint` passes — `ruff format --check` + `ruff check` clean on changed files)
- [x] Self-review completed
- [ ] Documentation updated where necessary (n/a)

## Related Issues
Refs #1272, #1500 (completes it on the live generic path). Related to #789 and PR #1436 (resolves the duplicate/swapped-edge symptom for `edge_bfs_search`; `get_embeddings_for_edges` from #789 remains).


Naseem77:
@kramerica-inc-dev Verified live on FalkorDB: identical BFS results, MENTIONS edges still filtered by the type guard, and the O(matches x graph) re-MATCH is gone from the emitted query. This completes #1500 on the path that actually runs.

kramerica-inc-dev:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 59801654+kramerica-inc-dev@users.noreply.github.com

kramerica-inc-dev:

kramerica-inc-dev:
recheck

kramerica-inc-dev:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: kramerica-inc-dev@users.noreply.github.com

kramerica-inc-dev:
recheck

prasmussen15 (approved):

prasmussen15 (approved):

PR #1655, opened 2026-07-15
avoid mutating shared search config

adolphinvx:
## Summary

- copy the shared edge search recipe before applying the per-request result limit
- prevent `Graphiti.search()` calls from mutating module-level search configuration
- add regression coverage for searches with and without a center node

## Testing

- `pytest tests/test_graphiti_search.py -q`
- `make lint`
- unit test suite: 341 passed, 11 skipped

## API / schema impact

None.


adolphinvx:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: exarunner@gmail.com

PR #1658, opened 2026-07-17
fix(communities): scope remove_communities to the group_ids being rebuilt

### LINKED ISSUE #1657: build_communities(group_ids=...) wipes ALL communities, then rebuilds only the selected groups
## Problem

`Graphiti.build_communities(group_ids=[...])` documents scoped behavior:

> `group_ids : list[str] | None` — Optional. Create communities only for the listed group_ids.

But the implementation first clears communities with **no group filter**:

```python
# graphiti.py — build_communities
# Clear existing communities
await remove_communities(driver)
```

```python
# utils/maintenance/community_operations.py
async def remove_communities(driver: GraphDriver):
    ...
    await driver.execute_query(
        """
        MATCH (c:Community)
        DETACH DELETE c
        """
    )
```

`remove_communities` deletes **every Community node in the database**, then the rebuild recreates communities only for the selected groups. So a scoped rebuild silently **destroys all other groups' communities** — the opposite of the documented contract, and invisible unless you happen to census the other groups afterwards.

## Repro

On a graph with communities in multiple `group_id`s:

```python
await graphiti.build_communities(group_ids=["group_a"])
# → group_a communities rebuilt; every other group's communities are GONE
```

Observed live on a 5-group graph (Neo4j 5.26, graphiti-core 0.29.2; code unchanged on `main` at time of filing): rebuilding 2 stale groups deleted the other 3 groups' communities (71 Community nodes) with no error or log line.

## Impact

Any multi-tenant / multi-namespace deployment that rebuilds communities per group — e.g. refreshing only the tenant that changed, which is exactly what per-group scoping invites — loses every other tenant's communities on each call. Community summaries are LLM-generated, so recovering them costs a full re-run across all groups.

## Suggested fix

Thread `group_ids` through the clear step and scope the delete when provided:

```python
async def remove_communities(driver: GraphDriver, group_ids: list[str] | None = None):
    ...
    if group_ids:
        await driver.execute_query(
            """
            MATCH (c:Community)
            WHERE c.group_id IN $group_ids
            DETACH DELETE c
            """,
            group_ids=group_ids,
        )
    else:
        await driver.execute_query("MATCH (c:Community) DETACH DELETE c")
```

…and pass `group_ids` from `build_communities`. The `graph_operations_interface.remove_communities(driver)` fast path needs the same parameter so driver-specific implementations can scope too.

Happy to open a PR if useful.

Saltasm:
## Problem

Fixes #1657.

`build_communities(group_ids=[...])` documents scoped behavior — *"Create communities only for the listed group_ids"* — but its clear step deletes **every Community node in the database**:

```python
# graphiti.py — build_communities
await remove_communities(driver)   # no group filter
```

```cypher
MATCH (c:Community)
DETACH DELETE c
```

The rebuild then recreates communities only for the selected groups, so a scoped rebuild silently destroys all other groups' communities. Observed live on a 5-group graph: rebuilding 2 stale groups deleted the other 3 groups' 71 communities with no error or log line. Any multi-tenant deployment refreshing communities per tenant hits this on every call, and recovery costs a full LLM re-summarization across all groups.

## Fix

Thread `group_ids` through the clear step:

- `community_operations.remove_communities(driver, group_ids=None)` — scoped `DETACH DELETE` with `WHERE c.group_id IN $group_ids` when `group_ids` is provided; the existing full wipe when it is not (blank/None keeps the documented whole-graph behavior).
- Same optional parameter on `GraphOperationsInterface.remove_communities` and all four driver implementations (neo4j, falkordb, kuzu, neptune) — each mirrors the `group_ids is None` branching already used by sibling operations in the same files.
- `build_communities` passes its `group_ids` through.

## Tests

`tests/utils/maintenance/test_remove_communities.py` (mocked driver, no DB):
- unscoped call still deletes all (unchanged behavior)
- scoped call emits the `WHERE c.group_id IN $group_ids` form with the right params
- `group_ids=[]` behaves like None (no scoping requested, matching `build_communities`' blank semantics)
- the `graph_operations_interface` fast path receives `group_ids`


Naseem77:
@Saltasm Reproduced the data loss: rebuilding communities for one group wiped every group's communities. With your patch only the requested group is deleted (verified live on FalkorDB) and all backends' ops are covered.

prasmussen15 (approved):

PR #1663, opened 2026-07-17
fix: preserve relationship direction in get_by_node_uuid and BFS edge search

### LINKED ISSUE #1662: [BUG] : EntityEdge.get_by_node_uuid returns swapped source/target when queried from the target side
## Bug Description

  `EntityEdge.get_by_node_uuid` returns `source_node_uuid` and `target_node_uuid` swapped relative to the stored
  relationship direction, whenever the queried node is the relationship's actual target. The same class of bug affects
  `edge_bfs_search`.

  ## Steps to Reproduce

  ```python
  from datetime import datetime
  from graphiti_core.nodes import EntityNode

  # Two entity nodes
  alice = EntityNode(name='Alice', group_id='g', summary='alice', created_at=datetime.now())
  bob = EntityNode(name='Bob', group_id='g', summary='bob', created_at=datetime.now())
  await alice.save(driver)
  await bob.save(driver)

  # Edge with explicit direction: alice -> bob
  edge = EntityEdge(
      source_node_uuid=alice.uuid,
      target_node_uuid=bob.uuid,
      name='likes',
      fact='Alice likes Bob',
      episodes=[],
      created_at=datetime.now(),
      group_id='g',
  )
  await edge.save(driver)

  # Query from the TARGET side
  retrieved = await EntityEdge.get_by_node_uuid(driver, bob.uuid)
  print(retrieved[0].source_node_uuid)  # actual:   bob.uuid   (the queried node)
  print(retrieved[0].target_node_uuid)  # actual:   alice.uuid

  Expected Behavior

  The returned edge should preserve the stored relationship direction:
  - source_node_uuid == alice.uuid
  - target_node_uuid == bob.uuid

  Actual Behavior

  The returned edge has its direction reversed:
  - source_node_uuid == bob.uuid (the queried node, which is the actual target)
  - target_node_uuid == alice.uuid

  No exception is raised — this is a silent logic bug.

  Environment

  - Graphiti Version: 0.29.2
  - Python Version: 3.13
  - Operating System: Windows 11
  - Database Backend: Affects Neo4j, FalkorDB, and Neptune (KUZU is unaffected — see below)
  - LLM Provider & Model: N/A (bug is independent of LLM)

  Installation Method

  - [ ] pip install
  - [ ] uv add

  Error Messages/Traceback

  No traceback — the query succeeds but returns incorrect source_node_uuid/target_node_uuid values.

  Configuration

  Default driver initialization. Reproduces with any of the affected backends, e.g.:

  from graphiti_core.driver.neo4j_driver import Neo4jDriver

  driver = Neo4jDriver(
      uri='bolt://localhost:7687',
      user='neo4j',
      password='...',
  )

  Additional Context

  - Consistency: Happens consistently for any relationship where the queried node is the target side.
  - Component: Core library (graphiti_core.edges.EntityEdge and per-driver operation classes).
  - Root cause: get_entity_edge_return_query in graphiti_core/models/edges/edge_db_queries.py reads source/target from
  the MATCH variables n/m:

  n.uuid AS source_node_uuid,
  m.uuid AS target_node_uuid,

  - But get_by_node_uuid uses an undirected MATCH:

  MATCH (n:Entity {uuid: $node_uuid})-[e:RELATES_TO]-(m:Entity)

  - Cypher binds the queried node to n on every row regardless of the actual relationship direction, so when the queried
   node is the relationship's target, n is bound to the target and the returned source/target are swapped.
  - Affected call sites:
    - EntityEdge.get_by_node_uuid — Neo4j / FalkorDB / Neptune implementations, plus the fallback in
  graphiti_core/edges.py
    - edge_bfs_search — graphiti_core/driver/{neo4j,falkordb}/operations/search_ops.py
    - edge_bfs_search — graphiti_core/search/search_utils.py:548-563
  - Existing precedent: The NEPTUNE branch of edge_bfs_search at search/search_utils.py:520-546 already inlines
  startNode(e)/endNode(e), which is the correct pattern for undirected MATCH.
  - KUZU is unaffected: Its get_by_node_uuid MATCH is already directed, so n/m reflect the true source/target. (KUZU has
   a separate limitation of only matching outgoing edges, but that's a different issue.)

  Possible Solution

  Switch the NEO4J/FALKORDB/NEPTUNE branches of get_entity_edge_return_query to read source/target from the relationship
   itself via Cypher's built-in startNode(e)/endNode(e):

  startNode(e).uuid AS source_node_uuid,
  endNode(e).uuid   AS target_node_uuid,

  This works regardless of MATCH directionality and matches the pattern Neptune's BFS already uses. KUZU stays unchanged
   since its e is a RelatesToNode_ intermediate node (not a relationship) and its MATCH is already directed.

  Backward compatibility: startNode(e)/endNode(e) reads relationship metadata that is immutable at write time, so
  historical data reads correctly without any backfill. The only behavior change is the bug fix itself in the scenarios
  above.

6or7:
## Summary

  Fixes #1662

  `get_entity_edge_return_query` now reads `source_node_uuid`/`target_node_uuid` from the relationship itself via
  Cypher's `startNode(e)`/`endNode(e)` for NEO4J/FALKORDB/NEPTUNE, instead of from the `n`/`m` MATCH variables.

  This fixes direction reversals in:
  - `EntityEdge.get_by_node_uuid` — when the queried node is the relationship's target
  - `edge_bfs_search` (Neo4j/FalkorDB/search_utils) — same class of bug for undirected MATCH

  KUZU is unchanged: its `e` is a `RelatesToNode_` intermediate node, not a relationship, so `startNode`/`endNode` don't
   apply. KUZU's MATCH is always directed, so `n`/`m` already reflect the true source/target.

  ## Root cause

  For undirected MATCH like `MATCH (n {uuid: $node_uuid})-[e]-(m)`, Cypher binds the queried node to `n` on every row.
  The old `RETURN n.uuid AS source_node_uuid, m.uuid AS target_node_uuid` therefore reported the queried node as source
  regardless of the stored relationship direction. See #1662 for full analysis.

  ## Changes

  - `graphiti_core/models/edges/edge_db_queries.py`: switch NEO4J/FALKORDB/NEPTUNE branches of
  `get_entity_edge_return_query` to `startNode(e)/endNode(e)`. KUZU branch unchanged with an explanatory comment.
  - `tests/test_edge_db_queries.py`: new unit tests covering the return shape per provider.
  - `tests/test_edge_int.py`: regression assertion — `get_by_node_uuid(target_uuid)` must preserve `source=A, target=B`.

  No call sites needed to change — the fix lives entirely in the return-query builder.

  ## Backward compatibility

  No data migration required. `startNode(e)/endNode(e)` reads immutable relationship metadata, so historical data reads
  correctly without any backfill.

  ## Out of scope

  KUZU's `get_by_node_uuid` only matches outgoing edges due to its directed MATCH pattern. Fixing this requires storing
  `source_uuid`/`target_uuid` on `RelatesToNode_` and a data backfill — left for a follow-up PR.


6or7:

Naseem77:
@6or7 Reproduced the swap: reading an edge via its target node returned the endpoints reversed. With your startNode/endNode change the stored direction is preserved (verified live on FalkorDB), and the Kuzu special case is handled correctly.

prasmussen15 (approved):

prasmussen15 (approved):

PR #1667, opened 2026-07-18
Return entity edge reference time

### LINKED ISSUE #1661: EntityEdge.reference_time is write-only: persisted on every edge, never returned by any read path
**Version:** `graphiti-core` 0.29.1 (observed on FalkorDB; the return-query code path looks provider-independent)

## Summary

`EntityEdge.reference_time` is written to the database but can never be read back. Every edge loaded from any read path has `reference_time=None`, regardless of what is stored.

## Reproduction

```python
result = await g.add_episode(..., reference_time=datetime(2024, 12, 9))
# stored correctly -- direct Cypher confirms:
#   MATCH ()-[e:RELATES_TO]->() WHERE e.reference_time IS NOT NULL RETURN count(e)  -> 72

edges = (await g.search_("...", group_ids=["t"])).edges
[e.reference_time for e in edges]   # -> [None, None, None, ...]
```

## Cause

Two independent places drop it.

**1. No RETURN projection selects it.** `graphiti_core/models/edges/edge_db_queries.py::get_entity_edge_return_query` returns `uuid`, `source_node_uuid`, `target_node_uuid`, `group_id`, `created_at`, `name`, `fact`, `episodes`, `expired_at`, `valid_at`, `invalid_at`, `properties(e) AS attributes` — no `reference_time`. The fulltext/similarity search queries in `graphiti_core/search/search_utils.py` project the same field list.

**2. Hydration discards the fallback.** `graphiti_core/edges.py::get_entity_edge_from_record` reads:

```python
reference_time=parse_db_date(record.get('reference_time')),
```

`record` has no `reference_time` key (per 1), so this is always `None`. The value *is* present inside `properties(e) AS attributes` — but a few lines earlier it is explicitly removed:

```python
attributes.pop('reference_time', None)
```

So it is popped out of the one place it survived and then looked up somewhere it never existed. Net effect: unrecoverable from a search result.

## Impact

`reference_time` is the only per-edge record of *which episode's timestamp produced this fact*. For temporal work it's valuable precisely when `valid_at` is absent — a fact extracted from a rule that states no explicit date is still known to have been asserted on that document's publication date. In our corpus 28 of 89 edges had no `valid_at`; `reference_time` would have grounded 20 of them. Because it reads back as `None`, that inference is unavailable, and an ungrounded `valid_at` otherwise reads as "true since the beginning of time".

## Suggested fix

Add `e.reference_time AS reference_time` to `get_entity_edge_return_query` (and the search projections in `search_utils.py`). Keeping the `attributes.pop` is then correct, since the value arrives as a first-class field.

## Workaround

Re-query it by edge uuid after search and assign it onto the hydrated models.

sakshichitnis27:
## Summary
  Fixes `EntityEdge.reference_time` being lost when edges are read from the database.

  Although `reference_time` is persisted on entity edges, it was omitted from the
  Cypher return projections. As a result, edge hydration always received a missing
  value and set `reference_time` to `None`.

  This PR:

  - Adds `reference_time` to the shared entity-edge return projection.
  - Adds it to the standalone Neptune full-text, similarity, and BFS search
  projections.
  - Adds regression tests for query projection and edge hydration.

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Performance improvement
  - [ ] Documentation/Tests

  ## Objective

  Ensure persisted `EntityEdge.reference_time` values are preserved across standard
  edge reads and search results.

  ## Testing

  - [ ] All existing tests pass

  Regression tests verify that:

  - Entity-edge return queries select `reference_time` for every graph provider.
  - Both entity-edge hydration paths preserve the returned timestamp.

  The full test suite was not run locally due to environment limitations.

  ## Breaking Changes

  No breaking changes.

  ## Checklist

  - [x] Self-review completed

  No documentation changes were necessary because this restores the existing model
  field’s expected read behavior.

  ## Related Issues

  Closes #1661

### DIFF
--- a/graphiti_core/models/edges/edge_db_queries.py
+++ b/graphiti_core/models/edges/edge_db_queries.py
@@ -202,6 +202,7 @@ def get_entity_edge_return_query(provider: GraphProvider) -> str:
         e.expired_at AS expired_at,
         e.valid_at AS valid_at,
         e.invalid_at AS invalid_at,
+        e.reference_time AS reference_time,
         properties(e) AS attributes
     """
 
@@ -217,6 +218,7 @@ def get_entity_edge_return_query(provider: GraphProvider) -> str:
         e.expired_at AS expired_at,
         e.valid_at AS valid_at,
         e.invalid_at AS invalid_at,
+        e.reference_time AS reference_time,
     """ + (
         'e.attributes AS attributes'
         if provider == GraphProvider.KUZU
--- a/graphiti_core/search/search_utils.py
+++ b/graphiti_core/search/search_utils.py
@@ -253,6 +253,7 @@ async def edge_fulltext_search(
                     e.expired_at AS expired_at,
                     e.valid_at AS valid_at,
                     e.invalid_at AS invalid_at,
+                    e.reference_time AS reference_time,
                     properties(e) AS attributes
                 ORDER BY score DESC LIMIT $limit
                             """
@@ -398,6 +399,7 @@ async def edge_similarity_search(
                     r.expired_at AS expired_at,
                     r.valid_at AS valid_at,
                     r.invalid_at AS invalid_at,
+                    r.reference_time AS reference_time,
                     properties(r) AS attributes
                 ORDER BY i.score DESC
                 LIMIT $limit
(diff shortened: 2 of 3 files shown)


sakshichitnis27:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: sakshigchitnis27@gmail.com

PR #1669, opened 2026-07-20
feat(mcp): use a provider-matched cross-encoder to avoid a hard OpenAI dependency

abouchard11:
## Problem

The MCP server builds an LLM client and an embedder client from config, but never passes a `cross_encoder` to `Graphiti()`. When `Graphiti()` receives no cross-encoder it hard-defaults to `OpenAIRerankerClient()`, which requires `OPENAI_API_KEY`.

The result: a deployment configured entirely for a non-OpenAI provider (e.g. Gemini LLM + Gemini embedder) still fails to start without an OpenAI key, purely because of the reranker default.

## Fix

Add a `CrossEncoderFactory` that mirrors the existing `LLMClientFactory` / `EmbedderFactory`:

- for the `gemini` provider it returns a `GeminiRerankerClient` (already shipped in `graphiti-core`) built from the configured Gemini API key;
- every other provider returns `None`, which preserves the current default (`graphiti-core` applies `OpenAIRerankerClient()`).

The result is wired into both the FalkorDB and Neo4j `Graphiti()` constructors.

## Compatibility

No behavior change for existing OpenAI users: unmatched providers return `None`, so `graphiti-core` keeps applying its `OpenAIRerankerClient()` default exactly as before. An all-Gemini deployment can now start without an `OPENAI_API_KEY`.

## Verification

- Both changed files byte-compile.
- Imports resolve in the `providers` extra (`HAS_GEMINI_RERANKER = True`).
- `gemini` → constructs a `GeminiRerankerClient`; `anthropic`/other → returns `None` (default preserved).

## Scope

Two files, additive:

- `mcp_server/src/services/factories.py` — new `CrossEncoderFactory` + a guarded `GeminiRerankerClient` import.
- `mcp_server/src/graphiti_mcp_server.py` — build the cross-encoder via the factory and pass it into both `Graphiti()` constructors.

### DIFF
--- a/mcp_server/src/graphiti_mcp_server.py
+++ b/mcp_server/src/graphiti_mcp_server.py
@@ -37,7 +37,12 @@ from models.response_types import (
     SuccessResponse,
     TripletResponse,
 )
-from services.factories import DatabaseDriverFactory, EmbedderFactory, LLMClientFactory
+from services.factories import (
+    CrossEncoderFactory,
+    DatabaseDriverFactory,
+    EmbedderFactory,
+    LLMClientFactory,
+)
 from services.queue_service import QueueService
 from utils.formatting import format_fact_result, to_edge_result, to_node_result
 from utils.type_config import (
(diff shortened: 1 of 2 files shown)


abouchard11:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: alex@midnightdev.dev

abouchard11:

Naseem77:
@abouchard11 Tested the factory: Gemini gets a proper reranker, but Anthropic/other providers return None and still fall back to the OpenAI reranker, so the hard OPENAI_API_KEY requirement remains for them. #1637 covers all providers with a local BGE fallback and tests, probably best to consolidate there.

abouchard11:
Consolidating into #1637 as @Naseem77 suggested — it's a superset of this change (provider inference including the embedder, local BGE fallback, tests) and covers the Anthropic case flagged above. I've tested #1637 against my production Neo4j + all-Gemini stack and left a review with findings there. Closing.

PR #1670, opened 2026-07-20
fix(falkordb): route single group_id search to the matching graph

### LINKED ISSUE #1659: FalkorDB: add_episode re-binds the driver but single-group_id search does not, so writes and reads use different graphs
**Version:** `graphiti-core` 0.29.1, FalkorDB driver

## Summary

On FalkorDB, `add_episode` routes a write to the graph named after `group_id`, but a `search` for a **single** `group_id` does not route the same way. If the driver's database differs from the `group_id`, data is written to one graph and searched for in another — and the search silently returns nothing.

## Reproduction

Driver bound to a base database (`reggraph`), data ingested under `group_id="acptprobe"`:

```python
g = Graphiti(graph_driver=FalkorDriver(..., database="reggraph"), ...)
await g.add_episode(..., group_id="acptprobe")   # -> lands in graph `acptprobe`

# fresh process, same construction:
await g.search("...", group_ids=["acptprobe"])                  # -> 0 results
await g.search("...", group_ids=["acptprobe", "anything_else"]) # -> 5 results
```

Adding a **second, non-existent** group_id makes the search succeed. That's the tell.

## Cause

`add_episode` (`graphiti_core/graphiti.py`, ~line 1081) re-binds the shared driver when the group_id differs from the current database:

```python
```

`search`/`search_` instead rely on `@handle_multiple_group_ids` (`graphiti_core/decorators.py`), which only clones when there is more than one group_id:

```python
if (... and group_ids and len(group_ids) > 1):
    ... driver.clone(database=gid)
```

With one group_id the decorator is a pass-through, so the query runs against whatever `self.driver._database` happens to be. The write path routes by group_id; the read path routes by driver state.

## Consequences

1. **Reads return nothing in a fresh process.** The mutation in `add_episode` is what makes reads work afterwards, so search only succeeds if an ingest for that same group_id happened earlier *in the same process*. Ingest-then-search tests pass; a server that ingests and queries in separate processes silently returns empty results.
2. **`add_episode` mutates shared state.** `self.driver` is replaced for the lifetime of the instance, so an ingest for tenant A re-points subsequent single-group_id reads for tenant B. With concurrent multi-tenant ingest on one instance this is a race.

## Suggested fix

Route reads by group_id the same way writes are routed — clone for a single group_id too, not just for `len(group_ids) > 1`. Scoping the clone to the call rather than reassigning `self.driver` would also remove the shared-state mutation.

## Workaround

Bind the driver's `database` to the `group_id` so the two always agree.

Solaris-star:
## Summary
On FalkorDB, `add_episode` re-binds `self.driver` when `group_id` differs from the driver's database, but `@handle_multiple_group_ids` only clones when `len(group_ids) > 1`. A single-`group_id` search in a fresh process therefore queries the default graph and silently returns nothing. Adding a second dummy group_id made search work — that was the tell.

## Change
- For FalkorDB with **one** `group_id` that differs from `driver._database`, pass a **call-scoped** `driver.clone(database=gid)` into the wrapped method (no shared-state reassignment)
- Keep multi-`group_id` concurrent clone + merge behavior
- Unit tests covering single/multi/same-db/non-Falkor paths

## Test plan
- [x] `pytest tests/test_handle_multiple_group_ids.py -q` → 4 passed
- [ ] CI full suite

Fixes #1659

### DIFF
--- a/graphiti_core/decorators.py
+++ b/graphiti_core/decorators.py
@@ -30,6 +30,11 @@ def handle_multiple_group_ids(func: F) -> F:
     """
     Decorator for FalkorDB methods that need to handle multiple group_ids.
     Runs the function for each group_id separately and merges results.
+
+    Also routes a *single* group_id to the matching FalkorDB graph via a
+    call-scoped driver clone. Without this, add_episode re-binds the shared
+    driver for writes while search/retrieve with one group_id query the
+    driver's default database and silently return empty results (#1659).
     """
 
     @functools.wraps(func)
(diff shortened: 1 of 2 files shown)


Solaris-star:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 820622658@qq.com

Naseem77:
@Solaris-star Verified live: single-group retrieve_episodes and search_ returned nothing on main and return correct results with your patch, and the call-scoped clone avoids the shared-state mutation that broke earlier attempts. This supersedes #1326/#1572/#1170 nicely.

prasmussen15 (approved):

PR #1673, opened 2026-07-22
resolve cross-list duplicate citations in edge extraction

### LINKED ISSUE #1672: resolve_extracted_edge's range guard discards a correctly-cited cross-list duplicate, leaking duplicate edges
**Summary.** `resolve_extracted_edge` builds one continuous idx namespace across `related_edges` (duplicate candidates) + `existing_edges` (invalidation candidates), and the judge is told duplicates should only ever be cited from the first list. But when `related_edges` is empty and the true duplicate only surfaces via the broader `existing_edges` search, the judge — reasonably, since it's looking at one continuous list — sometimes cites the correct idx anyway. The range guard then discards it as "invalid" purely because it falls outside `len(related_edges)`, and the new edge is kept as if it were novel. A real duplicate silently leaks into the graph.

**Reproduction.** Live, in a real app (not synthetic), reproduced end-to-end:

- User states a fact (`related_edges=[]`, several unrelated `existing_edges` from a broader search).
- A later, overlapping/concurrent `add_episode` call restates the same fact in different words. This time `related_edges` is *still* empty (the pair-scoped duplicate-candidate search hadn't yet indexed the sibling edge from the concurrent draw), but `existing_edges` now contains the duplicate at idx 0 (broader invalidation-candidate search did find it).
- Judge response: `duplicate_facts=[0]`, reasoning: *"...exactly matches the fact at idx 0 in the invalidation candidates list...this fact is a duplicate..."* — correct.
- Guard: `invalid_duplicates = [i for i in duplicate_facts if i < 0 or i >= len(related_edges)]` → `[0]` (since `len(related_edges) == 0`). Logs a warning, drops it.
- Result: `resolved_edge` stays the brand-new extracted edge, `n_invalidated == 0`. Duplicate now exists twice in the graph.

Confirmed **patch-independent** — reproduces identically against the stock `EdgeDuplicate` schema (no custom reasoning field), so this isn't an artifact of any downstream fork.

**Relevant code** (`graphiti_core/utils/maintenance/edge_operations.py::resolve_extracted_edge`):

```python
llm_response = await llm_client.generate_response(
    prompt_library.dedupe_edges.resolve_edge(context),
    response_model=EdgeDuplicate,
    model_size=ModelSize.small,
    prompt_name='dedupe_edges.resolve_edge',
)
response_object = EdgeDuplicate(**llm_response)
duplicate_facts = response_object.duplicate_facts

# Validate duplicate_facts are in valid range for EXISTING FACTS
invalid_duplicates = [i for i in duplicate_facts if i < 0 or i >= len(related_edges)]
if invalid_duplicates:
    logger.warning(
        'LLM returned invalid duplicate_facts idx values %s (valid range: 0-%d for EXISTING FACTS)',
        invalid_duplicates,
        len(related_edges) - 1,
    )

duplicate_fact_ids: list[int] = [i for i in duplicate_facts if 0 <= i < len(related_edges)]

resolved_edge = extracted_edge
for duplicate_fact_id in duplicate_fact_ids:
    resolved_edge = related_edges[duplicate_fact_id]
    break
```

The prompt's own instructions tell the model duplicates should only ever come from the `EXISTING FACTS` (i.e. `related_edges`) list — but nothing enforces that on the model side, and the guard's only response to a violation is to silently drop the citation rather than use it.

**Suggested fix.** Rather than treating an out-of-`related_edges`-range `duplicate_facts` idx as noise, distinguish "cross-list duplicate citation" (a real signal — idx falls within the invalidation-candidate range) from "genuinely out of range" (idx is negative or beyond the total idx space):

```python
max_valid_idx = len(related_edges) + len(existing_edges) - 1

duplicate_fact_ids = [i for i in duplicate_facts if 0 <= i < len(related_edges)]
cross_list_duplicates = [
    i for i in duplicate_facts if len(related_edges) <= i <= max_valid_idx
]
truly_invalid = [i for i in duplicate_facts if i < 0 or i > max_valid_idx]

if truly_invalid:
    logger.warning(
        'LLM returned out-of-range duplicate_facts idx values %s (valid range: 0-%d)',
        truly_invalid, max_valid_idx,
    )

resolved_edge = extracted_edge
for duplicate_fact_id in duplicate_fact_ids:
    resolved_edge = related_edges[duplicate_fact_id]
    break
if not duplicate_fact_ids and cross_list_duplicates:
    # Duplicate cited only against an invalidation candidate — merge into it
    # directly instead of discarding a correct verdict as "invalid."
    resolved_edge = existing_edges[cross_list_duplicates[0] - len(related_edges)]
```

(`max_valid_idx` is already computed a few lines later for `contradicted_facts` — this just hoists it above the `duplicate_facts` handling.)

This doesn't require changing the prompt or the response schema — it makes the existing guard robust to a case the prompt already implicitly permits (the model is looking at one continuous idx list; nothing stops it from applying "duplicate" semantics to an item that happens to sit in the second half). Happy to open a PR for this if useful — we hit it via our own downstream reasoning-first `EdgeDuplicate` patch (see #1666) but confirmed it reproduces against the stock schema too.

vvivekranjan:
## Summary
This PR fixes a bug in `resolve_extracted_edge` where valid duplicate edge citations identified by the LLM within `existing_edges` (invalidation candidates) were incorrectly discarded by the duplicate-range guard. The duplicate index validation has been expanded to support the entire continuous index space (up to `max_valid_idx`), allowing cross-list duplicates to be correctly mapped, resolved, attributed, and saved rather than leaked as new duplicate edges.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
*N/A - Bug fix*

## Testing
- [x] All existing tests pass

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1672

### DIFF
--- a/graphiti_core/utils/maintenance/edge_operations.py
+++ b/graphiti_core/utils/maintenance/edge_operations.py
@@ -732,32 +732,40 @@ async def resolve_extracted_edge(
     response_object = EdgeDuplicate(**llm_response)
     duplicate_facts = response_object.duplicate_facts
 
-    # Validate duplicate_facts are in valid range for EXISTING FACTS
-    invalid_duplicates = [i for i in duplicate_facts if i < 0 or i >= len(related_edges)]
-    if invalid_duplicates:
-        logger.warning(
-            'LLM returned invalid duplicate_facts idx values %s (valid range: 0-%d for EXISTING FACTS)',
-            invalid_duplicates,
-            len(related_edges) - 1,
-        )
-
-    duplicate_fact_ids: list[int] = [i for i in duplicate_facts if 0 <= i < len(related_edges)]
-
-    resolved_edge = extracted_edge
-    for duplicate_fact_id in duplicate_fact_ids:
-        resolved_edge = related_edges[duplicate_fact_id]
-        break
-
-    if duplicate_fact_ids and episode is not None:
-        resolved_edge.episodes.append(episode.uuid)
-
     # Process contradicted facts (continuous indexing across both lists)
     contradicted_facts: list[int] = response_object.contradicted_facts
     invalidation_candidates: list[EntityEdge] = []
 
     # Only process contradictions if there are edges to check against
+    max_valid_idx = len(related_edges) + len(existing_edges) - 1
+
+    # Validate duplicate_facts are in valid range across both lists
+    duplicate_fact_ids = [i for i in duplicate_facts if 0 <= i < len(related_edges)]
+    cross_list_duplicates = [i for i in duplicate_facts if len(related_edges) <= i <= max_valid_idx]
+    truly_invalid_duplicates = [i for i in duplicate_facts if i < 0 or i > max_valid_idx]
+
+    if truly_invalid_duplicates:
+        logger.warning(
+            'LLM returned invalid duplicate_facts idx values %s (valid range: 0-%d)',
+            truly_invalid_duplicates,
+            max_valid_idx,
+        )
+    
+    resolved_edge = extracted_edge
+    for duplicate_facts_id in duplicate_fact_ids:
+        resolved_edge = related_edges[duplicate_facts_id]
+        break
+
+    if not duplicate_fact_ids and cross_list_duplicates:
+        # Duplicate cited only against an invalidation candidate — merge into it
+        # directly instead of discarding a correct verdict as "invalid."
+        resolved_edge = existing_edges[cross_list_duplicates[0] - len(related_edges)]
+    
+    if (duplicate_fact_ids or cross_list_duplicates) and episode is not None:
+        if episode.uuid not in resolved_edge.episodes:
+            resolved_edge.episodes.append(episode.uuid)
+
     if related_edges or existing_edges:
-        max_valid_idx = len(related_edges) + len(existing_edges) - 1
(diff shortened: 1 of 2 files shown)


vvivekranjan:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: vvivekranjan20@gmail.com

mehulp93:
Thanks for the contribution and for digging into #1672. 

We're going to hold off on this approach, though. `related_edges` (merge pool) and `existing_edges` (invalidation pool) are intentionally separate: the merge pool is pair-scoped so duplicates only attach to edges with the same endpoints, while the invalidation pool is a broader similarity search that is only safe for contradiction handling. Accepting a cross-list `duplicate_facts` citation and merging into an invalidation candidate collapses that boundary. That can attribute a fact to the wrong entity pair when the candidate shares similar text but different node UUIDs (the added test even covers a different `target_node_uuid`).

The range guard discarding out-of-merge-pool idxs is enforcing that invariant, not an accidental bug. If the merge pool is empty because of race/visibility or an upstream entity-dedupe miss, the right fix is to address that path — not to treat invalidation candidates as merge targets.

I'm going to close this PR, but thanks for your contribution.

Issue #1651, opened 2026-07-15
MCP server: get_episodes and delete_episode query the wrong graph on FalkorDB (bypass per-group routing)


kramerica-inc-dev:
## Bug description

On FalkorDB every `group_id` is a separate graph (multi-tenant), but the MCP server's `get_episodes` and `delete_episode` tools query whatever graph the shared driver happens to be bound to — `default_db` at startup, or the graph of the **last-written group** after any `add_episode` (which rebinds the driver via `self.driver = self.driver.clone(database=group_id)`). Neither tool routes to the graph of the requested group, so on FalkorDB they operate on the wrong graph and report data as missing that demonstrably exists.

This is distinct from #1161 / #1325 (single-group routing in `handle_multiple_group_ids`) and is **not** fixed by #1572: these MCP handlers bypass the decorator entirely —

- `get_episodes` calls `EpisodicNode.get_by_group_ids(client.driver, ...)` directly (`mcp_server/src/graphiti_mcp_server.py`), and
- `delete_episode` calls `client.remove_episode(uuid)`, which does `EpisodicNode.get_by_uuid(self.driver, ...)` (`graphiti_core/graphiti.py`) —

both against the driver's current database.

## Steps to reproduce

Observed on graphiti-core 0.28.2 (standalone MCP image) against FalkorDB; code inspection shows the same flow on current `main`.

1. `add_memory(name="x", episode_body="...", group_id="groupA")` → episode processes and lands; the Episodic node is visible in the `groupA` graph:
   ```
   GRAPH.RO_QUERY groupA "MATCH (e:Episodic) WHERE e.name = 'x' RETURN e.uuid"   → 1 row
   ```
2. `get_episodes(group_ids=["groupA"])` → `"No episodes found"` (the query ran against `default_db` — or against the last-written group's graph — filtered by `group_id`, so it matches nothing).
3. `delete_episode(uuid=<uuid from step 1>)` → `"Error deleting episode: node <uuid> not found"` for the same reason.

## Expected behavior

`get_episodes` returns the episodes of the requested groups; `delete_episode` finds and removes the episode regardless of which graph the shared driver was last bound to.

## Suggested fix directions

- In the MCP handlers: clone the driver per requested `group_id` (as `add_episode` does) and query each group's graph, instead of calling `EpisodicNode.get_by_group_ids` on the pinned driver; or route through the decorated search/retrieval methods once #1572 lands (that PR fixes the decorator, but these handlers currently bypass it).
[...]

Issue #1325, opened 2026-03-15
FalkorDB: handle_multiple_group_ids skips graph routing for single group_id


spencer2211:
## Bug description

The `handle_multiple_group_ids` decorator in `graphiti_core/decorators.py` only switches the FalkorDB driver to the correct graph database when `len(group_ids) > 1`. For a single `group_id` (the most common case), the decorator falls through to the normal execution path, which queries `default_db` — a graph that typically has no data.

This means any FalkorDB deployment using a single `group_id` per query (the standard pattern) silently returns empty results from `search`, `search_`, `retrieve_episodes`, and `build_communities`.

## Root cause

In `decorators.py`, the condition:

```python
and len(group_ids) > 1
```

should be:

```python
and len(group_ids) >= 1
```

FalkorDB uses separate graph databases per `group_id` (unlike Neo4j, which uses `group_id` as a node property filter within a single database). The decorator needs to switch the driver's target graph for *any* number of group_ids, not just multiple.

## Additional issue: `get_by_group_ids` classmethods bypass the decorator

The `get_by_group_ids` classmethods on `EpisodicNode`, `EntityNode`, `CommunityNode`, and `SagaNode` in `nodes.py` (and similarly `EpisodicEdge`, `EntityEdge`, `CommunityEdge`, `HasEpisodeEdge`, `NextEpisodeEdge` in `edges.py`) query the driver directly without going through the `handle_multiple_group_ids` decorator. For FalkorDB, these queries always hit whatever graph the driver is currently pointed at (typically `default_db`), rather than routing to the per-group graph.

These classmethods need FalkorDB-specific handling to clone the driver with the correct database for each group_id.

## Reproduction

1. Set up graphiti-core with a FalkorDB backend
2. Add episodes with `group_id="my-group"`
3. Call `graphiti.search(group_ids=["my-group"], query="...")`
4. Result: empty — the query hits `default_db` instead of the `my-group` graph

## Proposed fix

1. Change `len(group_ids) > 1` → `len(group_ids) >= 1` in `decorators.py`
2. Add FalkorDB graph routing to `get_by_group_ids` classmethods in `nodes.py` and `edges.py`

PR: #1326

765007025:
[...]

PR #1675, opened 2026-07-23
fix(falkordb): route namespace/ops get_by_group_ids reads to the group's graph

spencer2211:
## Summary

`#1670` fixed single-`group_id` routing for the **decorated** path (`@handle_multiple_group_ids`, which clones `self.clients.driver` and passes it down as `driver=`). The **namespace/ops read path** is a separate implementation and is still unrouted on `main`.

`graphiti.nodes.episode.get_by_group_ids(...)` delegates to the FalkorDB `*Operations` classes (`FalkorEpisodeNodeOperations.get_by_group_ids` etc.). Those methods receive the base driver as `executor` — pointed at `default_db` — and run the query directly, so they never go through the decorator's clone. On FalkorDB, where each `group_id` is its own graph, a single-group read therefore queries the empty default graph and returns 0 rows.

This is the path reported in #1651 (`get_episodes` / `delete_episode` querying the wrong graph from the MCP server). There are currently no `clone` call sites anywhere under `graphiti_core/driver/falkordb/operations/`.

Related history: #1325 is the original report; #1326 covered both the classmethod path and this ops path, and was closed in favour of #1670, which covered only the former. This PR carries the remaining half forward against current `main`.

## Change

Apply single-group `executor.clone(database=gid)` routing plus multi-group fan-out to all nine FalkorDB operations `get_by_group_ids` methods:

- **nodes:** episode, entity, community, saga
- **edges:** episodic, entity, community, has_episode, next_episode

The clone is guarded by `isinstance(executor, GraphDriver)` so the `QueryExecutor` parameter narrows to a clone-capable driver (the `QueryExecutor` protocol does not declare `clone`/`provider`), keeping pyright clean.

This is complementary to #1670, not overlapping: the decorator routes at the search/retrieve entry point via `driver=`, this routes at the ops layer via `executor=`.

## Test plan

Adds `tests/driver/test_falkordb_ops_routing.py` — a package-free regression using a `spec=GraphDriver` mock, so it needs no live FalkorDB and runs in CI. Covers single-group routing, multi-group fan-out, and the empty-`group_ids` pass-through.

- [x] `uv run pytest tests/driver/test_falkordb_ops_routing.py -q` → 5 passed
- [x] `uv run pytest tests/driver/ tests/test_handle_multiple_group_ids.py -q` → 34 passed, 1 skipped (includes #1670's decorator tests — no interaction regression)
- [x] `make lint` → ruff clean, pyright 0 errors

Happy to consolidate the repeated fan-out into a shared helper rather than one block per ops class if you'd prefer that shape.


spencer2211:

spencer2211:
@Naseem77 this is the namespace-path gap you flagged on #1326 — carried forward onto post-#1670 `main` as a standalone change, since #1670 covered the decorated path only. Same nine ops methods and the same package-free regression test you reviewed there.

cc @prasmussen15 for context: #1326 was closed in favour of #1670, which fixed `@handle_multiple_group_ids` but not the `*Operations.get_by_group_ids` read path (#1651).

spencer2211:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: spencer.p.fuller@gmail.com

spencer2211:
recheck

prasmussen15 (approved):

PR #1680, opened 2026-07-24
avoid mutating shared search config

adolphinvx:
## Summary

- copy the shared edge search recipe before applying the per-request result limit
- prevent Graphiti.search() calls from mutating module-level search configuration
- add regression coverage for searches with and without a center node

## Testing

- uv run pytest tests/test_graphiti_search.py -q
- make lint

## API / schema impact

None.

Supersedes #1655 after renaming the source branch.


adolphinvx:
@prasmussen15 Could you please help manually triage or review this when convenient?  The remaining triage failure is workflow-side—the action rejects external contributors with read-only access. CLA, check-fork, Ruff, local Pyright, and the focused tests pass.

Issue #1342, opened 2026-03-22
[BUG]  `episode_mentions_reranker` sorts nodes ascending (fewest mentions first)


jake158:
## Bug Description
`episode_mentions_reranker` in `graphiti_core/search/search_utils.py` sorts nodes in ascending order of mention count, so nodes mentioned in the fewest episodes rank highest. This is the opposite of what the reranker name implies and what the equivalent edge reranker does.

## Steps to Reproduce
Provide a minimal code example that reproduces the issue:

```python
from collections import defaultdict                                                        

def rrf(results, rank_const=1, min_score=0):
    scores = defaultdict(float)
    for result in results:
        for i, uuid in enumerate(result):
            scores[uuid] += 1 / (i + rank_const)                                                        
    sorted_uuids = sorted(scores.keys(), key=lambda u: scores[u], reverse=True)
    return (                                                                                            
        [u for u in sorted_uuids if scores[u] >= min_score],                                          
        [scores[u] for u in sorted_uuids if scores[u] >= min_score],                                    
    )                                                                                                 

# Simulate episode_mentions_reranker with mocked DB counts                                              
def simulate_reranker(node_uuid_lists, db_mention_counts):
    sorted_uuids, _ = rrf(node_uuid_lists)                                                              
    scores = {}                                                                                         
    for uuid in sorted_uuids:                                                                           
        if uuid in db_mention_counts:                                                                   
            scores[uuid] = db_mention_counts[uuid]                                                    
    for uuid in sorted_uuids:                                                                           
        if uuid not in scores:                                                                          
            scores[uuid] = float('inf')                                                                 
    sorted_uuids.sort(key=lambda u: scores[u])  # line 1894 — ascending                                 
    return [(u, scores[u]) for u in sorted_uuids]                                                       
                                                                                                          
result = simulate_reranker(                                                                             
    [["node-A", "node-B", "node-C"]],                                                                   
    {"node-A": 20, "node-B": 5, "node-C": 1},                                                           
)                                                                                                       
for rank, (uuid, score) in enumerate(result, 1):                                                        
    print(f"rank {rank}: {uuid}  mentions={score}")
```

## Expected Behavior
```
rank 1: node-A  mentions=20                                                                           
rank 2: node-B  mentions=5                                                                              
rank 3: node-C  mentions=1
```

## Actual Behavior
```
rank 1: node-C  mentions=1                                                                           
rank 2: node-B  mentions=5                                                                           
rank 3: node-A  mentions=20
[...]

PR #1687, opened 2026-07-26
fix(falkordb): episode_mentions_reranker inverted ordering and min_score bypass

Naseem77:
## What it fixes

Two related bugs in `FalkorSearchOperations.episode_mentions_reranker` (`graphiti_core/driver/falkordb/operations/search_ops.py`):

**1. Ranking is inverted**

Scores are episode-mention counts (higher = more relevant), but the sort was ascending, so the *least*-mentioned nodes ranked first. A node mentioned by 3 episodes ranked below a node mentioned by 1.

**2. Unmentioned nodes bypass `min_score`**

Nodes with no `MENTIONS` edges were given `float('inf')` as a score, which passes any `min_score` threshold (and, combined with the ascending sort, also pushed them to the end rather than filtering them). They now score `0`, so any positive `min_score` filters them out.

Related: #1342 reported the inverted ordering for the legacy `search_utils.py` path (fix pending in #1640/#1512), but this ops-layer copy has the same bugs and is not covered by those PRs.

## Before / after

With `Popular` mentioned by 3 episodes, `Rare` by 1, `Unmentioned` by 0:

```python
await driver.search_ops.episode_mentions_reranker(driver, uuids)
# before: ['Rare', 'Popular', 'Unmentioned']
# after:  ['Popular', 'Rare', 'Unmentioned']

await driver.search_ops.episode_mentions_reranker(driver, uuids, min_score=2)
# before: ['Popular', 'Unmentioned']   (inf >= 2)
# after:  ['Popular']
```

## Validation

- New FalkorDB integration test covering both ordering and `min_score`; verified it fails on the unfixed code and passes with the fix
- Full `tests/driver` suite passes against a live FalkorDB

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -674,7 +674,7 @@ class FalkorSearchOperations(SearchOperations):
 
         for uuid in node_uuids:
             if uuid not in scores:
-                scores[uuid] = float('inf')
+                scores[uuid] = 0
 
         sorted_uuids = list(node_uuids)
         sorted_uuids.sort(reverse=True, key=lambda cur_uuid: scores[cur_uuid])
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -677,7 +677,7 @@ class FalkorSearchOperations(SearchOperations):
                 scores[uuid] = float('inf')
 
         sorted_uuids = list(node_uuids)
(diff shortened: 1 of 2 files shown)


Naseem77:
Closing this. The fix is right but it lands in `driver/falkordb/operations/search_ops.py`, which nothing reaches — `search_interface` is never assigned, so `search_utils.episode_mentions_reranker` always falls through to its own copy. My test passed only because it called `driver.search_ops...` directly instead of going through `Graphiti.search()`. #1720 applies the same fix on the live path in `search_utils.py`, which is the correct place.

PR #1688, opened 2026-07-26
fix: include Saga nodes when clearing data by group_ids

Naseem77:
## What it fixes

`clear_data(group_ids=[...])` silently leaks Saga nodes. The group-scoped deletion iterates a hardcoded label list (`Entity`, `Episodic`, `Community`) that was never updated when the `Saga` node type was added, so Saga nodes belonging to the cleared groups survive:

```python
# graph contains (:Saga {group_id: 'g1'}) and (:Entity {group_id: 'g1'})
await clear_data(driver, group_ids=['g1'])
# Entity deleted, Saga still present
```

This PR adds `Saga` to the label list in all places that share the omission:

- `graphiti_core/driver/falkordb/operations/graph_ops.py`
- `graphiti_core/driver/neo4j/operations/graph_ops.py`
- `graphiti_core/driver/neptune/operations/graph_ops.py`
- `graphiti_core/driver/kuzu/operations/graph_ops.py`
- `graphiti_core/utils/maintenance/graph_data_operations.py` (the default public path)

All backends define the Saga node type (`graph_queries.py` creates Saga indexes for each), so the label is valid everywhere. `clear_data()` without `group_ids` is unaffected (it already deletes everything).

## Validation

- New FalkorDB integration test: a Saga node in the cleared group is deleted while a Saga node in another group survives; verified the test fails on the unfixed code and passes with the fix
- `tests/driver` and `tests/utils/maintenance` suites pass against a live FalkorDB


prasmussen15 (approved):

PR #1689, opened 2026-07-26
fix(falkordb): stop persisting labels list as a property in entity bulk save

Naseem77:
## What it fixes

The FALKORDB case of `get_entity_node_save_bulk_query` (`graphiti_core/models/nodes/node_db_queries.py`) has two problems:

**1. The `labels` list is persisted as a node property**

The query runs `SET n = node` with the raw node dict, which still contains the `labels` key, so every bulk-saved entity gets a stray `labels` list property:

```
MATCH (n:Entity {uuid: 'bulk-1'}) RETURN labels(n), n.labels
# before: ['Entity', 'Person'], ['Entity', 'Person']   <- stray property
# after:  ['Entity', 'Person'], null
```

The single-node save path does not have this problem, so bulk-saved and individually-saved entities end up with inconsistent properties. (Neptune's case already strips the key via `removeKeyFromMap`.)

**2. One query per label instead of one per node**

A node with N labels generated N separate `MERGE ... SET` queries. Labels are now applied in a single query via `SET n:Label1:Label2`.

## Validation

- New FalkorDB integration test asserting bulk-saved nodes get real labels, no `labels` property, and one query per node; verified it fails on the unfixed code and passes with the fix
- Live end-to-end check through `entity_node_ops.save_bulk` confirms `labels(n) == ['Entity', 'Person']` and `n.labels IS NULL`
- `tests/driver` and `tests/utils` suites pass against a live FalkorDB


gkorland:
Tagging @galshubeli for FalkorDB-team review.

gkorland:

prasmussen15 (approved):

prasmussen15:
Landed onto current main (including #1688) as https://github.com/getzep/graphiti/pull/1852 after resolving the test-file conflict. Original PR cannot be pushed from this maintainer token.

prasmussen15:
The original fork still cannot merge into current `main` (test-file conflict with #1688). Combined landing PR: https://github.com/getzep/graphiti/pull/1856

prasmussen15:
Landed on \`main\` via https://github.com/getzep/graphiti/pull/1856 (\`3ff5c16\`). You are credited as a co-author on that commit. Closing this fork PR to avoid a duplicate merge.

PR #1690, opened 2026-07-26
fix(falkordb): stop leaking source_uuid/target_uuid into edge attributes

Naseem77:
## What it fixes

`EntityEdge.save` on FalkorDB pollutes every edge with stray `source_uuid` / `target_uuid` properties that then leak into `EntityEdge.attributes` on every read.

The FALKORDB save query MATCHes the endpoints via `$edge_data.source_uuid` / `$edge_data.target_uuid` and then runs `SET e = $edge_data`, storing both MATCH keys as real edge properties. The record parsers strip `source_node_uuid` / `target_node_uuid` from attributes but not these keys, so:

```python
edge = EntityEdge(source_node_uuid=a.uuid, target_node_uuid=b.uuid, ...)
await edge.save(driver)
loaded = await EntityEdge.get_by_uuid(driver, edge.uuid)
loaded.attributes
# {'source_uuid': '637dbdbe-...', 'target_uuid': '6409d11f-...'}  <- should be {}
```

Since attributes are merged back into `edge_data` on save, the stray keys are re-persisted forever. The bulk save path uses `source_node_uuid` names and is unaffected.

## Changes

- **Write side:** the FALKORDB save query now takes `$source_uuid` / `$target_uuid` as separate parameters, keeping them out of the map assigned to the edge; both call sites (`falkordb/operations/entity_edge_ops.py` and the legacy `EntityEdge.save` in `edges.py`) pass them separately
- **Read side:** both edge record parsers additionally pop `source_uuid` / `target_uuid`, so edges already polluted by earlier versions are cleaned on read instead of leaking

## Validation

- New FalkorDB integration test asserting no stray properties on the stored edge and `attributes == {}` after a round trip; verified it fails on the unfixed code and passes with the fix
- Live check through the legacy `EntityEdge.save` path: custom attributes are preserved (`{'custom': 'kept'}`) with no leaked keys
- `tests/driver` and `tests/test_edge_int.py` pass against a live FalkorDB


linhongyu510 (changes requested):
The separate MATCH parameters fix the normal empty/custom-attribute path, and both record parsers correctly remove legacy keys. However, the operations save path still runs `edge_data.update(edge.attributes or {})` after constructing the payload and never removes `source_uuid` / `target_uuid`. A caller-created `EntityEdge(attributes={"source_uuid": "old-s", "target_uuid": "old-t"})` therefore persists those exact stray properties through `FalkorEntityEdgeOperations.save`, so the PR’s no-leak invariant is not complete. I reproduced this at `32a15c7a` with an `AsyncMock` executor: the trusted separate MATCH args are `s`/`t`, but `edge_data` still contains `old-s`/`old-t`. The integration test uses empty attributes and misses this case. Please remove both reserved legacy keys from `edge_data` after merging attributes (or filter them while merging), and add a no-database regression with pre-polluted attributes. Also consider filtering the canonical endpoint keys in the bulk path, where `edge_data.update(...)` can currently overwrite trusted `source_node_uuid` / `target_node_uuid`. Validation performed: targeted Ruff check/format and `git diff --check` pass; the live FalkorDB test skips locally because the service is unavailable.

gkorland:
Tagging @galshubeli for FalkorDB-team review.

gkorland:

prasmussen15:
Not merging yet: CLA is unsigned, and @linhongyu510 requested changes — `FalkorEntityEdgeOperations.save` still persists `source_uuid`/`target_uuid` from pre-polluted `edge.attributes`. Please sign the CLA and strip those reserved keys after merging attributes (plus a no-DB regression).

prasmussen15 (changes requested):
This is not ready to merge.

1. CLA is still unsigned — the CLA Assistant check is failing.
2. The operations save path still leaks reserved keys. After `edge_data.update(edge.attributes or {})`, `source_uuid` / `target_uuid` from a caller-supplied `attributes` dict are persisted through `FalkorEntityEdgeOperations.save`. Strip those reserved legacy keys after merging attributes (and filter canonical endpoint keys in the bulk path so `update(...)` cannot overwrite trusted `source_node_uuid` / `target_node_uuid`).
3. Add a no-database regression that starts from pre-polluted `attributes={"source_uuid": ..., "target_uuid": ...}` and asserts those keys never land in the saved payload.

Please address the above before we merge.

gkorland:

PR #1691, opened 2026-07-27
fix(falkordb): isolate per-group drivers

### LINKED ISSUE #1676: [BUG] Concurrent multi-group_id episode processing silently corrupts data across FalkorDB graphs (shared self.driver mutation race)

jatinder14:
## Summary
- keep FalkorDB database clones scoped to a copied `Graphiti` instance
- prevent concurrent `add_episode` and `add_episode_bulk` calls from rebinding shared driver state
- add a regression test proving each group retains its own driver while the shared client remains unchanged

Fixes #1676

## Test plan
- [x] `uv run pytest tests/test_handle_multiple_group_ids.py`
- [x] `uv run ruff check graphiti_core/graphiti.py tests/test_handle_multiple_group_ids.py`

### DIFF
--- a/graphiti_core/driver/driver.py
+++ b/graphiti_core/driver/driver.py
@@ -128,6 +128,14 @@ class GraphDriver(QueryExecutor, ABC):
     async def build_indices_and_constraints(self, delete_existing: bool = False):
         raise NotImplementedError()
 
+    async def wait_for_initialization(self) -> None:
+        """Wait for any index/constraint build scheduled when the driver was created.
+
+        Drivers that bootstrap themselves in the background override this so
+        callers can surface failures instead of losing them on an orphaned task.
+        """
+        return None
+
     def clone(self, database: str) -> GraphDriver:
         """Clone the driver with a different database or graph name."""
         return self
--- a/graphiti_core/driver/falkordb_driver.py
+++ b/graphiti_core/driver/falkordb_driver.py
@@ -180,6 +180,7 @@ class FalkorDriver(GraphDriver):
         try:
             loop = asyncio.get_running_loop()
             self._init_task = loop.create_task(self.build_indices_and_constraints())
+            self._init_task.add_done_callback(self._log_init_task_result)
         except RuntimeError:
             pass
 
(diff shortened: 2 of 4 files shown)


jatinder14:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: jatinder14122000@gmail.com

jatinder14:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: jatinder14122000@gmail.com

Naseem77:
@jatinder14 I tested it live before and after: adding episodes for different groups at the same time no longer makes them land in the wrong graph, and the regression test fails without the patch and passes with it. The fix holds, LGTM.

jatinder14:
Thanks @Naseem77 for the live before/after check and confirming the regression test fails without the patch. Really appreciate you validating the group isolation fix.

Naseem77:
@jatinder14 Following up on my earlier review, the routing fix still holds live but each cloned group driver spawns an unawaited `build_indices_and_constraints()` that dies with `ConnectionError` and logs "Task exception was never retrieved", so index creation silently fails per group. #1699 avoids it and #1765 deliberately works around the same `clone()` behaviour. Worth awaiting or guarding that task.

jatinder14:
Good catch @Naseem77 — that was a real hole. Fixed in `af69303`.

You're right that the routing fix alone left every cloned driver spawning its own fire-and-forget `build_indices_and_constraints()`. `FalkorDriver.__init__` schedules it with `loop.create_task(...)` and nothing ever held the handle, so a failing build surfaced only as "Task exception was never retrieved" and that group's indices were quietly missing.

Two changes:

- `GraphDriver` gains a no-op `wait_for_initialization()`, overridden in `FalkorDriver` to await the task scheduled at construction. `_with_database` is now async and awaits it before handing the scoped instance back, so a group's first write no longer races its own index build.
- `__init__` also attaches a done callback that logs the outcome, so even on paths that never await, the exception is retrieved and you get a real `logger.warning` naming the graph instead of an orphaned-task traceback.

The regression test now asserts each clone's build is awaited exactly once and the shared driver's is not touched. Also rebased onto current `main`.

Re #1699 and #1765 working around the same `clone()` behaviour — happy to follow up on consolidating that if the maintainers would rather the guard live in `clone()` itself than at the call site.

PR #1698, opened 2026-07-28
feat(mcp): allow explicit reranker provider selection

### LINKED ISSUE #1697: RFC: allow explicit MCP reranker provider configuration
## Problem

After #1637, the MCP server automatically selects a reranker from the configured LLM provider, then the embedder provider, and finally local BGE. That fixes the hard OpenAI dependency, but operators cannot explicitly choose a different available reranker or force the local option.

This matters when, for example, an OpenAI LLM deployment wants to use an already-configured Gemini reranker, or an operator wants deterministic local BGE selection instead of provider inference.

## Proposed configuration

Add a small top-level configuration block:

```yaml
reranker:
  provider: auto  # auto, openai, azure_openai, gemini, or bge
```

- `auto` preserves the behavior merged in #1637.
- API-backed selections reuse credentials already present under the matching `llm.providers` or `embedder.providers` entry.
- `bge` selects the existing local `BGERerankerClient`.
- No new provider integration or dependency is introduced.

The setting would also be available as `RERANKER_PROVIDER` through the shipped YAML and `RERANKER__PROVIDER` through direct nested environment configuration.

## Scope

The follow-up would be limited to the MCP configuration schema, factory selection, service wiring, tests, and MCP documentation. It builds on #1637 and is related to the earlier provider-matched proposal in #1669.

abouchard11:
## Summary

Add an explicit MCP `reranker.provider` configuration while preserving the automatic selection behavior merged in #1637.

- `auto` keeps the current LLM → embedder → local BGE fallback chain.
- `openai`, `azure_openai`, and `gemini` reuse credentials already configured under the matching LLM/embedder provider entry.
- `bge` explicitly selects the existing local reranker.
- Document both YAML and direct environment overrides.

This is the explicit configuration follow-up discussed around #1637 and the earlier provider-matched proposal #1669.

Closes #1697.

## Testing

- Test-first coverage for configuration parsing, explicit Gemini override behavior, and Graphiti service wiring.
- `68 passed` across MCP factory, configuration, core-parity, transport, and stdio unit tests.
- `ruff check` and `ruff format --check` pass for all changed Python files.
- Focused Pyright passes with 0 errors, 0 warnings.

### DIFF
--- a/mcp_server/README.md
+++ b/mcp_server/README.md
@@ -169,9 +169,20 @@ llm:
   # Optional for OpenAI-compatible providers: "json_schema" (default) or "json_object"
   structured_output_mode: "json_schema"
 
+reranker:
+  # "auto" infers from the LLM, then embedder, then falls back to local BGE.
+  # Or select "openai", "azure_openai", "gemini", or "bge" explicitly.
+  provider: "auto"
+
 database:
   provider: "falkordb"  # Default. Options: "falkordb", "neo4j"
 ```
+
+An explicit API-backed reranker reuses credentials from the matching `llm.providers` or
+`embedder.providers` entry, even when that provider is not selected for the LLM or embedder. For
+example, `reranker.provider: "gemini"` uses the configured Gemini API key. Set
+`RERANKER_PROVIDER` when using the shipped `config.yaml`, or `RERANKER__PROVIDER` with direct
+environment-based configuration.
 
 ### Using Ollama for Local LLM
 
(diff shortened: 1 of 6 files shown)


abouchard11:
Hi maintainers—checking whether this explicit reranker configuration aligns with the direction you want for the MCP server. The PR preserves the existing auto behavior and adds explicit provider selection with focused tests. Happy to adjust the scope or implementation based on your feedback.

abouchard11:
Withdrawing while I review this — will follow up.

PR #1699, opened 2026-07-29
fix: use request-scoped driver for concurrent multi-group_id isolation (#1676)

### LINKED ISSUE #1676: [BUG] Concurrent multi-group_id episode processing silently corrupts data across FalkorDB graphs (shared self.driver mutation race)

ozdemirsarman:
## Summary

Fixes #1676. Concurrent `add_episode` / `add_episode_bulk` calls for different `group_id`s could silently persist episodes under the wrong database because both methods mutate the shared `self.driver`.

Both methods contained:

```python
```

These coroutines have many `await` points (LLM extraction, embeddings, DB reads/writes). While call A for `group_id="A"` is suspended at an `await`, a concurrent call B for `group_id="B"` runs this block and reassigns `self.driver` to B's database. When A resumes, its remaining reads/writes go to B's graph. No error is raised — the corruption is silent. The reporter observed 19 misplaced episodes across 5 FalkorDB group graphs in production.

## Fix

Replace the shared-state mutation with a **request-scoped** driver/clients bundle. A new helper resolves the effective `group_id` plus a per-call driver and a matching `GraphitiClients` copy:

```python
def _resolve_request_scope(self, group_id):
    if group_id is None:
        return get_default_group_id(self.driver.provider), self.driver, self.clients
    validate_group_id(group_id)
    if group_id == self.driver._database:
        return group_id, self.driver, self.clients
    clients = self.clients.model_copy(update={'driver': driver})
    return group_id, driver, clients
```

`add_episode`, `add_episode_bulk`, and the helper methods they call — `_extract_and_resolve_edges`, `_process_episode_data`, `_extract_and_dedupe_nodes_bulk`, `_resolve_nodes_and_edges_bulk`, `_get_or_create_saga`, `_saga_get_previous_episode_uuid` — now thread this `driver` / `clients` through instead of reading `self.driver` / `self.clients`. Each helper accepts an optional bundle that defaults to `self.clients` / `self.driver`, so their signatures stay backward compatible.

`self.driver` is never reassigned, so concurrent calls no longer interfere.

## Why this is safe / low-risk

- **Common case unchanged:** when `group_id` is `None` or equal to the driver's current database (the typical single-database deployment), the shared `self.driver` and `self.clients` are returned as-is — no clone, no copy, no behaviour or performance change.
- **FalkorDB single/multi `group_id` reads** already route correctly via the existing `@handle_multiple_group_ids` decorator, which clones a call-scoped driver per `group_id`. This fix removes the write-path mutation that the decorator's comment (see #1659) was already working around, making the two consistent.
- Helper method signatures only gain an optional trailing `clients` / `driver` parameter.

## Behaviour change to note

`self.driver` is no longer mutated as a side effect of `add_episode`. Any caller that relied on that side effect to implicitly route later `search` / `retrieve_episodes` calls should pass `group_ids` explicitly — which is already the supported API and is handled by `@handle_multiple_group_ids` for FalkorDB. This reliance was exactly the unsafe behaviour that caused the bug.

## Tests

Adds `tests/test_request_scope_concurrency.py` — database-free regression tests using a fake driver:

- `_resolve_request_scope` reuses shared instances for `None` / matching `group_id` (no clone).
- For a differing `group_id`, it returns a clone targeting that database and a clients copy that preserves the other clients — while leaving `self.driver` / `self.clients` untouched.
- A concurrency test interleaves three workers across an `await` boundary and asserts each keeps its own driver/database and that the shared driver is never repointed. This fails against the pre-fix mutation and passes with the fix.

## Verification performed

- `ruff format` + `ruff check` clean on changed files.
- `pyright` clean on changed files.
- New regression tests pass (4/4).
- Existing `tests/test_handle_multiple_group_ids.py` passes (4/4).

> Note: the DB-backed integration suites require a live Neo4j/FalkorDB/Kuzu and were not run locally. Worth running in CI against FalkorDB, since that is where the corruption was observed.


ozdemirsarman:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: ozdemir.sarman@ozdemirsarman.com

Naseem77:
@ozdemirsarman Reproduced live on FalkorDB: two concurrent `add_episode` calls for different group_ids put grpA's episode in grpB's graph (grpA=0, grpB=2), plus `self.driver` stayed mutated. Your patch gives 1/1 with no leak, and short-circuiting the default-database case keeps single-DB setups untouched. The tests call `_resolve_request_scope` directly so they fail on main with `AttributeError` instead of showing the bug, an end-to-end concurrent `add_episode` would prove it. Overlaps #1691/#1692.

prasmussen15 (approved):

PR #1700, opened 2026-07-29
Let .env reach the containers in docker compose

Ho1yShif:
Found while running the template quality bar on this repo.

## The problem

Both app services in `docker-compose.yml` listed their `environment:` explicitly, so only the vars named there were passed into the container. Everything else a user put in `.env` was silently dropped:

- `MODEL_NAME`
- `SEMAPHORE_LIMIT`
- `EMBEDDING_MODEL_NAME`
- `OPENAI_BASE_URL`
- `FALKORDB_USERNAME` / `FALKORDB_PASSWORD`

That contradicts three things we already tell people: the README says to copy `.env.example` to `.env` and that the falkordb profile "mirrors this Blueprint", and `.env.example` documents every var the app reads. In practice a local run used graphiti-core's default model while Render used `gpt-5.5`.

Confirmed before the change, on a clean clone:

```
$ docker compose --profile falkordb exec -T graph-falkordb printenv MODEL_NAME SEMAPHORE_LIMIT
(nothing)
```

## The fix

`env_file:` with `required: false`, rather than naming each var in `environment:`.

Naming them would hand a blank value to every var the user left out, and blank is not the same as unset — graphiti-core reads `int(os.getenv('SEMAPHORE_LIMIT', 20))`, so `''` is a crash on import. With `env_file`, absent from `.env` means absent from the environment. `.env.example` already leaves optional settings commented out for exactly this reason.

The compose-only wiring stays in `environment:`, which takes precedence over `env_file`, so the compose network hostnames still win over whatever `.env` says.

## Verification

On a fresh clone of `main`, from-scratch build:

- `MODEL_NAME=gpt-5.5` and `SEMAPHORE_LIMIT=10` now reach the container.
- Precedence holds: `FALKORDB_HOST` stays `falkordb` (not `.env`'s `localhost`), and on the default profile `DB_BACKEND` stays `neo4j` with `NEO4J_URI=bolt://neo4j:7687` despite `.env` saying `falkordb`.
- falkordb profile boots healthy, `/healthcheck` 200, no error-level logs, and a write-then-search round trip returns the temporal facts.
- With `.env` deleted entirely, the API still fails fast: `ValidationError ... openai_api_key: String should have at least 1 character`, container exits 3. No silent misconfiguration.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+# The root Dockerfile builds only server/, so exclude the whole repo and add that one
+# subtree back. An allow-list keeps the context minimal without needing an update every
+# time the repo grows a new top-level directory.
+*
+!server/
+
+# Re-exclude what lives under server/ but never belongs in an image layer: a local
+# virtualenv (larger than the image itself), local secrets, and build droppings.
+server/.venv
+server/.env
+server/.env.*
+server/tests
+server/**/__pycache__
+server/**/*.pyc
--- a/.env.example
+++ b/.env.example
@@ -38,8 +38,3 @@ SEMAPHORE_LIMIT=10
 # EMBEDDING_MODEL_NAME=
 # Point at an OpenAI-compatible gateway instead of api.openai.com.
 # OPENAI_BASE_URL=
-
-# --- Pinning ----------------------------------------------------------------------
-# The graphiti-core release docker-compose builds against, matching render.yaml. This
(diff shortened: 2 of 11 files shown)


Ho1yShif:
Opened by mistake against upstream — this was meant for my own fork (a Render deploy-template change, not an upstream fix). Closing. Apologies for the noise.

PR #1703, opened 2026-07-30
fix(falkordb): route grouped clears to tenant graphs

d3f4w2:
## Summary
Fix FalkorDB group-scoped clearing so MCP `clear_graph` deletes the requested tenant graphs instead of filtering the base/default graph and reporting a false success.

FalkorDB stores each `group_id` in a separate physical graph. `FalkorGraphMaintenanceOperations.clear_data` previously executed property-filtered deletes on the executor's current graph, which is normally `default_db`. The fix clones the driver per requested group, clears that physical graph, and preserves the existing transaction-like fallback for executors that cannot be cloned.

Duplicate group IDs are de-duplicated while preserving order, and groups are cleared sequentially to avoid concurrent operations on the shared FalkorDB connection.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
Restore the destructive-operation contract for FalkorDB multi-tenant deployments: a successful group-scoped clear must actually remove data from every requested group graph.

This is the remaining `clear_graph` portion of #1651 after #1670 and #1675 fixed the corresponding read paths. UUID-only lookup/delete tools are intentionally out of scope because their cross-graph lookup contract needs separate design.

## Testing
- [ ] All existing tests pass

Validated locally:
- `uv run pytest tests/driver/test_falkordb_ops_routing.py -q` — 8 passed
- relevant driver/decorator suite — 37 passed, 1 skipped
- `uv run ruff check` — passed
- `uv run pyright ./graphiti_core` — 0 errors
- `uv run ruff format --check` and `git diff --check` — passed

New unit coverage verifies single-group routing, multi-group routing, duplicate group IDs, and clearing the current graph when no group scope is supplied. The tests use `GraphDriver` mocks and do not require a live FalkorDB.

## Breaking Changes

## Checklist
- [x] Code follows project style guidelines (`make lint` equivalent commands pass)
- [x] Self-review completed

## Related Issues
Addresses the `clear_graph` path in #1651.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/graph_ops.py
+++ b/graphiti_core/driver/falkordb/operations/graph_ops.py
@@ -18,7 +18,7 @@ import asyncio
 import logging
 from typing import Any
 
-from graphiti_core.driver.driver import GraphProvider
+from graphiti_core.driver.driver import GraphDriver, GraphProvider
 from graphiti_core.driver.operations.graph_ops import GraphMaintenanceOperations
 from graphiti_core.driver.operations.graph_utils import Neighbor, label_propagation
 from graphiti_core.driver.query_executor import QueryExecutor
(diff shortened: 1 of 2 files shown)


d3f4w2:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: liujiahuishen@gmail.com

d3f4w2:

Naseem77:
@d3f4w2 The routing idea is right but the patch lands in `FalkorGraphMaintenanceOperations.clear_data`, which the pipeline never calls (`graph_operations_interface` is None for FalkorDB). Tested live: seeded grpX and grpY with 4 nodes each, `clear_data(group_ids=['grpX'])` left grpX at 4 both before and after your patch, and the method is invoked 0 times. The live path is `utils/maintenance/graph_data_operations.py`. Same trap as #1510/#1500/#1471.

PR #1704, opened 2026-07-30
Report OpenAI failures instead of a bare 500, and keep the ingest worker alive

Ho1yShif:
An exhausted OpenAI quota made this service look broken: `POST /search` answered a bare
`500 Internal Server Error`, and the ingestion worker died on the first failing job — silently,
since nothing awaits its task — so `POST /messages` kept answering 202 into a queue with no
reader. Both read as bugs in the graph query; the real cause was only in the logs, as a traceback.

## Changes

- `graph_service/openai_errors.py` maps an `openai.APIError` to a status and a message naming
  what to go and fix: 429 for a rate limit or exhausted quota, 502 when OpenAI rejected this
  service's credentials (not 401/403 — the caller's `GRAPHITI_API_KEY` was fine), 504 when OpenAI
  was never reached. Registered in `main.py` on the base class, so a subclass the SDK adds later
  is still mapped.
- `AsyncWorker.worker` drops a failing episode and keeps going instead of ending ingestion for the
  life of the process. No retry: `add_episode` isn't idempotent and a dead key would retry forever.
- `openai` is declared as a direct dependency, since `main.py` imports it.
- README: `"success": true` means queued, not ingested; and which endpoint fails first when
  `OPENAI_API_KEY` is unusable.

## Testing

`tests/test_upstream_failures.py` — needs no OpenAI key and no database, so it runs in the default
`make test`. Covers each mapped status, plus that the worker survives a failing job and logs it.

`make lint` clean, 33 tests passing.

## Flagged for a reviewer, not changed

- Responses append OpenAI's own error text (`Upstream said: ...`). Deliberate — the caller here is
  the operator who set the key, and the message is the only place the rate-limit/quota distinction
  survives — but it does pass upstream account state to any bearer-token holder.
- `openai.OpenAIError` siblings that aren't `APIError` (`LengthFinishReasonError`,
  `ContentFilterFinishReasonError`) still surface as 500s. They describe a response that arrived
  rather than an upstream that refused us, so neither gateway status fits.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,14 @@
+# The root Dockerfile builds only server/, so exclude the whole repo and add that one
+# subtree back. An allow-list keeps the context minimal without needing an update every
+# time the repo grows a new top-level directory.
+*
+!server/
+
+# Re-exclude what lives under server/ but never belongs in an image layer: a local
+# virtualenv (larger than the image itself), local secrets, and build droppings.
+server/.venv
+server/.env
+server/.env.*
+server/tests
+server/**/__pycache__
+server/**/*.pyc
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,11 +1,10 @@
-# The root Dockerfile builds only server/, so exclude the whole repo and add that one
-# subtree back. An allow-list keeps the context minimal without needing an update every
-# time the repo grows a new top-level directory.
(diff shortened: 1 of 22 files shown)


Ho1yShif:
Opened against the wrong repository by mistake — this belongs on our fork. Closing; sorry for the noise.

PR #1706, opened 2026-07-30
feat(mcp): allow deployments to override the server instructions

sgral:
## Summary

`GRAPHITI_MCP_INSTRUCTIONS` is a module-level constant passed straight into `FastMCP(...)`, so the only way to change what MCP clients are told on connect is to edit `graphiti_mcp_server.py`. For anyone running the server as a container that means maintaining a patched image purely to carry a few paragraphs of text.

That text is exactly where a deployment's own conventions belong — which `group_id`s exist, how to route facts between them, when to search before answering. The built-in default cannot know any of that, and it is the first thing every connecting client reads.

## Change

A small `resolve_server_instructions()` helper, checked in order:

1. `GRAPHITI_MCP_INSTRUCTIONS` — the instructions themselves
2. `GRAPHITI_MCP_INSTRUCTIONS_FILE` — a path to read them from
3. the built-in default (unchanged)

```python
mcp = FastMCP(
    'Graphiti Agent Memory',
    instructions=resolve_server_instructions(),
)
```

**Why environment and not `config.yaml`:** the `FastMCP` instance is constructed at import time, before the configuration file is loaded. `FastMCP.instructions` is a read-only property over `self._mcp_server.instructions`, so setting it later in `initialize_server()` would mean writing to a private attribute. Resolving before construction keeps this to public API only.

**Why a file path as well as an inline variable:** this field is normally multi-paragraph, which is awkward to carry in an environment variable and unpleasant to quote in compose files.

**Unreadable file falls back rather than raising** — it logs a warning and uses the default. Wrong instructions are a much smaller problem than a memory server that will not start.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [x] Documentation/Tests

## Objective

Let deployments describe their own memory conventions without forking the image. This came out of running the MCP server against FalkorDB with several active `group_id`s, where the routing rules an agent needs are entirely site-specific.

## Testing

- [x] Unit tests added — `mcp_server/tests/test_server_instructions.py`, five cases: default, inline override, file override, inline-beats-file precedence, and unreadable-file fallback.
- [x] Existing tests unaffected — the change is additive and the default path is byte-identical to before.

> Note: I verified the resolution order by exercising the function's logic directly (all five cases pass) but could not run the repository's `pytest` suite in my environment. Worth a CI run.

## Breaking Changes

With neither variable set, behaviour is exactly as before.

## Checklist

- [x] Self-review completed
- [x] Documentation updated where necessary — both variables added to the environment table in `mcp_server/README.md`

### DIFF
--- a/mcp_server/README.md
+++ b/mcp_server/README.md
@@ -254,6 +254,8 @@ The `config.yaml` file supports environment variable expansion using `${VAR_NAME
 - `AZURE_OPENAI_API_VERSION`: Optional Azure OpenAI API version
 - `USE_AZURE_AD`: Optional use Azure Managed Identities for authentication
 - `SEMAPHORE_LIMIT`: Episode processing concurrency. See [Concurrency and LLM Provider 429 Rate Limit Errors](#concurrency-and-llm-provider-429-rate-limit-errors)
+- `GRAPHITI_MCP_INSTRUCTIONS`: Optional replacement for the instructions sent to MCP clients on connect
+- `GRAPHITI_MCP_INSTRUCTIONS_FILE`: Optional path to read those instructions from, which is usually easier than a multi-paragraph environment variable. Ignored if `GRAPHITI_MCP_INSTRUCTIONS` is set; if the file cannot be read the built-in instructions are used and a warning is logged
 
 You can set these variables in a `.env` file in the project directory.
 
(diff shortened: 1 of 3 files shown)


sgral:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: sebastian.gral96@gmail.com

PR #1710, opened 2026-08-01
Fix infinite loop in label_propagation on disconnected graphs

KaloyanBorisov:
## Summary

`label_propagation` (the community-detection clustering algorithm used by `build_communities`) can loop forever on graphs with several small disconnected components containing degree-1 nodes. Its convergence loop built a snapshot of updated labels and swapped it in at the end of each pass; on the right graph shape, two nodes can end up trading community labels back and forth in lockstep every pass, so the `no_change` check never becomes `True`.

Since the loop is synchronous with no `await` inside it, this doesn't just hang the one request — it blocks the entire asyncio event loop, so a single unlucky `build_communities` call freezes the whole server for every other in-flight and incoming request too.

There are two copies of this same algorithm in the codebase (`graphiti_core/utils/maintenance/community_operations.py` and the shared `graphiti_core/driver/operations/graph_utils.py` used by the Neo4j/FalkorDB/Kuzu/Neptune drivers) — both had the identical bug and both are fixed here.

## Fix

- Apply label updates in place as they're computed (asynchronous propagation) instead of via a separate snapshot swapped in afterward. This breaks the lockstep symmetry that causes the oscillation.
- Add a hard iteration cap (`MAX_LABEL_PROPAGATION_ITERATIONS = 100`) as a backstop in case some other graph shape still doesn't converge, so the loop is now guaranteed to terminate regardless.

## How this was found / verified

Reproduced by building a small knowledge graph (via `add_memory`) that ended up with four disconnected clusters, then calling `build_communities` — it hung indefinitely at high CPU with no progress. Isolated the repro by pulling the exact neighbor projection from Neo4j and running the algorithm standalone: the original implementation did not converge after 1000 iterations. The fixed version converges immediately (~0.3s end-to-end through `get_community_clusters`) and produces the expected 4 clusters. Confirmed the fix through the actual `build_communities` MCP tool end-to-end afterward (4 communities built with real LLM summaries, no hang).

## Test plan
- [x] Existing community-building tests still pass
- [x] Manual repro: a group with several small disconnected clusters (e.g. multiple unrelated 2-3 node components) no longer hangs `build_communities`

### DIFF
--- a/graphiti_core/driver/operations/graph_utils.py
+++ b/graphiti_core/driver/operations/graph_utils.py
@@ -24,12 +24,21 @@ class Neighbor(BaseModel):
     edge_count: int
 
 
+MAX_LABEL_PROPAGATION_ITERATIONS = 100
+
+
 def label_propagation(projection: dict[str, list[Neighbor]]) -> list[list[str]]:
+    # Updates are applied in place (asynchronous propagation) rather than via a
+    # separate snapshot swapped in at the end of each pass. Batching all updates
+    # together lets two nodes trade labels back and forth in lockstep on some
+    # graph shapes (e.g. several small components with degree-1 nodes),
+    # oscillating forever instead of converging; applying updates as they're
+    # computed breaks that symmetry. The iteration cap is a backstop in case a
+    # graph shape still oscillates despite that.
     community_map = {uuid: i for i, uuid in enumerate(projection.keys())}
 
-    while True:
+    for _ in range(MAX_LABEL_PROPAGATION_ITERATIONS):
         no_change = True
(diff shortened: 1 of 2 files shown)


KaloyanBorisov:
I have read the CLA Document and I hereby sign the CLA behalf of myself, e-mail: kaloqn.borisov@gmail.com

Naseem77:
@KaloyanBorisov Confirmed the hang and confirmed your fix, all shapes converge correctly and patching the driver copy is right since that's the one backends call. Note #1719 fixes the same bug, also deletes the duplicated implementation so they can't drift again, and ships tests. Maintainers should probably pick one.

PR #1711, opened 2026-08-01
fix(falkordb): avoid a full :Entity scan per hit in edge_fulltext_search

trevor-sykes:
On FalkorDB, `edge_fulltext_search` resolves each matched relationship's endpoints with

```cypher
```

FalkorDB plans that as a `Node By Label Scan` over **every** `:Entity` node, once per row the
fulltext index yields, so the search costs `O(hits × entities)` even though the trailing
`LIMIT $limit` keeps only `RELEVANT_SCHEMA_LIMIT` rows:

```
Results
  Project
    Filter
      Edge By Index Scan | [e:RELATES_TO]
        Node By Label Scan | (n:Entity)      <-- once per fulltext hit
          ProcedureCall
```

The relationship the index already returned knows its own endpoints, so this reads them with
`startNode()`/`endNode()` instead. The added label predicate keeps the result set identical to
the pattern's. Upstream already uses this same idiom in the Neptune branch a few lines below.

### Measurements

Graph: 5,683 nodes / 20,374 `RELATES_TO` edges / 3,725 `:Entity` nodes, graphiti-core 0.29.2,
FalkorDB 4.18.11. Same query both ways, output sorted, timing footer stripped:

| | time |
|---|---|
| before | 33,383 ms |
| after | **1.6 ms** |

Result sets are byte-identical — md5 `86e64ec9d154a91dadffc3f7eac8e403` on both sides over
4,114 output lines for a 1,027-hit query. Also checked explicitly: self-loops (10 in the
graph) return with `source == target` on both sides, and `startNode` maps to the pattern's
left-hand variable `n`.

In practice this was the difference between an episode write taking ~12 minutes (often
exceeding the query timeout and failing) and ~45 seconds.

### Scope

FalkorDB only. Neo4j, Kuzu and Neptune paths are untouched — and neither Neo4j nor Kuzu has
this problem, because `get_relationships_query` passes their limit at the index call
(`{limit: $limit}` / `TOP := $limit`) while the FalkorDB branch passes none.

`ruff check` and `ruff format --check` pass on the changed file.


trevor-sykes:

trevor-sykes:

trevor-sykes:

trevor-sykes:
recheck

sgral:
Independent reproduction on a different graph and version — this fix is correct and the impact is not niche.

**Setup:** graphiti-core 0.29.3, FalkorDB, GTX 1080 Ti host, one graph per `group_id`. Same fulltext query run both ways, results deserialised through `entity_edge_from_record`.

| graph | `:Entity` | `RELATES_TO` | before | after |
|---|---|---|---|---|
| `homelab` | 1,008 | 992 | 0.35 s | 0.00 s |
| `n55tuning` | 2,377 | 3,439 | **57.39 s** | **0.05 s** |

3.5× the edges cost 162× the time. Result sets are identical — same UUIDs, same order.

The user-visible symptom is worth spelling out because it does not look like a performance problem: `add_episode` returns HTTP 200 "queued", the dedup search then exceeds the FalkorDB query timeout, and the episode is **silently dropped**. Small graphs completed and hid it, so the failure looked group-specific and random. Before finding this I had raised `max_tokens`, `OLLAMA_CONTEXT_LENGTH` and the FalkorDB `TIMEOUT` in turn; none of them helped, because none of them were the cause.

**One trap for anyone else patching this.** The same query exists twice:

- `graphiti_core/search/search_utils.py` — what this PR fixes
- `graphiti_core/driver/falkordb/operations/search_ops.py`

`search_utils.edge_fulltext_search` delegates to the driver-specific one only `if driver.search_interface`, and on `FalkorDriver` that attribute is `None`. So the driver-specific copy is dead code today, and patching only it builds clean, passes a grep assertion, and changes nothing at runtime. I shipped exactly that mistake first and the query kept timing out; the tell was `routing_='r'` in the logged params, which only the generic path passes. Worth noting since #1510 and #1471 currently touch only that file.

Also confirming the `WHERE n:Entity AND m:Entity` predicate matters if anyone is tempted to drop it: `RELATES_TO` is Entity→Entity by construction today, so results happen to match without it, but keeping it preserves the pattern's semantics rather than relying on that invariant.

Running this patch in production since 2026-08-02; the graph that had been dropping 100% of episodes now ingests normally.

Naseem77:
@trevor-sykes `GRAPH.PROFILE` confirms it: the old pattern scans 25,000 records for 50 fulltext hits over 500 entities, yours produces 50 with no label scan, and at 800+ entities the old query never finished. Results are identical and a `RELATES_TO` ending on `:Community` is still excluded, so the label predicate is equivalent. That equivalence is the only risky part and has no test, adding one would make this solid.

prasmussen15 (approved):

PR #1719, opened 2026-08-03
fix: prevent label_propagation infinite loop on edge_count >= 2

adnanbasil10:
## Summary
`build_communities()` hangs indefinitely (100% CPU, never returns) whenever the graph contains two entity pairs connected by 2 or more `RELATES_TO` edges - a common condition in real workloads, since multiple episodes referencing the same two entities naturally produce `edge_count >= 2` between them.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
N/A (bug fix to existing functionality, not a new feature).

**Root cause:** `label_propagation()` in `graphiti_core/utils/maintenance/community_operations.py` computed each round's new community assignments from a single snapshot of the previous round (a **synchronous** update) into a separate `new_community_map`, swapping it in only at the end of the round. When two nodes are joined by `edge_count >= 2`, each node's new label is computed from the *other's* old label at the same time, so they swap communities with each other every round, forever, and the `while True` loop never converges.

**Fix:** apply label updates asynchronously, in place, within the same pass - the standard correction for this class of algorithm. The second node in a pair now observes the first node's new label within the same round, which converges instead of oscillating. A bounded iteration cap (`MAX_LABEL_PROPAGATION_ITERATIONS = 100`, with a warning log if hit) was also added as a defensive backstop.

## Testing
- [x] All existing tests pass

Added `tests/utils/maintenance/test_community_operations.py` with 8 unit tests covering the exact reported repro (mutual double edge), a 3-node variant, isolated nodes, plurality-community joins, and `edge_count` in `{2, 3, 5, 10}`.

**Before the fix:** 5 of 8 tests failed with `TimeoutError` (hang reproduced via a daemon-thread timeout guard).
**After the fix:** 8/8 pass in 2.20s.

```
tests/utils/maintenance/test_community_operations.py::test_label_propagation_converges_with_mutual_double_edge PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_converges_with_triangle_of_double_edges PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_isolated_nodes_stay_separate PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_joins_plurality_community PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_converges_for_various_high_edge_counts[2] PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_converges_for_various_high_edge_counts[3] PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_converges_for_various_high_edge_counts[5] PASSED
tests/utils/maintenance/test_community_operations.py::test_label_propagation_converges_for_various_high_edge_counts[10] PASSED
8 passed in 2.20s
```

## Breaking Changes

## Checklist
- [x] Self-review completed
- [ ] Documentation updated where necessary (N/A - no user-facing docs affected)

## Related Issues
Closes #1355

### DIFF
--- a/graphiti_core/driver/operations/graph_utils.py
+++ b/graphiti_core/driver/operations/graph_utils.py
@@ -14,9 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import logging
 from collections import defaultdict
 
 from pydantic import BaseModel
+
+MAX_LABEL_PROPAGATION_ITERATIONS = 100
+
+logger = logging.getLogger(__name__)
 
 
 class Neighbor(BaseModel):
@@ -25,13 +30,33 @@ class Neighbor(BaseModel):
 
 
 def label_propagation(projection: dict[str, list[Neighbor]]) -> list[list[str]]:
-    community_map = {uuid: i for i, uuid in enumerate(projection.keys())}
+    # Implement the label propagation community detection algorithm.
+    # 1. Start with each node being assigned its own community
+    # 2. Each node will take on the community of the plurality of its neighbors
+    # 3. Ties are broken by going to the largest community
+    # 4. Continue until no communities change during propagation
+    #
+    # Nodes are processed in a fixed, deterministic order (sorted by uuid)
+    # rather than projection's insertion order, which may vary run-to-run
+    # since it is typically built from unordered DB query results. Because
+    # updates are applied asynchronously (see below), the result is sensitive
+    # to processing order, so a stable order is required for reproducible
+    # community assignments across runs on the same underlying graph (#1355).
+    node_order = sorted(projection.keys())
+    community_map = {uuid: i for i, uuid in enumerate(node_order)}
 
-    while True:
+    # Updates are applied asynchronously (in place), not from a single prior
+    # snapshot. A synchronous update lets two nodes joined by edge_count >= 2
+    # swap communities with each other every round forever, since both compute
+    # their new label from the other's *old* label at the same time (#1355).
+    # Updating in place means the second node in a pair already observes the
+    # first node's new label within the same pass, which converges instead of
+    # oscillating.
+    for _ in range(MAX_LABEL_PROPAGATION_ITERATIONS):
         no_change = True
-        new_community_map: dict[str, int] = {}
 
-        for uuid, neighbors in projection.items():
+        for uuid in node_order:
+            neighbors = projection[uuid]
             curr_community = community_map[uuid]
 
             community_candidates: dict[int, int] = defaultdict(int)
@@ -48,15 +73,18 @@ def label_propagation(projection: dict[str, list[Neighbor]]) -> list[list[str]]:
             else:
                 new_community = max(community_candidate, curr_community)
 
-            new_community_map[uuid] = new_community
-
             if new_community != curr_community:
+                community_map[uuid] = new_community
                 no_change = False
 
         if no_change:
             break
-
-        community_map = new_community_map
+    else:
+        logger.warning(
+            'label_propagation did not converge after %d iterations; '
+            'returning best-effort communities',
+            MAX_LABEL_PROPAGATION_ITERATIONS,
+        )
 
     community_cluster_map: dict[int, list[str]] = defaultdict(list)
     for uuid, community in community_map.items():
--- a/graphiti_core/utils/maintenance/community_operations.py
+++ b/graphiti_core/utils/maintenance/community_operations.py
@@ -2,9 +2,8 @@ import asyncio
 import logging
 from collections import defaultdict
 
-from pydantic import BaseModel
-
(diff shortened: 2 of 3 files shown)


adnanbasil10:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: adnanbasil2004@gmail.com

adnanbasil10:
Both points are correct and fixed.

**The duplication catch is the important one** - every driver-specific \`get_community_clusters()\` (Neo4j/FalkorDB/Kuzu/Neptune) imports its own \`label_propagation()\` from \`graphiti_core/driver/operations/graph_utils.py\`, separate from the copy in \`community_operations.py\` that this PR originally fixed. Since \`community_operations.get_community_clusters()\` delegates to \`driver.graph_operations_interface.get_community_clusters()\` whenever it's implemented (true for all four backends), the first fix never actually reached the path any real deployment uses.

Fixed by deduplicating rather than patching both copies separately: \`graph_utils.py\` is now the single source of truth, and \`community_operations.py\` imports \`Neighbor\`/\`label_propagation\` from there instead of defining its own. Added a test that asserts all five import sites (community_operations + the four driver graph_ops modules) resolve to the exact same function object, so a future re-split fails loudly instead of silently reintroducing the hang.

Also addressed the iteration-order concern - nodes are now processed in a fixed, sorted-by-uuid order rather than \`projection\`'s insertion order (which depends on unordered DB query results), so the same underlying graph always converges to the same community grouping. Added a test covering that directly.

Re-verified against the pre-fix \`graph_utils.py\` (i.e. today's actual backend code path before this update): 6 of 10 tests failed, including a reproduced hang. 10/10 pass after the fix, in 0.33s. Pushed in 5c3200e.

Naseem77:
@adnanbasil10 Verified live: `graph_utils.label_propagation` hangs on bipartite, disconnected pairs, chain and star shapes with `edge_count >= 2`, and all of them converge correctly with your patch. Removing the duplicate copy is the right call, that duplication is why the #1355 fix missed the path every backend calls. Supersedes #1710 and #1742. LGTM!

linhongyu510 (commented):
**Correctness issue: the fixed 100-iteration cap truncates normally converging connected graphs.**

I replayed `5c3200e` cleanly on current `main` (`c18d677`). The PR's 10 tests pass, and Ruff, format, and `git diff --check` are clean. However, the deterministic in-place traversal propagates a label by at most one hop per round on an unweighted chain whose UUIDs are in ascending order, so convergence can legitimately require `O(|V|)` rounds.

Pure-function reproduction against this PR:

```python
from graphiti_core.driver.operations import graph_utils as g

def chain(n):
    return {
        f"n{i:03d}": [
            g.Neighbor(node_uuid=f"n{j:03d}", edge_count=1)
            for j in (i - 1, i + 1)
            if 0 <= j < n
        ]
        for i in range(n)
    }

g.MAX_LABEL_PROPAGATION_ITERATIONS = 100
print([len(c) for c in g.label_propagation(chain(150))])
# 50 clusters: 49 singletons + one 101-node cluster

g.MAX_LABEL_PROPAGATION_ITERATIONS = 150
print([len(c) for c in g.label_propagation(chain(150))])
# [150]
```

The 102-node boundary already reproduces the problem: cap 100 returns two clusters (`[1, 101]`) even though the same connected chain reaches one fixed-point community with enough rounds. For 150 nodes, cap 100 returns 50 communities, while cap 150 returns one.

Please keep the hard termination guarantee, but make it graph-size-aware (for example, a configurable limit with a default of at least `max(100, len(projection))`) and add a chain regression above 100 nodes. Otherwise this fixes the infinite loop by silently returning a known non-fixed-point partition for ordinary larger components.

prasmussen15:
Not merging yet: @linhongyu510's review still applies. The fixed 100-iteration cap truncates normally converging chains that need O(|V|) rounds (e.g. a 150-node chain yields 50 communities at cap 100). Please make the cap graph-size-aware (at least `max(100, len(projection))`) and add a chain regression above 100 nodes.

prasmussen15 (changes requested):
This is not ready to merge.

The infinite-loop fix is real, but the hard 100-iteration cap truncates graphs that still converge. On an unweighted chain, in-place label propagation moves one hop per round, so a connected component can legitimately need O(|V|) rounds. With `MAX_LABEL_PROPAGATION_ITERATIONS = 100`, a 150-node chain returns 50 communities; with cap 150 it returns one. A 102-node chain already splits (`[1, 101]`) even though it has a single fixed-point community.

Please keep a hard termination guarantee, but make the cap graph-size-aware (for example a configurable limit with a default of at least `max(100, len(projection))`) and add a chain regression above 100 nodes. Until then this can silently return a non-fixed-point partition for ordinary larger components.

PR #1721, opened 2026-08-03
fix: Neo4j similarity search crashes on null/invalid embeddings

### LINKED ISSUE #1328: [BUG] `edge_similarity_search` crashes on edges with null/invalid `fact_embedding`

adnanbasil10:
## Summary
`edge_similarity_search`, `node_similarity_search`, and `community_similarity_search` in `graphiti_core/driver/neo4j/operations/search_ops.py` compute `vector.similarity.cosine()` over every matched row with no guard against a missing or dimension-mismatched embedding. As a graph grows over time, some `RELATES_TO` edges (or `Entity`/`Community` nodes) can end up with a `null`, empty, or dimension-mismatched embedding - e.g. after an embedder change, a partial write, or a migration. Neo4j's `vector.similarity.cosine()` raises `Neo.ClientError.Statement.ArgumentError` on either case, so a single bad row fails the **entire** search instead of just being excluded.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
N/A (bug fix to existing functionality, not a new feature).

**Fix:** add both an `IS NOT NULL` guard and a `size(...) = size($search_vector)` dimension-match guard on the relevant embedding property (`e.fact_embedding` / `n.name_embedding` / `c.name_embedding`) before the cosine call, in all three functions. This reuses the existing `filter_queries` list construction so it composes correctly with `group_id` and source/target filters.

## Testing
- [x] All existing tests pass

Added `tests/driver/test_neo4j_search_ops.py` with 6 unit tests that mock the query executor and assert on the constructed Cypher text (mirroring the existing `_build_neo4j_fulltext_query` query-shape test convention) - no live Neo4j instance required. Covers all three functions, both alone and combined with `group_id`/source/target filters, and asserts both guards gate the `MATCH` before the cosine call.

**Before the fix:** all 6 tests failed (no guards present in the generated query) in 3.31s.
**After the fix:** 6/6 pass in 0.28s (including the dimension-check addition requested in review).

```
tests/driver/test_neo4j_search_ops.py::test_node_similarity_search_excludes_null_embeddings PASSED
tests/driver/test_neo4j_search_ops.py::test_node_similarity_search_combines_null_check_with_group_ids PASSED
tests/driver/test_neo4j_search_ops.py::test_edge_similarity_search_excludes_null_embeddings PASSED
tests/driver/test_neo4j_search_ops.py::test_edge_similarity_search_combines_null_check_with_group_ids PASSED
tests/driver/test_neo4j_search_ops.py::test_community_similarity_search_excludes_null_embeddings PASSED
tests/driver/test_neo4j_search_ops.py::test_community_similarity_search_combines_null_check_with_group_ids PASSED
6 passed in 0.28s
```

## Breaking Changes

## Checklist
- [x] Self-review completed
- [ ] Documentation updated where necessary (N/A - no user-facing docs affected)

## Related Issues
Closes #1328

### DIFF
--- a/graphiti_core/driver/neo4j/operations/search_ops.py
+++ b/graphiti_core/driver/neo4j/operations/search_ops.py
@@ -142,6 +142,13 @@ class Neo4jSearchOperations(SearchOperations):
             filter_queries.append('n.group_id IN $group_ids')
             filter_params['group_ids'] = group_ids
 
+        # Exclude nodes with a missing/invalid embedding before computing cosine
+        # similarity - vector.similarity.cosine() raises
+        # Neo.ClientError.Statement.ArgumentError on a null vector, which
+        # otherwise fails the entire search once any node lacks (or has a
+        # stale-dimension) name_embedding (#1328).
+        filter_queries.append('n.name_embedding IS NOT NULL')
+
         filter_query = ''
         if filter_queries:
             filter_query = ' WHERE ' + (' AND '.join(filter_queries))
@@ -295,6 +302,13 @@ class Neo4jSearchOperations(SearchOperations):
         filter_queries, filter_params = edge_search_filter_query_constructor(
             search_filter, GraphProvider.NEO4J
         )
+
+        # Exclude edges with a missing/invalid embedding before computing cosine
+        # similarity - vector.similarity.cosine() raises
+        # Neo.ClientError.Statement.ArgumentError on a null vector, which
+        # otherwise fails the entire search once any edge lacks (or has a
+        # stale-dimension) fact_embedding (#1328).
+        filter_queries.append('e.fact_embedding IS NOT NULL')
 
         if group_ids is not None:
             filter_queries.append('e.group_id IN $group_ids')
@@ -490,9 +504,14 @@ class Neo4jSearchOperations(SearchOperations):
     ) -> list[CommunityNode]:
         query_params: dict[str, Any] = {}
 
-        group_filter_query = ''
+        # Exclude communities with a missing/invalid embedding before computing
+        # cosine similarity - vector.similarity.cosine() raises
+        # Neo.ClientError.Statement.ArgumentError on a null vector, which
+        # otherwise fails the entire search once any community lacks (or has a
+        # stale-dimension) name_embedding (#1328).
+        group_filter_query = ' WHERE c.name_embedding IS NOT NULL'
         if group_ids is not None:
-            group_filter_query += ' WHERE c.group_id IN $group_ids'
+            group_filter_query += ' AND c.group_id IN $group_ids'
             query_params['group_ids'] = group_ids
 
         cypher = (
--- a/graphiti_core/driver/neo4j/operations/search_ops.py
+++ b/graphiti_core/driver/neo4j/operations/search_ops.py
@@ -142,12 +142,13 @@ class Neo4jSearchOperations(SearchOperations):
             filter_queries.append('n.group_id IN $group_ids')
             filter_params['group_ids'] = group_ids
 
-        # Exclude nodes with a missing/invalid embedding before computing cosine
-        # similarity - vector.similarity.cosine() raises
-        # Neo.ClientError.Statement.ArgumentError on a null vector, which
-        # otherwise fails the entire search once any node lacks (or has a
-        # stale-dimension) name_embedding (#1328).
+        # Exclude nodes with a missing or dimension-mismatched embedding before
+        # computing cosine similarity - vector.similarity.cosine() raises
+        # Neo.ClientError.Statement.ArgumentError on a null vector, and errors
+        # on a dimension mismatch too, which otherwise fails the entire search
+        # once any node lacks (or has a stale-dimension) name_embedding (#1328).
         filter_queries.append('n.name_embedding IS NOT NULL')
+        filter_queries.append('size(n.name_embedding) = size($search_vector)')
 
         filter_query = ''
         if filter_queries:
@@ -303,12 +304,13 @@ class Neo4jSearchOperations(SearchOperations):
             search_filter, GraphProvider.NEO4J
         )
 
-        # Exclude edges with a missing/invalid embedding before computing cosine
-        # similarity - vector.similarity.cosine() raises
-        # Neo.ClientError.Statement.ArgumentError on a null vector, which
-        # otherwise fails the entire search once any edge lacks (or has a
-        # stale-dimension) fact_embedding (#1328).
+        # Exclude edges with a missing or dimension-mismatched embedding before
+        # computing cosine similarity - vector.similarity.cosine() raises
+        # Neo.ClientError.Statement.ArgumentError on a null vector, and errors
+        # on a dimension mismatch too, which otherwise fails the entire search
+        # once any edge lacks (or has a stale-dimension) fact_embedding (#1328).
         filter_queries.append('e.fact_embedding IS NOT NULL')
+        filter_queries.append('size(e.fact_embedding) = size($search_vector)')
 
         if group_ids is not None:
             filter_queries.append('e.group_id IN $group_ids')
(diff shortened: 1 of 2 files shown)


adnanbasil10:
Good catch - the original issue report also called out dimension-mismatched embeddings, not just null ones, and the first pass here missed that case. Added a \`size(...) = size(\$search_vector)\` check alongside the \`IS NOT NULL\` guard in all three functions (node/edge/community similarity search), and extended the query-shape tests to assert both guards gate the \`MATCH\` before the cosine call. 6/6 tests pass in 0.28s. Pushed in 1bea4a5.

adnanbasil10:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: adnanbasil2004@gmail.com

adnanbasil10:
recheck

linhongyu510 (commented):
**Correctness gap: same-dimension NaN embeddings still crash the query.**\n\nI replayed `1bea4a5` cleanly on current `main` (`c18d677`) and the added suite passes (`6 passed`); Ruff, format, and `git diff --check` also pass. However, the issue explicitly includes NaN embeddings, while the new guards only check non-nullness and dimension.\n\nAgainst a real `neo4j:5.26-community` instance:\n\n```cypher\nWITH [1.0, toFloat("NaN")] AS v\nWHERE v IS NOT NULL AND size(v) = size([1.0, 2.0])\nRETURN vector.similarity.cosine(v, [1.0, 2.0])\n```\n\nstill fails with `Invalid input for vector.similarity.cosine(): Argument a is not a valid vector for this similarity function.` The current query-shape tests cannot catch this because they only assert string placement. A finite-value predicate such as `all(x IN v WHERE NOT isNaN(x))` skips this row in the same Neo4j instance; please add the equivalent guard for node/edge/community embeddings and ideally one live-query regression (or narrow the PR claim if NaN is intentionally out of scope).\n\nFor completeness, Neo4j 5.26 returned `NULL` for a null operand and rejected dimension mismatch as expected; the demonstrated unresolved case is specifically same-dimension NaN.

linhongyu510:
Formatting correction for my review above (the API rendered its line breaks literally):

**Correctness gap: same-dimension NaN embeddings still crash the query.**

I replayed `1bea4a5` cleanly on current `main` (`c18d677`) and the added suite passes (`6 passed`); Ruff, format, and `git diff --check` also pass. However, the issue explicitly includes NaN embeddings, while the new guards only check non-nullness and dimension.

Against a real `neo4j:5.26-community` instance:

```cypher
WITH [1.0, toFloat("NaN")] AS v
WHERE v IS NOT NULL AND size(v) = size([1.0, 2.0])
RETURN vector.similarity.cosine(v, [1.0, 2.0])
```

still fails with `Invalid input for vector.similarity.cosine(): Argument a is not a valid vector for this similarity function.` The current query-shape tests cannot catch this because they only assert string placement. A finite-value predicate such as `all(x IN v WHERE NOT isNaN(x))` skips this row in the same Neo4j instance; please add the equivalent guard for node/edge/community embeddings and ideally one live-query regression (or narrow the PR claim if NaN is intentionally out of scope).

For completeness, Neo4j 5.26 returned `NULL` for a null operand and rejected dimension mismatch as expected; the demonstrated unresolved case is specifically same-dimension NaN.

linhongyu510:
Small correction to my suggested predicate: `NOT isNaN(x)` excludes NaN but is not by itself a full finite-value check. I also verified on the same Neo4j 5.26 instance that `toFloat("Infinity")` passes `NOT isNaN(x)` and still makes `vector.similarity.cosine()` fail. This complete predicate excluded both NaN and ±Infinity while retaining valid vectors in the live query: `all(x IN v WHERE NOT isNaN(x) AND abs(x) <= 1.7976931348623157e308)`. The core finding remains: same-size invalid coordinates bypass the current PR guards; the exact production predicate should cover both non-finite cases.

Issue #1087, opened 2025-11-29
[BUG] Embedding truncation reduces retrieval quality for text-embedding-3-small


mengbangjie88:
## Bug Description

The `OpenAIEmbedder` in `graphiti_core/embedder/openai.py` hard-truncates embeddings using array slicing `[: self.config.embedding_dim]`, which causes information loss and degrades retrieval quality.

**Key Issue:**
- Default `EMBEDDING_DIM` is 1024 (defined in `graphiti_core/embedder/client.py`)
- Default embedding model `text-embedding-3-small` produces 1536-dimensional embeddings
- The code truncates embeddings from 1536 to 1024 dimensions (~35% information loss)
- This happens in both `create()` (line 60) and `create_batch()` (line 66) methods

## Steps to Reproduce

```python
from graphiti_core.embedder.openai import OpenAIEmbedder, OpenAIEmbedderConfig

# Create embedder with default config
config = OpenAIEmbedderConfig()  # embedding_dim defaults to 1024
embedder = OpenAIEmbedder(config=config)

# Create an embedding
embedding = await embedder.create("test text")

# Check the dimension
print(f"Embedding dimension: {len(embedding)}")  # Output: 1024

# However, text-embedding-3-small actually produces 1536-dimensional embeddings
# The last 512 dimensions are discarded by the truncation at line 60:
# return result.data[0].embedding[: self.config.embedding_dim]
```

**Code location:**
- `graphiti_core/embedder/openai.py` line 60: `return result.data[0].embedding[: self.config.embedding_dim]`
- `graphiti_core/embedder/openai.py` line 66: `return [embedding.embedding[: self.config.embedding_dim] for embedding in result.data]`

## Expected Behavior

When using OpenAI's `text-embedding-3-small` model with a target dimension of 1024, the API should be called with the `dimensions` parameter to generate embeddings at the target dimension directly. This preserves the full information content of the embedding at the specified dimension, rather than generating a higher-dimensional embedding and then discarding part of it.

The OpenAI API supports the `dimensions` parameter for `text-embedding-3-small` and `text-embedding-3-large` models, allowing direct generation of embeddings at the desired dimension.

## Actual Behavior

The code generates a full 1536-dimensional embedding from `text-embedding-3-small`, then hard-truncates it to 1024 dimensions using array slicing. This discards the last 512 dimensions, losing approximately 35% of the embedding information. This information loss negatively impacts retrieval quality in semantic search operations.

## Environment

[...]

PR #1722, opened 2026-08-03
feat(embedder): support native OpenAI embedding dimensions instead of truncating (#1087)

Anai-Guo:
## What

Adds an opt-in `dimensions` field to `OpenAIEmbedderConfig` so `OpenAIEmbedder` can request natively reduced embeddings via the OpenAI [`dimensions`](https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions) parameter, instead of naively slicing the full vector.

Related to #1087.

## Why

Today `create`/`create_batch` return `result.data[0].embedding[: self.config.embedding_dim]`. Slicing a `text-embedding-3-*` vector drops the trailing components **without renormalizing**, so the resulting vector is no longer unit-length and cosine similarity degrades — the exact retrieval-quality issue raised in #1087 (most severe for `text-embedding-3-large`, 3072 → 1024 = 66% dropped).

The OpenAI models support generating the embedding at a target dimensionality server-side, which returns a properly normalized reduced vector. As @danielchalef noted on the issue, truncation is intentional for storage cost, so this is added as an **opt-in** rather than a behavior change.

## Behavior

- `dimensions=None` (default): **unchanged** — full vector requested and truncated to `embedding_dim`. Safe for models/endpoints that don't accept the parameter (e.g. `text-embedding-ada-002`, custom `base_url` deployments).
- `dimensions=N`: `N` is forwarded to the API and the natively sized, renormalized vector is returned without truncation. Users setting this should match `embedding_dim` for downstream storage consistency (documented on the field).

## Tests

Extends `tests/embedder/test_openai.py`:
- existing tests assert `dimensions` is **not** sent when unconfigured;
- new tests assert it is forwarded and that the full native vector is returned (no truncation) for both `create` and `create_batch`.

All four pass; the two new tests fail against the pre-change implementation, confirming they exercise the new path. `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/graphiti_core/embedder/openai.py
+++ b/graphiti_core/embedder/openai.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 from collections.abc import Iterable
+from typing import Any
 
 from openai import AsyncAzureOpenAI, AsyncOpenAI
 from openai.types import EmbeddingModel
@@ -28,6 +29,16 @@ class OpenAIEmbedderConfig(EmbedderConfig):
     embedding_model: EmbeddingModel | str = DEFAULT_EMBEDDING_MODEL
     api_key: str | None = None
     base_url: str | None = None
+    dimensions: int | None = None
+    """Requested output dimensionality.
+
+    When set, it is forwarded to the OpenAI ``dimensions`` parameter (supported by
+    ``text-embedding-3-*`` models) so the API returns a natively reduced, renormalized
+    vector instead of one that is naively truncated. Leave as ``None`` (the default) for
+    backwards-compatible behavior and for models/endpoints that do not support the
+    parameter (e.g. ``text-embedding-ada-002`` or custom ``base_url`` deployments). When
+    set, prefer matching :attr:`embedding_dim` so downstream storage stays consistent.
+    """
 
 
 class OpenAIEmbedder(EmbedderClient):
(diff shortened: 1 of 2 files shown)


Anai-Guo:
Picking this up per @danielchalef's note on #1087 (the original reporter offered a PR ~8 months ago but none was filed). Kept it strictly opt-in so the default truncation behavior — and support for endpoints that don't accept `dimensions` — is unchanged. Happy to adjust the API shape (e.g. tie it to `embedding_dim` instead of a separate field) if you'd prefer.

xiaoyaner0201:
Adding a production data point in support of this, plus the workaround we ended up building — it may be useful to others hitting the same wall while this is open.

## Deployment

We run Graphiti (Neo4j backend) against a **non-OpenAI, OpenAI-compatible** embedding endpoint: Alibaba DashScope serving `qwen3.7-text-embedding` at **2560 dimensions**. Current graph: 48,647 `Entity` nodes / 76,333 `RELATES_TO` edges, all with 2560-dim `fact_embedding`.

That configuration cannot use `OpenAIEmbedder` as-is. `embedding_dim` only controls the **slice width** applied client-side:

```python
result = await self.client.embeddings.create(
    input=input_data, model=self.config.embedding_model
)
return result.data[0].embedding[: self.config.embedding_dim]
```

Since `dimensions` is never forwarded, the endpoint returns its own default width and Graphiti truncates to `embedding_dim`. For a provider whose native width is not what you asked for, `embedding_dim` silently becomes "hope the upstream default is at least this large", and the stored vector is a non-renormalized prefix — the exact degradation @danielchalef's reply and this PR describe.

## How we worked around it

Rather than patching `graphiti_core`, we put an **OpenAI-compatible sidecar proxy** between Graphiti and the provider, and pointed `EMBEDDER_OPENAI_API_URL` at it. For the one target model it:

1. overwrites or injects `dimensions=2560` into the outbound request;
2. forces `encoding_format=float`, because the OpenAI SDK may request base64 — without this the proxy cannot inspect the returned vectors;
3. forwards to the real upstream, passing `Authorization` through without logging it;
4. validates response cardinality and that every returned `index` is present;
5. validates every embedding is finite and **exactly** 2560-dim, failing loudly rather than storing a malformed vector;
6. sorts `data` by `index` before returning.

Requests for any other model pass through untouched.

Step 1 is precisely what this PR adds natively, and we would drop it immediately. Steps 2 and 4–6 are separate concerns (SDK encoding negotiation and response integrity) and are not something this PR needs to take on — I mention them only to be clear about what this PR would and would not let us delete.

Point 5 is worth calling out on its own: a silently wrong-width or non-finite embedding is very hard to trace back once it is in the graph. #1505 reports NaN/Inf embeddings breaking entity deduplication downstream, and #1332 / #1721 are both open on null/invalid embeddings crashing similarity search. Not forwarding `dimensions` widens the window for that class of bug, because the width mismatch is never detected at the boundary.

## On the PR itself

The opt-in shape looks right to me:

- `dimensions=None` keeping the current truncation behavior is the safe default, and matters specifically for OpenAI-compatible endpoints that reject unknown parameters — a hard failure there would be worse than the current silent truncation.
- Forwarding `dimensions` and skipping truncation is the correct pairing; requesting native width and *then* slicing would reintroduce the same problem.

One documentation-level suggestion, from the perspective of the third-party-endpoint case: the field docs currently frame `dimensions` as matching `embedding_dim` for storage consistency. For non-OpenAI backends it is stronger than that — it is the only way to make `embedding_dim` mean the actual vector width rather than a truncation bound. Worth a sentence, since that is the configuration most likely to be silently wrong today.

Related: hardcoded dimension assumptions show up elsewhere in the codebase too. #859 created Neo4j vector indexes with `` `vector.dimensions`: 1024 `` hardcoded; I filed #1793 about the Neo4j vector-index gap and made deriving the dimension from the configured embedder an explicit requirement there, with a regression test pinning it. Same root cause, different surface.

This PR has been mergeable and green since 2026-08-03. @danielchalef explicitly invited this contribution on #1087 — happy to help with review or testing against a non-OpenAI endpoint if that would move it along.

PR #1726, opened 2026-08-03
fix(mcp): get_episodes returns the most recent episodes, not a uuid-ordered slice

### LINKED ISSUE #1062: [BUG] get_episodes from graphiti-mcp returns stale/old data when used in n8n
## Bug Description
I am using the graphiti-mcp (version 1.0.1) as an external tool within an n8n workflow. The get_episodes method consistently returns outdated episode data immediately after a new episode has been successfully added to the graph using another tool/node.

## Steps to Reproduce

# This issue is reproduced using an n8n workflow. The logic is as follows:

# Tool executed: add_episode
# Parameters: {"max_episodes": 5} or null

## Expected Behavior
I expected the result should be the most recent entry. 

## Actual Behavior
The results return data from six months ago.
There are tests that add new data, but the result still returns old data.

## Environment
- Graphiti Version: 1.0.1

## Installation Method
- Docker

## Error Messages/Traceback
```
No explicit error/traceback. The issue is incorrect/stale data returned
```

## Configuration
Standard GraphitiMCP configuration.

## Additional Context
This situation only occurs after upgrading to the new version; the old version works fine.

## Possible Solution
This might be related to a database transaction commit or caching issue within the Graphiti MCP server, where the data added via add_episode is not immediately committed or visible to the subsequent get_episodes read operation.

### LINKED ISSUE #1724: MCP get_episodes returns episodes ordered by uuid, not recency — regression since MCP v1.0.0
## Summary

The MCP server's `get_episodes` tool returns episodes sorted by uuid descending. Because uuid4 values are random, the result is effectively an arbitrary fixed subset of a group's episodes: newly ingested episodes usually do not appear, and the same stale set is returned indefinitely no matter how much new content is added.

This is a regression introduced in MCP server v1.0.0 (#1024). Before that release the tool returned episodes by recency, and its docstring said so.

I believe this is also the true root cause of #1062, which is currently attributed to ingestion-queue latency.

## Impact

`get_episodes` is the natural tool for an agent to answer "what has been stored recently?" or "did the episode I just added get ingested?" It cannot answer either question today.

The failure is silent. The tool returns a plausible-looking list of real episodes with no error and no indication that ordering is unrelated to time, so an agent reasonably concludes that recent writes were lost.

## Root cause

`get_episodes` calls `EpisodicNode.get_by_group_ids()`:

```python
# mcp_server/src/graphiti_mcp_server.py
episodes = await EpisodicNode.get_by_group_ids(
    client.driver, effective_group_ids, limit=max_episodes
)
```

That method is the keyset-pagination helper, and it hardcodes uuid ordering to match its `uuid_cursor` parameter:

```python
# graphiti_core/nodes.py  (EpisodicNode.get_by_group_ids)
cursor_query: LiteralString = 'AND e.uuid < $uuid' if uuid_cursor else ''
...
ORDER BY uuid DESC
```

For pagination this is correct: a keyset cursor must sort on the column it filters, and that column must be unique. The problem is only that a browsing tool reuses it.

Two details make the reuse hard to notice and impossible to work around:

1. **`get_episodes` exposes no cursor.** Its parameters are `group_ids` and `max_episodes` only. The single benefit of uuid ordering — stable paging — is unreachable through the tool, so the ordering it inherits buys the caller nothing.

2. **Sorting the response client-side cannot fix it.** `LIMIT` is applied inside the uuid-ordered query, so a caller who re-sorts the returned page is sorting the wrong rows, not merely the wrong order.

## Regression point

The pre-1.0 MCP server ordered by recency, via `retrieve_episodes` (which orders by `valid_at DESC`):

```python
# mcp_server/graphiti_mcp_server.py, before #1024
async def get_episodes(group_id: str | None = None, last_n: int = 10):
    """Get the most recent memory episodes for a specific group. ..."""
    episodes = await client.retrieve_episodes(
        group_ids=[effective_group_id], last_n=last_n,
    )
```

The v1.0.0 rewrite switched to `EpisodicNode.get_by_group_ids`, renamed `last_n` to `max_episodes`, and dropped "most recent" from the docstring. The behaviour change appears incidental rather than intended.

This matches the report in #1062 exactly, including the reporter's observation that "this situation only occurs after upgrading to the new version; the old version works fine." Queue latency, the current explanation on that issue, would delay an episode's appearance briefly; it would not return six-month-old episodes indefinitely across repeated writes. Random uuid ordering does.

## How uuid ordering came to be

I traced this expecting a deliberate design decision and did not find one recorded.

| PR | Date | Change |
|---|---|---|
| [#218](https://github.com/getzep/graphiti/pull/218) | 2 Dec 2024 | Added keyset pagination to `get_by_group_ids`. Cursor filtered on `created_at`, but the query already ordered `BY uuid DESC` — paging on one key, sorting by another. |
| [#322](https://github.com/getzep/graphiti/pull/322) | 8 Apr 2025, 14:48 UTC | Resolved the mismatch toward time: `ORDER BY n.created_at DESC` for nodes, `ORDER BY e.created_at, e.uuid DESC` for edges. Title: "for better chronological sorting". |
| [#335](https://github.com/getzep/graphiti/pull/335) | 8 Apr 2025, 21:59 UTC | Resolved it the other way instead: cursor became `uuid_cursor`, filter became `AND n.uuid < $uuid`, ordering returned to `uuid DESC`. |

#335 landed seven hours after #322, with no commits in between, and reversed it.

**What the record says about intent.** All three PR descriptions are generated summaries. Across the three there are zero human comments, no linked issues, and no cross-references. The commit messages are "update pagination", "update pagination", "uuid_cursor".

The one substantive review remark went unanswered. On #322 the review bot asked whether `ORDER BY e.created_at, e.uuid DESC` was intentional, since it sorts `created_at` *ascending* while the sibling node queries use `DESC`; it also suggested adding uuid as a secondary sort so ties stayed deterministic. No reply, and #335 replaced the clause entirely hours later.

So the strongest available reading of intent is structural rather than documented: a keyset cursor needs a unique, monotonic sort key, and `created_at` is not one. It is assigned client-side in Python at model construction (`created_at: datetime = Field(default_factory=lambda: utc_now())`), not by the database, so bulk ingestion produces ties, and ties under a strict `<` cursor either skip rows or stall the page. uuid is unique and indexed. Read that way, #335 is correct for pagination — and #322's edge clause was independently broken, since ascending `created_at` cannot serve a descending cursor.

If a maintainer remembers a different reason, that would be worth capturing in a comment; I could not find one in the repository.

## Scope

Only `get_episodes` is reported here, but the same ordering is inherited by every `get_by_group_ids` on the same pattern — `EntityNode`, `CommunityNode`, `SagaNode` (`graphiti_core/nodes.py:455`, `673`, `853`, `1013`) and all five edge classes (`graphiti_core/edges.py:247`, `527`, `675`, `808`, `943`).

`get_episodes` is the only MCP tool affected. The one other reachable call site, `summarize_saga`, also goes through `SagaNode.get_by_group_ids`, but it passes no `limit` and then resolves a saga by name in Python — so it reads every saga in the group and the uuid ordering is harmless there.

I would keep the pagination helper as it is in all cases. The fix belongs at the call sites that want recency.

## Options

Assuming the goal is "listing episodes by recency is possible", three shapes:

| Option | Behaviour | Trade-off |
|---|---|---|
| **A. Keep uuid default, add `order_by`** | Existing callers unchanged; `order_by='valid_at'` or `'created_at'` opts in. | Preserves a default that serves no purpose in a tool with no cursor, and leaves #1062's reporter — and every new user — on the broken path unless they know to ask. |
| **B. Default `valid_at`, `order_by` accepts `created_at` and `uuid`** *(recommended)* | Recency by default; `created_at` for ingest verification; `uuid` retained as an escape hatch. | Changes the default. Mitigated by the fact that this *restores* documented pre-1.0 behaviour rather than inventing new behaviour. |
| **C. Default `created_at`** | Sorts by ingest time. | Answers "did my write land?" but misorders backfilled content: an episode about May events ingested today sorts ahead of July's episodes. Wrong for browsing, which is the tool's main use. |

**Recommendation: B.** `get_episodes` is a browsing tool, not a retrieval path — relevance recall goes through `search_memory_facts` / `search_nodes` — so its default should be the one a human or agent means by "recent". Since episodes are immutable and carry no `invalid_at`, that is `valid_at`: when the described events occurred. `created_at` stays available for the ingest-verification case, and `valid_at` was the pre-1.0 default, so B is the smaller change to the tool's contract despite being the larger change to current code behaviour.

Two related fixes belong with it:

- **Return `valid_at` in the response.** The payload is currently `uuid`, `name`, `content`, `created_at`, `source`, `source_description`, `group_id`. Sorting by a field the caller cannot see is not verifiable.
- **Do not filter by `reference_time`.** `retrieve_episodes` applies `valid_at <= reference_time`, which hides future-dated episodes. For a browsing tool that silently drops rows; ordering should be applied without the filter.

## Proposed PR outline

Small and additive. Feedback encouraged. 🙏

1. **`graphiti_core`:** add `EpisodicNode.get_recent_by_group_ids(driver, group_ids, limit, order_by)` emitting `ORDER BY e.<key> DESC, e.uuid DESC` — the uuid tiebreak keeps ties deterministic, which is what the #322 review asked for. New method rather than a parameter on `get_by_group_ids`, so the pagination helper and its callers are untouched.
2. **MCP server:** add `order_by: Literal['valid_at', 'created_at', 'uuid'] = 'valid_at'` to `get_episodes`, route to the new method, and document in the docstring that `valid_at` is event time and `created_at` ingest time — the distinction an agent needs to pick correctly.
3. **Response:** include `valid_at` in the episode dict.
4. **Tests:** episodes whose uuid order, `created_at` order, and `valid_at` order all disagree — asserting each `order_by` value returns the right sequence, and that the uuid-ordered pagination helper is unchanged. A backfill case (recent `created_at`, old `valid_at`) covers the distinction that motivates the default.
5. **Docs:** note the restored behaviour in the MCP server README / changelog, since v1.0.0 changed it silently.

## Environment

- graphiti-core v0.29.3, and confirmed present on `main` at the time of writing
- MCP server, streamable-http transport
- FalkorDB backend (the FalkorDB driver does not override `episodic_node_get_by_group_ids`, so the generic uuid-ordered query in `nodes.py` runs)
- Gemini LLM and embedder

## Reproduction

1. Add several episodes to a group over time, then add one more.
2. Call `get_episodes(group_ids=['<group>'], max_episodes=8)`.
3. Observe: results are in strictly descending uuid order, `created_at` values appear in arbitrary sequence spanning months, and the episode just added is absent unless its uuid happens to sort high.

In my own run, the eight returned uuids descended monotonically while their `created_at` values jumped across a three-month range in no order.

## Related

- #1062 — same symptom, currently attributed to ingestion-queue latency. I believe this is the root cause.
- #1024 — the v1.0.0 MCP rewrite where the behaviour changed.
- #218, #322, #335 — origin of uuid ordering.

brentkearney:
## Summary

`get_episodes` sorts by uuid, so it returns an arbitrary slice of a group rather than its most recent episodes. This PR gives episode listing a time ordering, defaulting to `valid_at`.

Background, provenance, and the alternatives considered are in #1724. This implements the option recommended there.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

### `graphiti_core`

New `EpisodicNode.get_recent_by_group_ids(driver, group_ids, limit, order_by)`, emitting `ORDER BY <key> DESC, uuid DESC`. The uuid tiebreak makes the ordering total, so a page is stable across identical calls even when timestamps collide.

It is a new method rather than an argument on `get_by_group_ids`, because that method backs keyset pagination: its cursor filters on uuid and so must sort on uuid too. Leaving it untouched keeps paging correct and this change additive.

### MCP server

`get_episodes` takes `order_by`, one of:

- **`valid_at`** (default) — when the events an episode describes happened. What "most recent" usually means, and what the tool returned before v1.0.0.
- **`created_at`** — when the episode was ingested. The one to use for "did the episode I just added arrive?". Not the default, because a backfilled episode about last May has a recent ingest time and an old event time, which is the wrong order for browsing.
- **`uuid`** — the current behaviour, for anyone relying on it.

Responses now include `valid_at` next to `created_at`, so a caller can see the field it sorted on.

`reference_time` filtering is deliberately not applied. `retrieve_episodes` drops rows where `valid_at > reference_time`, which is right for point-in-time retrieval but would silently hide future-dated episodes from a listing.

### Driver interface

`episodic_node_get_recent_by_group_ids` added to `GraphOperationsInterface` with the usual `NotImplementedError` default, so a driver can specialise it. Drivers that don't fall through to the generic query.

## Testing

- [x] All existing tests pass

**`tests/test_episode_recency_ordering.py`** — 13 unit tests against a mock driver, so no database is needed. They pin the emitted `ORDER BY` for each `order_by` value, the uuid tiebreak, that `LIMIT` follows the ordering, that an invalid `order_by` raises before any query runs, that no cursor predicate is emitted, that Neptune keeps its own return clause, and that rows come back in driver order. Two are regression guards on `get_by_group_ids`: still uuid-ordered, still accepts its cursor.

**`test_episodic_node_recency_ordering`** in `tests/test_node_int.py` — three episodes whose uuid, `created_at`, and `valid_at` orders all disagree, so each ordering is distinguishable from the other two. One is a backfill: newest ingest time, oldest event time. It also checks that a truncated page contains the newest episodes rather than an arbitrary subset. Green on FalkorDB v4.18.10.

As a control I inverted the default ordering to `ASC` and confirmed the integration test fails, so it is asserting the order rather than passing by coincidence.

Existing suites: 401 passed, 35 skipped with `-k "not _int"`. `tests/embedder/test_voyage.py` is excluded from that count — it fails collection on my environment with or without this change, under Python 3.14.

## Breaking Changes

- [x] This PR contains breaking changes

`get_episodes` returns different episodes than before — the recent ones. No argument is removed or newly required, and `order_by='uuid'` reproduces the old behaviour exactly.

I raised this trade-off in #1724 and still think changing the default is right, since it restores what the tool did before v1.0.0 rather than introducing new behaviour, and leaving uuid as the default keeps every new user on the broken path unless they know to ask. Happy to make it opt-in instead if you would rather not move the default.

`graphiti_core` is additive only.

## Checklist

- [x] Self-review completed

`make lint` is clean — ruff plus pyright, no errors. Docs: the MCP server's tool list and the `get_episodes` entry in `mcp_server/README.md` now say which clock each `order_by` value selects.

## Related Issues

Closes #1724
Also likely closes #1062, which reports the same symptom and is currently attributed to queue latency.


brentkearney:

brentkearney:
> This bot will be retriggered when the Contributor License Agreement comment has been provided. Posted by the **CLA Assistant Lite bot**.

Apparently not...

brentkearney:
recheck

Issue #1666, opened 2026-07-18
Contradiction detection in EdgeDuplicate (dedupe judge) collapses on non-reasoning small_models — a reasoning-first response schema restores it (measured)


mecampbellsoup:
**Summary.** When `small_model` is a non-reasoning model (we run `gpt-4.1-nano` for cost; the library default is a reasoning model, which likely masks this), the stock `EdgeDuplicate` response schema produces near-total failure on contradiction detection in `resolve_edge`: in a graded eval, **1 of 9 replicates correct** across contradiction-bearing cases, including missing a two-fact clear case ("User sold X" vs "User owns X"). Duplicate detection is unaffected. Since `contradicted_facts` drives `invalid_at`, this means temporal invalidation silently underperforms for anyone on a non-reasoning judge — stale facts survive their own contradiction.

**Measurement.** 5 ground-truth cases (clear-contradiction / clear-duplicate / neither / 4-fact index-stress / cross-list invalidation-candidate), 3 replicates each, called through `prompt_library.dedupe_edges.resolve_edge` + `generate_response(model_size=small)` with production-faithful context shapes (lists of `{'idx': N, 'fact': ...}` dicts); model/temperature/schema verified on the wire per call. Stock schema: **7/15**. Same everything, with a `reasoning: str` field declared *before* the two arrays: **14/15**.

| Case | Stock `EdgeDuplicate` | Reasoning-first variant |
|---|---|---|
| clear-contradiction (2 facts) | 1/3 | 3/3 |
| clear-duplicate | 3/3 | 3/3 |
| neither | 3/3 | 3/3 |
| index-stress (4 facts, contradiction at idx 2) | 0/3 | 2/3 |
| cross-list invalidation-candidate idx | 0/3 | 3/3 |

```python
class EdgeDuplicateWithReasoning(BaseModel):
    reasoning: str = Field(..., description=(
        "Step-by-step reasoning about duplicates and contradictions, written "
        "BEFORE the final answer. For every duplicate or contradiction you "
        "identify, explicitly state its idx number in this reasoning "
        "(e.g. 'idx 2 is contradicted because...') — do not just describe "
        "the fact's content, name the idx."))
    duplicate_facts: list[int] = Field(...)      # unchanged
    contradicted_facts: list[int] = Field(...)   # unchanged
```

[...]

PR #1729, opened 2026-08-04
fix: only invalidate edges the new edge could plausibly replace

### LINKED ISSUE #1728: Bug: Edge invalidation searches the whole graph, so unrelated facts retire each other
## Environment

- **Graphiti version**: based on 0.29.3. Line numbers below are from a fork with unrelated local patches, so they may be off by a few lines from stock 0.29.3, but the code quoted is unmodified.
- **Database backend**: FalkorDB
- **LLM provider**: Google Gemini (small model used for the dedupe/contradiction judge)

## Problem

Saving a memory that merely *mentions* an entity can retire an unrelated, still-true fact about that entity. The retired fact stays in the graph but is stamped with an `invalid_at` date, so searches downrank or drop it and the knowledge is effectively gone — silently, with no signal that anything was lost.

In one production graph of roughly 3,950 facts, 1,616 (41%) carry an `invalid_at`. A hand-audit of four of them found three were collateral rather than genuine change:

| Retired fact | Killed by | Verdict |
|---|---|---|
| A person holds a particular job title at a company (`WORKS_AT` person→company) | The same person administers that company's source-control organization (`ResponsibleFor` person→org account) | **wrong** — additive role, different target entity |
| A person is employed by a company, where they manage a compliance function (`WORKS_AT` person→company) | The same person imposed an audit requirement (`IMPOSED_REQUIREMENT` person→requirement) | **wrong** — unrelated fact, different target entity |
| A company must maintain a security-standard posture (`SUBJECT_TO` company→posture) | That company requires a review against the same standard (`REQUIRES` company→standards review) | **wrong** — narrower restatement, different target entity |
| `helper.py` is a script belonging to a plugin (`PART_OF` .py file→plugin) | `helper.rb` is a script belonging to that plugin (`PART_OF` .rb file→plugin) | **right** — the script really was ported to Ruby |

Four is too small a sample to characterize the other 1,612. But the three wrong cases share a shape: the killing edge shares *one* endpoint with its victim — the same person, the same company — while its other endpoint is a completely different thing, and it carries a different relation name. The one correct case has the opposite shape: a different endpoint, but the same relation name (`PART_OF`) pointing at the same shared parent.

The third case has a second problem worth naming: "posture" and "standards review" are arguably the same entity and should have merged during node deduplication. That's a separate weakness. It doesn't explain the first two, where the target entities are unmistakably different things.

## Why it happens

**1. The candidate pool lost its scoping in a refactor.** Two searches run per extracted edge in `graphiti_core/utils/maintenance/edge_operations.py`. Duplicate detection is scoped to edges between the same two nodes (line 399). Invalidation candidates are not scoped at all (line 414):

```python
search(
    clients,
    extracted_edge.fact,
    group_ids=[extracted_edge.group_id],
    config=EDGE_HYBRID_SEARCH_RRF,
    search_filter=SearchFilters(),   # no filter: any edge in the group is eligible
)
```

This looks like an unintended regression. Before #906 the same code path called a purpose-built helper:

```python
get_edge_invalidation_candidates(driver, extracted_edges, SearchFilters(), 0.2)
```

which restricted candidates to edges touching one of the new edge's endpoints:

```cypher
WHERE n.uuid IN [edge.source_node_uuid, edge.target_node_uuid]
   OR m.uuid IN [edge.target_node_uuid, edge.source_node_uuid]
```

#906 routed both searches through the generic `search()` so they would work across backends — necessary, since the helpers above issue raw Cypher. The duplicate search kept its scoping, because `edge_uuids` was added to `SearchFilters` in the same PR. The invalidation search had no equivalent filter to move to, so it went out unscoped and now nominates any semantically similar edge in the entire group. `get_edge_invalidation_candidates` still exists in `graphiti_core/search/search_utils.py` and is no longer reachable from ingestion — only from tests. The old `min_score` of `0.2` was dropped in the same change.

**2. Nothing checks the LLM's contradiction call.** Both candidate lists go to the LLM as bare fact strings, with no indication of which entities each fact connects (lines 700–713). Whatever comes back in `contradicted_facts` becomes the invalidation set. From there:

- `resolve_edge_contradictions` (lines 538–573) applies a purely temporal rule — if the candidate became valid earlier than the new edge, retire it. It has no way to tell whether the two facts are about the same thing.
- Lines 826–839 apply the mirror rule: if a flagged candidate became valid *later*, the new edge is written with `invalid_at` already set — born retired.

So one misjudgement by a small model, on facts stripped of their entity context, is acted on in both directions with nothing downstream to catch it. #1666 measures how unreliable that judge is on non-reasoning small models, reporting the opposite symptom — genuine contradictions *missed*. Both failures point at the same place, and a structural check helps in a way a better prompt cannot: it bounds what a wrong answer can destroy.

The `invalid_at` stamped on the retired edge is the new edge's `valid_at`, which usually derives from the ingesting episode's reference time. That's why collateral retirements are dated to the day something was *saved* rather than the day anything changed — in one ingest, eight separate edges all shared the same `valid_at`, to the second.

## Proposed fix

Restoring that earlier scoping is not enough on its own: "touches either endpoint" still admits all three collateral cases, because each killing edge shares a person or company with its victim.

An edge should only be retired by an edge that could plausibly *replace* it. Two shapes qualify:

- **Same two entities**, in either direction. This is a restatement of the same relationship, so a relabeled relation still supersedes correctly — "service `USES` datastore" giving way to "service `MIGRATED_OFF` datastore".
- **One shared entity and the same relation name.** This is what a rename or a port looks like — the `PART_OF` case above, where `helper.rb` legitimately replaces `helper.py` under the same parent.

```python
def _relation_key(name: str) -> str:
    """Fold case and separators: `MIGRATED_OFF`, `migrated-off` and `Migrated Off` all match."""
    return re.sub(r'[\s_-]+', '', name or '').casefold()


def _could_replace(candidate: EntityEdge, resolved_edge: EntityEdge) -> bool:
    """True when `resolved_edge` is plausibly a replacement for `candidate`.

    Either it connects the same two entities, and so can restate that relationship; or it shares
    one endpoint and carries the same relation name, which is what a rename or a port looks like.
    Without this, any semantically similar edge in the group is eligible, and a new fact about a
    person's side project can retire the fact recording their job.
    """
    candidate_pair = {candidate.source_node_uuid, candidate.target_node_uuid}
    resolved_pair = {resolved_edge.source_node_uuid, resolved_edge.target_node_uuid}

    if candidate_pair == resolved_pair:
        return True

    return bool(candidate_pair & resolved_pair) and _relation_key(
        candidate.name
    ) == _relation_key(resolved_edge.name)
```

Apply it once where the candidate list is assembled (~line 777, after the LLM's `contradicted_facts` are mapped back to edges):

```python
invalidation_candidates = [
    candidate for candidate in invalidation_candidates
    if _could_replace(candidate, resolved_edge)
]
```

Filtering here rather than inside `resolve_edge_contradictions` covers both symptoms — the retire-the-old-edge path and the born-retired-new-edge path share this list.

Against the four audited cases: the three wrong retirements stop and the correct one still fires.

Comparing entity pairs unordered is deliberate. Extraction direction is not stable — "a person works at a company" can come out as `person -WORKS_AT-> company` or `company -EMPLOYS-> person` — and a genuine supersession whose direction flipped between ingests should still invalidate.

### Trade-offs

- **A relation renamed *and* re-pointed at the same time will no longer invalidate.** Both clauses fail, and the old fact lingers as stale. That's the acceptable direction of error: a stale fact that lingers is visible and correctable, while a true fact silently retired is neither.
- **Relation-name equality carries real weight now**, which makes it sensitive to how consistently the extractor labels relations. Names are not normalized in the graph above (`Uses` and `USES` coexist, likewise `Affects`/`AFFECTS`), which is why the comparison folds case and separators rather than comparing raw strings.

### Considered: just restoring the earlier helper

Cheaper, and it would cut LLM cost by shrinking the pool. But measured against the audited cases it prevents none of the collateral, so it is not a fix on its own. Worth doing for cost and candidate quality; not a substitute.

There is also nothing to restore it *with*. `SearchFilters` has no way to express "edges touching these nodes," which is why #906 dropped the scoping rather than porting it — the duplicate path's scoping survived only because `edge_uuids` was added in the same PR. The driver layer can already do it (`edge_similarity_search` accepts `source_node_uuid` / `target_node_uuid`), but `search()` exposes no way to reach it. A `node_uuids` filter would close that gap for every backend and let the dropped `min_score=0.2` come back with it. Better as separate work: it touches each provider's filter constructor, and the guard above is what actually prevents the collateral.

## Tests

- Unit-test the filter through `resolve_extracted_edge` with a mocked LLM returning `contradicted_facts`, following the existing pattern in `tests/utils/maintenance/test_edge_operations.py`. Cases: same pair with an earlier `valid_at` → invalidated; same pair with the direction flipped → invalidated; one shared endpoint and the same relation name → invalidated; one shared endpoint and a different relation name → **not** invalidated; no shared endpoint → **not** invalidated.
- Cover the born-retired path: a candidate with a later `valid_at` that fails `_could_replace` must not set `invalid_at` on the new edge.
- Cover name folding: `USES` and `Uses` must be treated as the same relation.
- Run the existing suite — this is on the main ingest path. #1603 adds unit tests for `resolve_edge_contradictions` itself; this change leaves that function untouched, so the two should not conflict.

## Out of scope

- **Repairing existing damage.** This change stops further collateral but does not undo any that has already happened. Facts wrongly retired before the fix stay retired, and any deployment that has been ingesting for a while will have some. Identifying and reviving them means deciding, per edge, whether its `invalid_at` reflects real change or a bad match — separate work, with its own risk of clearing dates that were correct.
- **Search-side handling of invalidated facts.** `search_memory_facts` returns retired facts interleaved with live ones — across 195 recorded searches in the graph above, 30% of returned facts carried an `invalid_at`. Already proposed separately in #1645.

brentkearney:
## Summary
Edge invalidation currently accepts any semantically similar edge in the group as a candidate, and whatever the LLM flags as contradicted is acted on unchecked. The result is that saving a memory that merely *mentions* another entity can invalidate an unrelated, still-true fact about that entity — silently, since the retired fact just gains an `invalid_at` and drops out of search results. Details in #1728.

This PR adds a structural guard where the invalidation candidate list is assembled in `resolve_extracted_edge`: a candidate is kept only if the new edge could plausibly *replace* it, meaning either

- it connects the **same two entities** (in either direction — extraction direction is not stable), so it can restate that relationship; or
- it shares **one endpoint in the same position** (same source, or same target) and carries the same relation name (case- and separator-folded), which is what a rename or a re-point looks like. A shared endpoint in *opposite* positions is not a replacement, it is two unrelated facts chained through one node: `A -uses-> B` is not superseded by `B -uses-> C`.

Filtering at this point covers both downstream paths that act on the LLM's contradiction call: retiring the old edge (`resolve_edge_contradictions`) and back-dating the new edge (`invalid_at` set at creation when a flagged candidate is newer).

The guard bounds what a wrong LLM judgement can destroy — the judge runs on bare fact strings with a small model, and #1666 documents how unreliable that call is. The trade-off runs in the safe direction: a relation renamed *and* re-pointed in one step will no longer invalidate, leaving a stale-but-visible fact, instead of a true fact being silently retired.

### Known limitation

The guard does not decide the remaining case. When two edges share an endpoint in the same position and carry the same relation name, whether one supersedes the other depends on the relation's cardinality, which the graph does not record. `Dana WORKS_AT Northwind` superseded by `Dana WORKS_AT Contoso` is correct; `ServiceA USES Postgres` retired by `ServiceB USES Postgres` is not, and the guard admits both.

Restricting the rule by direction does not help, because hub entities occur on both sides. In one real graph a single entity was the target of the same relation from 8 distinct sources, and a single source was the origin of the same relation to 25 distinct targets. Whichever side you restrict, you break a legitimate supersession shape and leave the other exposed. That judgement needs cardinality and belongs with the judge. What this PR does is bound the blast radius to candidates that at least share an anchor and a relation name, instead of any semantically similar edge in the group.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
See #1728 for the full analysis, including why the candidate search lost its endpoint scoping in #906 and a four-case audit of collateral retirements. Restoring the pre-#906 scoping alone would not fix this — "touches either endpoint" still admits all three wrong cases in the audit — so this guard is the fix, and re-scoping the candidate search remains worthwhile follow-up for cost and candidate quality.

## Testing
- [x] All existing tests pass

Eight new unit tests in `tests/utils/maintenance/test_edge_operations.py` drive `resolve_extracted_edge` with a mocked LLM flagging a contradiction, per the plan in #1728: same pair (and same pair with direction flipped) still invalidates; shared endpoint with the same relation name still invalidates (including case/separator folding, `Uses` vs `USES`); shared endpoint with a different relation name, shared endpoint in the opposite position, and disjoint endpoints no longer invalidate; and a later-dated candidate that fails the guard no longer leaves the new edge born retired.

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1728


anujbolewar (commented):
Careful logic — same-pair and same-position rules prevent a new fact about one node from retiring an unrelated relationship while still catching renames and re-points. One question: in the exact same-pair branch the relation name is not compared, so an unrelated relation type between the same two nodes could still supersede the original. If that breadth is intentional, keep it; otherwise it may be worth requiring a compatible folded relation key there too.

brentkearney:
Intentional — and a test pins it. `test_invalidation_same_pair_invalidates` retires `USES` with `MIGRATED_OFF` between the same pair. Requiring a matching folded key there would break that case, which is the most common shape of real change: the relationship persists and the label for it moves.

Your concern is real, though, so here are numbers rather than an assurance.

I re-measured the graph from #1728 — an earlier snapshot than the one quoted there, 2,862 facts with 1,100 carrying an `invalid_at`. For each retired edge I looked for the edge that retired it, matching the retiring edge's `valid_at` to the victim's `invalid_at`. That is a proxy, not provenance: one ingest can stamp several edges with the same `valid_at` to the second, so each row below is an upper bound on "had a retiring edge of this shape". 1,082 of the 1,100 matched at least one.

| Shape of the retiring edge | Count |
|---|---|
| Same pair | 85 |
| — of those, different relation name | 70 |
| One shared endpoint, same position, same name | 69 |
| Neither, so blocked by this PR | 928 |

The guard's value is almost all in the last row: 86% of retirements came from edges that could not plausibly replace their victim. The same-pair branch you're asking about accounts for 8%.

Grouping those 70 same-pair renames by relation name, they split roughly evenly between two kinds:

- Genuine supersession, which a key check would break: `CONTAINS_FILE` → `IS_PART_OF`, `BUMPS_VERSION_OF` → `UPGRADES`, `CONSIDERING_BUNDLING` → `PLANS_TO_PACKAGE`.
- Additive facts that should have retired nothing: `USES` → `RENDERS`, `IMPLEMENTS` → `USES`, `PUBLISHED_TO` → `USES`.

Both kinds are same-pair with a different name, so no rule over endpoints and names separates them. Requiring a compatible key stops the second group and takes the first with it, and gives up recording a relabeled relationship at all. That trade belongs at the judge, not in a guard whose job is to bound what a single bad judgement can destroy.

What separates the two kinds is entity context — the second cause in #1728. The judge sees bare fact strings and cannot tell which entities each fact connects. #1784 adds a reasoning field, which gives the model room to think, but still does not show it the endpoints. So same-pair-with-a-different-name is this PR's known residual gap, and I'd rather state it than close it with a rule that costs more than it saves.

Happy to record that limit in the `_could_replace` docstring so the next reader finds it stated. If a maintainer prefers the stricter key check, it's a two-line change and the tests above say exactly which cases flip.

PR #1732, opened 2026-08-05
fix(embedder): pair batch embeddings by response index, not position

Sicatho:
## Summary

`create_batch` in both `OpenAIEmbedder` and `AzureOpenAIEmbedderClient` reads `response.data` **positionally**, but the OpenAI embeddings API documents that list as *index-addressed* and does not guarantee it arrives in input order. When a response comes back out of order, each embedding is attached to the wrong input.

```python
# graphiti_core/embedder/openai.py, before
return [embedding.embedding[: self.config.embedding_dim] for embedding in result.data]
```

This PR sorts by `index` before extracting, in both clients, and adds tests that fail without the fix.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Documentation/Tests

## Objective

**The failure is silent.** Every vector is well-formed and the batch size is correct, so nothing errors and nothing looks wrong — the embeddings simply describe different text than the nodes they are stored on. Downstream, a search returns the wrong node for a query, with no signal anywhere that it happened.

It is also invisible to the current tests. `test_create_batch_processes_multiple_inputs` asserts:

```python
assert result == [
    mock_openai_batch_response.data[0].embedding[:dim],
    mock_openai_batch_response.data[1].embedding[:dim],
    mock_openai_batch_response.data[2].embedding[:dim],
]
```

That is positional on both sides, so it encodes the same assumption the code makes — and the mock embeddings carry no `index` field at all, so the fixture cannot express an out-of-order response.

## Testing
- [x] All existing tests pass

- `create_openai_embedding()` now accepts an `index` (default `0`) so fixtures can express real responses. Existing fixtures pass `index=0,1,2`; their assertions are unchanged.
- **`test_create_batch_pairs_embeddings_by_index`** (new, one per client): a valid response whose `data` is shuffled while carrying correct `.index` values must still return embeddings in **input** order.
- **`tests/embedder/test_azure_openai.py`** (new file): the Azure client had no tests, and this PR changes it — covers single create, batch, and index pairing.

Verified the new tests are not vacuous: with the source change reverted, exactly the two `test_create_batch_pairs_embeddings_by_index` tests fail and the rest pass; with it applied, all 6 pass.

```
tests/embedder/test_openai.py tests/embedder/test_azure_openai.py -> 6 passed
ruff format: clean · ruff check: All checks passed
```

## Breaking Changes

Behaviour only changes when a provider returns `data` out of order, where the previous result was incorrect.

## Checklist
- [x] Code follows project style guidelines (`ruff format` / `ruff check` pass)
- [x] Self-review completed
- [ ] Documentation updated where necessary — no user-facing docs affected

## Related Issues

None found — I searched open issues and PRs for this before filing.

## Deliberately not in this PR

Two adjacent observations left alone to keep this reviewable. Happy to open issues or a follow-up PR if you'd like them:

1. `OpenAIEmbedder` truncates to `config.embedding_dim` while `AzureOpenAIEmbedderClient` does not, so the two clients can return different widths for the same configured model.
2. That truncation is silent — a model returning an unexpected dimension is reshaped rather than reported.

### DIFF
--- a/graphiti_core/embedder/azure_openai.py
+++ b/graphiti_core/embedder/azure_openai.py
@@ -65,7 +65,10 @@ class AzureOpenAIEmbedderClient(EmbedderClient):
                 model=self.model, input=input_data_list
             )
 
-            return [embedding.embedding for embedding in response.data]
+            # Pair by `index`, not position: the API does not guarantee that
+            # `data` arrives in input order (see OpenAIEmbedder.create_batch).
+            ordered = sorted(response.data, key=lambda embedding: embedding.index)
+            return [embedding.embedding for embedding in ordered]
         except Exception as e:
             logger.error(f'Error in Azure OpenAI batch embedding: {e}')
             raise
--- a/graphiti_core/embedder/openai.py
+++ b/graphiti_core/embedder/openai.py
@@ -63,4 +63,8 @@ class OpenAIEmbedder(EmbedderClient):
         result = await self.client.embeddings.create(
             input=input_data_list, model=self.config.embedding_model
         )
-        return [embedding.embedding[: self.config.embedding_dim] for embedding in result.data]
+        # The API documents `data` as index-addressed and does not guarantee it
+        # arrives in input order, so pair by `index` rather than by position:
+        # reading positionally attaches embeddings to the wrong inputs silently.
+        ordered = sorted(result.data, key=lambda embedding: embedding.index)
+        return [embedding.embedding[: self.config.embedding_dim] for embedding in ordered]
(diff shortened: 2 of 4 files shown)


Sicatho:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: work.usmansiddiqui@gmail.com

Sicatho:
@
Closing this. I checked the premise more carefully after opening it: OpenAI states batched embedding responses preserve input order, and openai-python removed this exact `sorted(..., key=index)` step as unnecessary in openai/openai-python#430. The current Cookbook still consumes `response.data` positionally. I do not have a reproducer from any real provider showing reordered embedding data, so this fixes a failure mode I cannot demonstrate — and naive sorting would break responses where `index` is absent, which the SDK materializes as `None`.

The Azure embedder client does lack test coverage entirely, so I will open a separate PR adding tests for it. Sorry for the noise.
@

PR #1733, opened 2026-08-05
test(embedder): add coverage for the Azure OpenAI embedder client

Sicatho:
@
`AzureOpenAIEmbedderClient` has no test module. The OpenAI, Gemini and Voyage clients each have one under `tests/embedder/`, so this adds the missing file.

**Tests only — no source changes.** Everything asserted is the client's current behaviour.

What it pins, chosen for things that break silently rather than loudly:

- `create` wraps a bare string into a one-element list, passes a list of strings through unchanged, and stringifies any other input
- `create` returns the first embedding; `create_batch` returns one embedding per input, in order
- the configured deployment name reaches the API on both code paths
- vectors come back at full length — unlike `OpenAIEmbedder` this client has no `embedding_dim` and does not truncate
- API errors are logged and re-raised, not swallowed into an empty or malformed result

The client is injected through the constructor, so the test passes a mock directly instead of patching a module-level symbol.

**Not vacuous:** I mutated `azure_openai.py` six ways — dropped the list-wrap, returned the last embedding instead of the first, reversed the batch, truncated the vector, swallowed batch errors, ignored the `model` argument — and each mutation fails at least one of these tests. All 9 pass against unmodified source.

`ruff check` and `ruff format --check` are clean. `tests/embedder/test_gemini.py` and `test_voyage.py` do not collect in my environment because their optional extras are not installed; that is unrelated to this change.
@

### DIFF
--- a/tests/embedder/test_azure_openai.py
+++ b/tests/embedder/test_azure_openai.py
@@ -1,0 +1,194 @@
+"""
+Copyright 2024, Zep Software, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


Sicatho:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: work.usmansiddiqui@gmail.com

Sicatho:
recheck

PR #1743, opened 2026-08-09
fix(server): include fact provenance in /search FactResult (#1735)

### LINKED ISSUE #1735: REST /search response omits fact provenance (source/target nodes, source episodes)
## Problem

The REST server's `/search` response (`FactResult` in `server/graph_service/dto/retrieve.py`) returns each fact as a bare sentence plus timestamps:

```python
class FactResult(BaseModel):
    uuid: str
    name: str
    fact: str
    valid_at: datetime | None
    invalid_at: datetime | None
    created_at: datetime
    expired_at: datetime | None
```

The underlying `EntityEdge` the fact comes from carries `source_node_uuid`, `target_node_uuid`, and `episodes` (the episode uuids the fact was extracted from), but none of these survive into the DTO.

## Why it matters

Consumers that need to connect recalled facts back to their own domain objects have no structural handle:

- Without **source/target node names**, mapping a fact to the entities it connects requires parsing the fact sentence text, which is lossy and fragile.
- Without **source episode references**, systems that encode identity in episode names (e.g. one episode per source document) cannot trace a fact back to the document it was learned from, even though the graph stores that linkage.

In practice this makes `/search` results useful for display but hard to integrate into ranking or citation pipelines, and the information is already in the graph.

## Proposal

Additive, backward-compatible fields on `FactResult`:

```python
    source_node_uuid: str | None = None
    source_node_name: str | None = None
    target_node_uuid: str | None = None
    target_node_name: str | None = None
    episodes: list[str] | None = None  # source episode uuids
```

Node names may need one extra lookup per result set (uuids alone would already help); `episodes` is a direct pass-through from `EntityEdge`.

This is the same class of gap as the score field addressed in #758: data present at the library layer, dropped at the REST boundary.

Happy to open a PR if the direction is acceptable.

Anai-Guo:
## Problem

The REST server's `/search` response (`FactResult` in `server/graph_service/dto/retrieve.py`) returns each fact as a bare sentence plus timestamps. The underlying `EntityEdge` carries `source_node_uuid`, `target_node_uuid`, and `episodes` (the episode uuids the fact was extracted from), but `get_fact_result_from_edge` never maps them into the DTO, so none survive into the response.

Consumers that need to connect recalled facts back to their own domain objects are left with no structural handle:

- Without **source/target node uuids**, mapping a fact to the entities it connects requires parsing the fact sentence text, which is lossy and fragile.
- Without **source episode references**, systems that encode identity in episode names (e.g. one episode per source document) cannot trace a fact back to the document it was learned from, even though the graph already stores that linkage.

## Change

- Add `source_node_uuid`, `target_node_uuid`, and `episodes` to `FactResult`.
- Populate them in `get_fact_result_from_edge` from the fields the `EntityEdge` already exposes.

This is purely **additive** to the response schema — existing consumers keep receiving every field they had before, and the new fields come straight from data the edge already holds (no extra queries). `episodes` defaults to an empty list to stay robust for edges with no recorded provenance.

Fixes #1735


mehulp93:
Thank you very much for contributing to the Graphiti project. This PR has been superseded by https://github.com/getzep/graphiti/pull/1750 - closing this one out, but appreciate the work you put in here.

PR #1761, opened 2026-08-13
fix: round-trip first/last_episode_uuid when re-fetching an existing saga (#1755)

### LINKED ISSUE #1755: [Bug] `_get_or_create_saga` re-fetch drops `first_episode_uuid`/`last_episode_uuid`, causing silent overwrite on the 2nd+ episode
# [Bug] `_get_or_create_saga` re-fetch drops `first_episode_uuid`/`last_episode_uuid`, causing silent overwrite on the 2nd+ episode

## Summary

`Graphiti._get_or_create_saga()`'s re-fetch query for an *existing* saga only
selects `uuid, name, group_id, created_at`. It omits `first_episode_uuid` and
`last_episode_uuid`, so the `SagaNode` it returns always has those two fields
as `None` — even when the database already has correct values. Every caller
of `add_episode(..., saga=...)` then runs:

```python
if saga_node.first_episode_uuid is None:
    saga_node.first_episode_uuid = primary_episode.uuid
saga_node.last_episode_uuid = primary_episode.uuid
await saga_node.save(self.driver)
```

Because the re-fetched node's `first_episode_uuid` is always `None`, this
guard fires on *every* episode added to the saga, not just the first — so
`first_episode_uuid` is silently overwritten to match the most recently added
episode on every call. The persisted value in the database is corrupted, not
just an in-memory artifact.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Objective

Fix the re-fetch query in `_get_or_create_saga` (`graphiti_core/graphiti.py`,
~lines 371-379) to select `first_episode_uuid` and `last_episode_uuid` (and
construct the returned `SagaNode` with them), so existing sagas round-trip
correctly and the first/last-episode tracking in the caller behaves as
documented.

## What I was trying to do

Add a second episode to an existing saga via `add_episode(..., saga="my-saga")`
and then read the saga back to confirm `first_episode_uuid` still pointed at
the first episode.

## What I expected

`saga.first_episode_uuid` to remain the UUID of the *first* episode added to
the saga, and `saga.last_episode_uuid` to track the most recent one.

## What actually happened

`saga.first_episode_uuid` was overwritten to the UUID of the *second* episode
— identical to `last_episode_uuid`. The original first-episode UUID is lost.

## Reproduction

**Why this calls `_get_or_create_saga` directly.** The bug is reached in
normal use through `add_episode(..., saga=...)`, but that path requires a
configured LLM client and a full extraction pass, which would make this
reproduction slow, non-deterministic and dependent on an API key. The script
below instead exercises the same two lines of library code that
`add_episode` runs, so the failure is isolated and deterministic.

The guard it reproduces is copied verbatim from the real caller in
`graphiti_core/graphiti.py` (~lines 775-779, with an equivalent in the bulk
path at ~1456):

```python
if saga_node.first_episode_uuid is None:
    saga_node.first_episode_uuid = primary_episode.uuid
saga_node.last_episode_uuid = primary_episode.uuid
await saga_node.save(self.driver)
```

The placeholder episode UUIDs are deliberate — the defect is in the saga
re-fetch, and the episode values only need to be distinguishable from each
other. Substituting real episode UUIDs changes nothing about the outcome.

Minimal script (run against a throwaway Neo4j instance; requires
`REPRO_NEO4J_URI`, `REPRO_NEO4J_USER`, `REPRO_NEO4J_PASSWORD` env vars):

```python
import asyncio
import os

from graphiti_core.nodes import SagaNode

uri = os.environ['REPRO_NEO4J_URI']
user = os.environ['REPRO_NEO4J_USER']
password = os.environ['REPRO_NEO4J_PASSWORD']


async def main():
    g = Graphiti(uri=uri, user=user, password=password)
    group_id = 'repro-saga-group'
    now = datetime.now(timezone.utc)

    saga = await g._get_or_create_saga('repro-saga', group_id, now)
    ep1_uuid = 'episode-1-uuid'
    if saga.first_episode_uuid is None:
        saga.first_episode_uuid = ep1_uuid
    saga.last_episode_uuid = ep1_uuid
    await saga.save(g.driver)

    saga = await g._get_or_create_saga('repro-saga', group_id, now)  # re-fetch
    ep2_uuid = 'episode-2-uuid'
    if saga.first_episode_uuid is None:  # BUG: True again, should be False
        saga.first_episode_uuid = ep2_uuid
    saga.last_episode_uuid = ep2_uuid
    await saga.save(g.driver)

    [readback] = await SagaNode.get_by_group_ids(g.driver, [group_id])
    print('episode 1 uuid:      ', ep1_uuid)
    print('episode 2 uuid:      ', ep2_uuid)
    print('saga.first_episode_uuid:', readback.first_episode_uuid)
    print('saga.last_episode_uuid: ', readback.last_episode_uuid)
    assert readback.first_episode_uuid == ep1_uuid, 'CORRUPTED: first_episode_uuid was overwritten'

    await g.close()


asyncio.run(main())
```

Actual output against `main` (commit at time of writing), throwaway Neo4j
5.26-community:

```
episode 1 uuid:       episode-1-uuid
episode 2 uuid:       episode-2-uuid
saga.first_episode_uuid: episode-2-uuid
saga.last_episode_uuid:  episode-2-uuid
AssertionError: CORRUPTED: first_episode_uuid was overwritten
```

## Testing

Repro above demonstrates the failure directly against the database; no mocks
involved. A fix should add `first_episode_uuid`/`last_episode_uuid` to the
re-fetch `RETURN` clause and to the `SagaNode(...)` construction on the
existing-saga branch, then re-run this repro to confirm
`first_episode_uuid` remains `episode-1-uuid` after the second call.

## Breaking Changes

None. This only corrects data that the current code silently corrupts; no
public API changes.

## Checklist

- [x] Reproduced against `main`, not just a released version
- [x] Reproduction is self-contained and requires only graphiti + a database
- [x] Root cause identified with file/line reference
- [ ] Fix implemented (not in scope for this issue — filed to establish the
      record before a PR)

## Related Issues

None found covering this defect in a targeted search of open/closed issues.

icn5381:
## Summary

Fix `Graphiti._get_or_create_saga()`'s re-fetch query so an existing saga round-trips its `first_episode_uuid` / `last_episode_uuid`. The query previously selected only `uuid, name, group_id, created_at`, so the returned `SagaNode` always carried both episode fields as `None` — even when the database had correct values. Every `add_episode(..., saga=...)` call then re-fired the first-episode guard and silently overwrote `first_episode_uuid` with the most recently added episode, corrupting the persisted value.

## Type of Change
- [x] Bug fix

## Objective

`add_episode`'s guard `if saga_node.first_episode_uuid is None` is only correct if the re-fetched node reflects the database. Restore that invariant by selecting and hydrating the two columns.

## Testing
- [x] Unit tests added/updated — new `tests/test_saga_refetch_episode_uuids.py`:
  - re-fetch of an existing saga preserves `first_episode_uuid` / `last_episode_uuid` and performs no write (red before the fix)
  - no matching row still mints a fresh node with both fields `None`
- [x] All existing tests pass — target file 2 passed; `make lint` passes

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1755


icn5381:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 255778606+icn5381@users.noreply.github.com

icn5381:

icn5381:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: icn5381@users.noreply.github.com

Naseem77:
@icn5381 Verified live on FalkorDB: re-fetching an existing Saga returns `None` for both episode uuids on main and the stored values with your patch. Test fails on main and passes after. Minimal and correct, LGTM!

icn5381:
Thanks for the live verification on FalkorDB! The branch is rebased onto current main (`45b4a86`); the regression test still passes (2/2).

prasmussen15 (approved):

prasmussen15:
Landed onto current main (including #1699) as https://github.com/getzep/graphiti/pull/1854. The added unit test used a `Z` suffix that `datetime.fromisoformat` rejects on Python 3.10; the landing PR uses `+00:00`.

prasmussen15:
The original fork still cannot merge into current `main` because the unit test uses a `Z` datetime suffix that Python 3.10 rejects. Combined landing PR: https://github.com/getzep/graphiti/pull/1856

prasmussen15:
Landed on \`main\` via https://github.com/getzep/graphiti/pull/1856 (\`3ff5c16\`). You are credited as a co-author on that commit. Closing this fork PR to avoid a duplicate merge.

PR #1255, opened 2026-02-21
docs: Dual-Brain Architecture and Operators Guide

yhl999:
Reframes memory system as a Dual-Brain architecture (Neo4j semantic + SQLite ledger logic) per founder direction.

PR #1762, opened 2026-08-13
add entity rule for machine-generated run ids and log scaffolding

Soju06:
## Problem

Session-transcript and tool-output episodes carry throwaway execution identifiers — process ids, temp script paths, `key=value` counters, bare log tokens. `ENTITY RULES` had no clause covering them, so they were extracted as first-class entities.

Measured on a real 11KB compaction episode (gpt-5.5): **9.5 junk nodes per episode**, e.g. `/tmp/bori-hub-test.log`, `/tmp/rvw/rvw-20260804-083837-pr-1255/report.md`, `/tmp/p34-watch2.sh`, `proc_*` ids.

Two costs: they are never searched for later, and they inflate the node set that `resolve_extracted_edges` scans on every subsequent write.

## Change

One new clause, `ENTITY RULE 6i`, alongside the existing 6a–6h banned-class list. The load-bearing part is the redirection, not the ban: keep the identifier inside fact text when it carries meaning (`"the hub-test run exited rc=1"`) and name the durable subject as the node (`"hub-test suite"`, `"PR #1255"`).

## Measurements

Real 11KB compaction episode, gpt-5.5, `reasoning_effort=none`, 2 runs per variant:

| variant | entities | junk nodes | gold-fact coverage | p50 |
|---|---:|---:|---:|---:|
| without 6i | 51 | 9.5 | 0.786 | 84s |
| **with 6i** | 57 | **0** | **0.929** | 121s |

Junk goes to zero. Fact coverage *improves* (0.786 → 0.929) rather than regressing — the redirection clause moves detail into fact text instead of dropping it, and the node budget shifts from run ids to real entities.

Latency rises 84s → 121s. That tracks the larger extracted fact set, not added prompt overhead (~700 chars).

Gold-fact coverage is scored against hand-labelled items from the source episode that a user would plausibly search for later (security blockers, run outcomes, count/status changes).

Cross-checked on `gpt-5.6-sol`: junk was already 0 in both variants, coverage 0.429 → 0.5. The rule does not hurt models that were already clean.

## Test plan

- `uv run ruff check` / `ruff format --check` on the changed file — clean
- CI unit gate (`DISABLE_NEO4J/FALKORDB/KUZU/NEPTUNE`, workflow ignore list): **373 passed, 11 skipped**
- `tests/embedder/test_voyage.py` fails to collect on `main` as well (pre-existing, unrelated — verified by stashing this change)

Prompt-only change: no schema, API, or driver impact.


Soju06:
Opened against the wrong remote by mistake (this targets our downstream fork, not upstream). Closing — sorry for the noise.

PR #1764, opened 2026-08-13
fix(attributes): preserve prior node attributes when no entity type applies

### LINKED ISSUE #1763: [BUG] Node attributes are silently cleared when no entity type applies to the node
## Bug Description
`extract_attributes_from_nodes` assigns `node.attributes = attributes` unconditionally
(`graphiti_core/utils/maintenance/node_operations.py:763-764`), and `_extract_entity_attributes`
returns a bare `{}` whenever no entity type applies to the node (`:790-791`):

```python
if entity_type is None or len(entity_type.model_fields) == 0:
    return {}
```

For a node that deduplicates onto an existing stored node, that empty dict replaces attributes a
previous typed pass had extracted. The loss reaches storage:

- The resolved node **is** the database node — `_promote_resolved_node` returns the existing
  candidate (`dedup_helpers.py:177/181`), and `get_entity_node_from_record` populates `attributes`
  from the record (`nodes.py:1052-1054`).
- `get_entity_node_save_query` replaces attributes wholesale — `SET n = $entity_data` for
  Neo4j/FalkorDB/Neptune, `n.attributes = $attributes` for Kuzu (`node_db_queries.py:137`).

Provider-agnostic; not specific to any backend.

## Steps to Reproduce
Two paths reach the early return. Either is enough:

```python

class Person(BaseModel):
    age: int | None = None
    city: str | None = None

# 1. A node's attributes are populated by a typed pass.
await graphiti.add_episode(..., entity_types={'Person': Person})

# 2a. A later episode omits entity_types entirely — the same node dedups and is wiped.
await graphiti.add_episode(..., entity_types=None)

# 2b. Or entity_types is supplied but lacks the node's label — `entity_types.get(...)`
#     resolves to None and takes the same path.
await graphiti.add_episode(..., entity_types={'Organization': Organization})
```

Note `next((item for item in node.labels if item != 'Entity'), '')` (`:752`) yields `''` for a node
whose only label is `Entity`, so that lookup takes the same path even when `entity_types` is passed.

Reproducible as a unit test with mocked clients and no database — I have two such tests ready in
`tests/utils/maintenance/test_node_operations.py`, which fail on `main` with
`assert {} == {'age': 30, 'city': 'New York'}`.

## Expected Behavior
Prior attributes are preserved. An empty extraction result means "nothing applicable to extract,"
not "the node has no attributes."

This is the invariant `add_triplet` already enforces (`graphiti.py:1676`):

```python
# Update attributes dictionary (merge rather than replace)
if source_node.attributes:
    resolved_source.attributes.update(source_node.attributes)
```

covered by `test_add_triplet_empty_attributes_preserved` (`tests/test_add_triplet.py:433`) —
*"Test that nodes with empty attributes don't overwrite existing attributes."*

## Actual Behavior

The stored node's `attributes` are replaced with `{}`, silently. No error, no warning.

## Environment
- **Graphiti Version**: `main` (`pyproject.toml` version 0.29.3)
- **Python Version**: 3.12
- **Operating System**: macOS 15
- **Database Backend**: reproduced provider-agnostically with mocked clients; originally observed on
  Amazon Neptune
- **LLM Provider & Model**: Google gemini-3.1-flash-lite (not relevant — the wipe happens on a path
  that makes no LLM call)

## Installation Method
- [ ] pip install
- [ ] uv add

## Error Messages/Traceback
```
None — the failure is silent. Surfaced by asserting on attributes after a second ingestion:
AssertionError: assert {} == {'age': 30, 'city': 'New York'}
```

## Configuration
```python
# No special configuration required. Triggered by any two ingestions where the second
# does not have the node's entity type in scope.
```

## Additional Context
- **Consistently**, not intermittently — it is a deterministic code path, not a race.
- Component: core library (`graphiti_core`).
- Why it looks like an oversight rather than intent: the typed path deliberately preserves prior
  values through `apply_capped_attributes(..., merge_mode='overlay')` → `{**prior_attributes, **kept}`
  (`attribute_utils.py:245-246`). Only the no-applicable-type early return bypasses it. The comment
  at the call site already asserts the invariant the code then breaks: *"`_extract_entity_attributes`
  returns the already-merged attribute dict (overlay of prior + cap-kept fields), so direct
  assignment is the merge."* That holds only on the typed path.
- Both the unconditional assignment and `apply_capped_attributes` arrived in the same commit,
  `7514b44` (#1498, 2026-05-20), whose stated scope includes attribute-hallucination guards.
- **Possibly why it went unnoticed:** `test_add_triplet_empty_attributes_preserved` encodes exactly
  this invariant, but it depends on the `graph_driver` fixture, which parametrizes over `drivers` —
  and the default unit-test CI job sets `DISABLE_NEO4J`, `DISABLE_FALKORDB`, `DISABLE_KUZU` and
  `DISABLE_NEPTUNE`, leaving that list empty. So the test that would have caught a regression here
  does not run in the job that gates PRs.

## Possible Solution
Return the node's prior attributes from the early return instead of `{}`. That makes the call site's
documented contract true on both paths, so no guard is needed there:

```python
if entity_type is None or len(entity_type.model_fields) == 0:
    return dict(node.attributes or {})
```

`len(model_fields) == 0` gets the same preservation. The edge attribute path is intentionally
`merge_mode='replace'` and would be untouched.

I have this fix with two regression tests ready (full unit suite passes, `make lint` clean) and will
open a PR referencing this issue.

yatharth-anand-2410:
## Summary

  `extract_attributes_from_nodes` assigns `node.attributes = attributes` unconditionally
  (`node_operations.py:763-764`), and `_extract_entity_attributes` returned a bare `{}` whenever no
  entity type applied to the node (`:790-791`). For a node that deduplicates onto an existing stored
  node, that empty dict silently cleared attributes a previous typed pass had extracted.

  The one-line fix returns the node's prior attributes from that early return instead of `{}`.

  ## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Performance improvement
  - [ ] Documentation/Tests

  ## Objective

  Not a new feature — restoring an invariant the codebase already asserts in three places:

  1. The typed path deliberately preserves prior values via
     `apply_capped_attributes(..., merge_mode='overlay')` → `{**prior_attributes, **kept}`
     (`attribute_utils.py:245-246`). Only the no-applicable-type early return bypassed it.
  2. The comment at the call site already claims the invariant the code then broke:
     *"`_extract_entity_attributes` returns the already-merged attribute dict (overlay of prior +
     cap-kept fields), so direct assignment is the merge."* That held only on the typed path.
  3. `add_triplet` enforces the same rule explicitly (`graphiti.py:1676`):
     ```python
     # Update attributes dictionary (merge rather than replace)
     if source_node.attributes:
         resolved_source.attributes.update(source_node.attributes)
     ```
     covered by `test_add_triplet_empty_attributes_preserved` (`tests/test_add_triplet.py:433`) —
     *"Test that nodes with empty attributes don't overwrite existing attributes."*

  Returning prior attributes makes the call site's documented contract true on **both** paths, so no
  guard is needed there and the two behaviours converge.

  ## Root cause

  `_extract_entity_attributes` conflates "no applicable type, nothing to extract" with "extracted
  nothing":

  ```python
  if entity_type is None or len(entity_type.model_fields) == 0:
      return {}
  ```

  The loss reaches storage because the deduplicated node is the database node:

  - `_promote_resolved_node` returns the existing candidate (`dedup_helpers.py:177/181`), and
    `get_entity_node_from_record` populates `attributes` from the record (`nodes.py:1052-1054`).
  - `get_entity_node_save_query` replaces attributes wholesale — `SET n = $entity_data` for
    Neo4j/FalkorDB/Neptune, `n.attributes = $attributes` for Kuzu (`node_db_queries.py:137`).

  Two paths reach the early return:

  1. `entity_types=None` on a later episode, for a node whose attributes were written by an earlier
     call that *did* pass `entity_types`.
  2. `entity_types` supplied but missing the node's label — `entity_types.get(...)` then resolves to
     `None`. Note `next((item for item in node.labels if item != 'Entity'), '')` (`:752`) yields `''`
     for a node whose only label is `Entity`, so that lookup takes the same path.

  Provider-agnostic; not specific to any backend.

  ## Changes

  - `graphiti_core/utils/maintenance/node_operations.py`: `_extract_entity_attributes` returns

- `graphiti_core/utils/maintenance/node_operations.py`: `_extract_entity_attributes` returns
  `dict(node.attributes or {})` from the no-applicable-type early return, with a comment explaining
  why. `len(model_fields) == 0` gets the same preservation.
- `tests/utils/maintenance/test_node_operations.py`: two regression tests, one per path above.

The edge attribute path is intentionally `merge_mode='replace'` and is untouched. No call sites
needed to change — the fix lives entirely in the extraction helper.

## Testing
- [x] All existing tests pass

Both new tests fail against `main` with `assert {} == {'age': 30, 'city': 'New York'}` and pass with
the fix. Full unit suite as CI runs it (`DISABLE_NEO4J/FALKORDB/KUZU/NEPTUNE=1`,
`-m "not integration"`): **403 passed, 42 skipped**. `make lint` clean (ruff + pyright 0 errors).

The new tests reuse the existing `_make_clients()` helper, so they need no database and run in the
default unit-test job.

**A note that may be worth acting on separately:** the test that encodes this invariant,
`test_add_triplet_empty_attributes_preserved`, depends on the `graph_driver` fixture, which
parametrizes over `drivers` — and that list is empty when all four backends are disabled, as they are
in `unit_tests.yml`. So the test that would have caught this regression does not run in the job that
gates PRs. That may be why this slipped through. Happy to open a separate issue if useful.

## Breaking Changes

Behaviour change is strictly in the direction of not losing data. Any caller depending on attributes
being cleared when `entity_types` is omitted would be relying on the bug; the equivalent
already-documented behaviour on the `add_triplet` path is preservation.

## Checklist
- [x] Self-review completed
- [x] Documentation updated where necessary — comment at the fix site; no user-facing docs affected

## Related Issues

Closes #1763

### DIFF
--- a/graphiti_core/utils/maintenance/node_operations.py
+++ b/graphiti_core/utils/maintenance/node_operations.py
@@ -788,7 +788,10 @@ async def _extract_entity_attributes(
     entity_type: type[BaseModel] | None,
 ) -> dict[str, Any]:
     if entity_type is None or len(entity_type.model_fields) == 0:
-        return {}
+        # No applicable type means nothing to extract, not "extracted nothing": return the
+        # node's prior attributes so the caller's assignment leaves them untouched. Returning
+        # {} here would clear attributes a previous typed pass stored on a deduplicated node.
+        return dict(node.attributes or {})
 
     attributes_context = _build_episode_context(
         # should not include summary
--- a/tests/utils/maintenance/test_node_operations.py
+++ b/tests/utils/maintenance/test_node_operations.py
@@ -918,3 +918,71 @@ async def test_batch_summaries_calls_llm_for_long_summary():
     # LLM should have been called to condense the long summary
     llm_client.generate_response.assert_awaited_once()
     assert node.summary == 'Condensed summary'
+
+
+@pytest.mark.asyncio
+async def test_extract_attributes_preserves_prior_attributes_when_entity_types_none():
+    """Prior attributes survive when no entity types are supplied.
+
+    Mirrors the invariant `add_triplet` already enforces ("merge rather than replace",
+    see `test_add_triplet_empty_attributes_preserved`): an empty extraction result must
+    not clear attributes a previous typed pass stored on the node.
+    """
+    clients, _ = _make_clients()
+    clients.llm_client.generate_response = AsyncMock(return_value={'summaries': []})
+    clients.embedder.create = AsyncMock(return_value=[0.1, 0.2, 0.3])
+    clients.embedder.create_batch = AsyncMock(return_value=[[0.1, 0.2, 0.3]])
+
+    node = EntityNode(
+        name='Alice',
+        group_id='group',
+        labels=['Entity', 'Person'],
+        summary='Existing summary',
+        attributes={'age': 30, 'city': 'New York'},
+    )
+
+    results = await extract_attributes_from_nodes(
+        clients,
+        [node],
+        episode=_make_episode(),
+        previous_episodes=[],
+        entity_types=None,
+    )
+
+    assert results[0].attributes == {'age': 30, 'city': 'New York'}
+
+
+@pytest.mark.asyncio
+async def test_extract_attributes_preserves_prior_attributes_when_label_not_in_entity_types():
+    """Prior attributes survive when the node's label is absent from `entity_types`.
+
+    `entity_types.get(...)` resolves to None for any label not in the map, which takes the
+    same no-applicable-type path as `entity_types=None`.
+    """
+    from pydantic import BaseModel
+
+    class Organization(BaseModel):
+        industry: str | None = None
+
+    clients, _ = _make_clients()
+    clients.llm_client.generate_response = AsyncMock(return_value={'summaries': []})
+    clients.embedder.create = AsyncMock(return_value=[0.1, 0.2, 0.3])
+    clients.embedder.create_batch = AsyncMock(return_value=[[0.1, 0.2, 0.3]])
+
+    node = EntityNode(
+        name='Alice',
+        group_id='group',
+        labels=['Entity', 'Person'],
+        summary='Existing summary',
+        attributes={'age': 30, 'city': 'New York'},
+    )
+
+    results = await extract_attributes_from_nodes(
+        clients,
+        [node],
+        episode=_make_episode(),
+        previous_episodes=[],
+        entity_types={'Organization': Organization},
+    )
+
+    assert results[0].attributes == {'age': 30, 'city': 'New York'}


yatharth-anand-2410:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: yatharthanand2410@gmail.com

yatharth-anand-2410:
recheck

mehulp93:
Thanks for the contribution. The fix and the two regression tests look good. Merging this.

mehulp93 (approved):

PR #1765, opened 2026-08-13
fix(mcp): route FalkorDB UUID operations to tenant graphs

### LINKED ISSUE #1651: MCP server: get_episodes and delete_episode query the wrong graph on FalkorDB (bypass per-group routing)

noamsiegel:
## Summary

- Resolve UUID-keyed MCP reads and deletes against the FalkorDB graph that owns the requested data.
- Reject UUIDs found in multiple tenant graphs rather than reading or deleting an arbitrary match.
- Batch provenance resolution with one probe per graph and avoid `FalkorDriver.clone()` index-build side effects during reads.
- Serialize queued cross-group writes because `Graphiti.add_episode()` mutates the shared driver binding.

## Problem

FalkorDB stores each `group_id` as a physical graph. The remaining UUID-keyed MCP handlers bypass group-aware routing and query whichever graph the shared driver currently targets. Concurrent per-group queue workers can also race while `Graphiti.add_episode()` rebinds that shared driver. Both cases can report missing data or operate on the wrong tenant graph.

`get_episodes` is intentionally not changed here because current `main` already routes `EpisodicNode.get_by_group_ids` through the merged FalkorDB operations fix.

Fixes #1651.

## Verification

- `python -m pytest -q tests/test_falkor_routing.py tests/test_core_parity.py` - 34 passed.
- `python -m ruff format --check ...` and `python -m ruff check ...` - clean.
- `python -m pyright --pythonpath <venv-python> src/utils/falkor_routing.py src/graphiti_mcp_server.py src/services/queue_service.py` - 0 errors.
- Mutation checks fail when UUID matching, duplicate detection, non-initializing driver scoping, or cross-group write serialization is removed.
- Live FalkorDB reproduction confirmed cross-graph edge lookup, cross-graph episode deletion, and distinct-group concurrent writes on the patched MCP service.

### DIFF
--- a/mcp_server/src/graphiti_mcp_server.py
+++ b/mcp_server/src/graphiti_mcp_server.py
@@ -17,6 +17,7 @@ from dotenv import load_dotenv
 from graphiti_core import Graphiti
 from graphiti_core.edges import EntityEdge
 from graphiti_core.nodes import EntityNode, EpisodeType, SagaNode
+from graphiti_core.search.search_config import SearchResults
 from graphiti_core.search.search_filters import SearchFilters
 from graphiti_core.utils.maintenance.graph_data_operations import clear_data
 from mcp.server.fastmcp import FastMCP
@@ -44,6 +45,15 @@ from services.factories import (
     LLMClientFactory,
 )
 from services.queue_service import QueueService
+from utils.falkor_routing import (
+    ENTITY_EDGE_UUID_QUERY,
+    EPISODE_UUID_QUERY,
+    EPISODE_UUIDS_QUERY,
+    client_bound_to,
+    driver_for_uuid,
+    drivers_for_uuids,
+    is_falkor,
+)
 from utils.formatting import format_fact_result, to_edge_result, to_node_result
 from utils.type_config import (
     build_edge_type_map,
@@ -659,10 +669,14 @@ async def delete_entity_edge(uuid: str) -> SuccessResponse | ErrorResponse:
     try:
         client = await graphiti_service.get_client()
 
-        # Get the entity edge by UUID
-        entity_edge = await EntityEdge.get_by_uuid(client.driver, uuid)
-        # Delete the edge using its delete method
-        await entity_edge.delete(client.driver)
+        driver = client.driver
+        if is_falkor(driver):
+            driver, group_id = await driver_for_uuid(driver, ENTITY_EDGE_UUID_QUERY, uuid)
+            if group_id is None:
+                return ErrorResponse(error=f'Entity edge with UUID {uuid} not found')
+
+        entity_edge = await EntityEdge.get_by_uuid(driver, uuid)
+        await entity_edge.delete(driver)
         return SuccessResponse(message=f'Entity edge with UUID {uuid} deleted successfully')
     except Exception as e:
         error_msg = str(e)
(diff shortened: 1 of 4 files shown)


noamsiegel:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: noam@noamsiegel.com

noamsiegel:

Naseem77:
@noamsiegel Routing logic looks sound and your seven tests pass. Two things: `driver_for_uuid`/`drivers_for_uuids` query every graph from `list_graphs()`, so one UUID lookup is O(number of tenant graphs) and won't scale with many groups. And the tests only cover the new module so they can't run on main or show the original wrong-graph bug, an end-to-end MCP test would.

PR #1767, opened 2026-08-15
fix(attributes): surface cap-dropped attribute fields from add_episode

### LINKED ISSUE #1757: [Bug] `apply_capped_attributes`'s `dropped` field-name set is discarded at every call site — silent, unsignaled data loss
# [Bug] `apply_capped_attributes`'s `dropped` field-name set is discarded at every call site — silent, unsignaled data loss

## Summary

`cap_string_attributes()` / `apply_capped_attributes()`
(`graphiti_core/utils/maintenance/attribute_utils.py`) compute and return a
`dropped: set[str]` of field names removed for exceeding the length cap. The
function itself is capable of surfacing this information. But every real
call site discards it:

```python
# graphiti_core/utils/maintenance/node_operations.py:815
merged, _ = apply_capped_attributes(...)

# graphiti_core/utils/maintenance/edge_operations.py:669 and :796
merged, _ = apply_capped_attributes(...)
```

The only observable signal that a field was dropped is an `INFO`-level log
line (`attribute_length_cap_exceeded ...`). Callers of the public
`add_episode()` API — which is where this ultimately gets triggered when the
LLM emits an over-length attribute — have no return value, exception, or
field on `AddEpisodeResults` indicating that an attribute was silently
removed from an entity or edge.

## Type of Change

- [x] Bug fix (silent data loss with no programmatic signal)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

Note: fixing this changes a caller-visible contract (some return type needs
to start carrying the dropped-field information), so it may attract
`needs-rfc` — that's upstream's call to make, not something to pre-empt here.

## Objective

Surface the `dropped` set computed in `attribute_utils.py` up through
`extract_attributes_from_nodes` / edge attribute extraction to a place a
caller of `add_episode()` can actually observe it, instead of discarding it
with `merged, _ = apply_capped_attributes(...)` at every call site.

## What I was trying to do

Trigger the length cap on an entity attribute and then check whether
anything in the return value of the extraction call indicated which field
was dropped.

## What I expected

Some programmatic way to learn that a field was dropped — a non-empty
`dropped` set exposed on the result, or at minimum documentation stating that
the only signal is a log line.

## What actually happened

The `INFO` log line fires, but the returned `merged` dict simply lacks the
field, indistinguishable from the LLM never having proposed it at all.

## Reproduction

Pure in-process repro, no database required:

```python
import logging
from graphiti_core.utils.maintenance.attribute_utils import apply_capped_attributes

logging.basicConfig(level=logging.INFO)

class Person(BaseModel):
    bio: str | None = None

llm_response = {'bio': 'x' * 5000}  # over the default cap
merged, _ = apply_capped_attributes(  # <-- exact pattern used at every call site
    llm_response, Person, {}, merge_mode='overlay',
    prompt_name='repro', entity_uuid='repro-entity', group_id='repro-group',
)
print('merged result:', merged)
assert 'bio' not in merged
```

Actual output against `main`:

```
INFO:graphiti_core.utils.maintenance.attribute_utils:attribute_length_cap_exceeded prompt=repro group_id=repro-group entity_uuid=repro-entity field=bio reason=per_item length=5000 cap=250
merged result: {}
```

The log line fires; `merged` carries nothing indicating `bio` was ever
present or dropped.

## Testing

Repro above demonstrates the discard pattern directly, matching the exact
`merged, _ = apply_capped_attributes(...)` call shape found at
`node_operations.py:815`, `edge_operations.py:669`, and `edge_operations.py:796`.
A fix should be validated by asserting the dropped-field information is
retrievable from whatever surface `add_episode()` exposes it through.

## Breaking Changes

Likely yes, depending on the fix shape (e.g. adding a field to
`AddEpisodeResults` is additive/non-breaking; changing an existing return
type's shape would not be). Flagging for upstream discussion rather than
prescribing the fix here.

## Checklist

- [x] Reproduced against `main`
- [x] Reproduction requires no database, matches real call-site pattern
      exactly
- [x] Framed as silent data loss, not an enhancement request

## Related Issues

None found covering this defect.

aryansk:
Closes #1757

## What

`apply_capped_attributes()` already computes and returns the `dropped` set of field names removed for exceeding the length cap — but every call site discarded it with `merged, _ = ...`, so an `INFO` log line was the only observable signal that an attribute was silently removed from an entity or edge. A caller of `add_episode()` had no way to tell "the LLM never proposed this field" from "the LLM proposed it but it was dropped for length".

This PR threads the dropped set through both extraction paths and surfaces it on the public result:

- **Node path** — `_extract_entity_attributes()` now returns `(merged, dropped)`; `extract_attributes_from_nodes()` returns `(nodes, dropped_by_uuid)`.
- **Edge path** — `resolve_extracted_edge()` and `resolve_extracted_edges()` return the dropped set / a `dropped_by_uuid` aggregate as an added tuple element.
- **Public API** — `AddEpisodeResults` and `AddBulkEpisodeResults` gain a `dropped_attributes: dict[str, list[str]]` field mapping entity/edge uuid → sorted dropped field names (empty by default, so existing deserialization is unaffected). `add_episode()`/`add_episode_bulk()` merge the node + edge maps into it.
- `bulk_utils` and `add_triplet` unpack the added tuple element (they don't need the info yet).

## Why

Silent, unsignaled data loss. When the LLM emits an over-length attribute, the merged result simply lacks the field — indistinguishable from the field never being proposed. The new field gives programmatic access to exactly which fields were dropped and for which entities/edges, without relying on parsing log lines.

## Tests

- `test_extract_attributes_from_nodes_surfaces_dropped_fields` — node path: over-cap `bio` with an explicit `max_length` is dropped, prior value retained by the overlay merge, and reported in `dropped_by_uuid`.
- `test_resolve_extracted_edge_surfaces_dropped_attributes` — edge path: over-cap `review` reported in the returned dropped set, absent from merged attributes (replace mode).
- Updated existing mocks/callers for the 4-tuple and 2-tuple returns.

Full non-DB unit suite passes (116 tests in `tests/utils/maintenance/` + `test_add_triplet.py`); ruff check + format clean; pyright clean on changed files (only pre-existing `dotenv` import resolution issue in `graphiti.py` when deps are missing).

### DIFF
--- a/graphiti_core/graphiti.py
+++ b/graphiti_core/graphiti.py
@@ -20,7 +20,7 @@ from time import time
 from uuid import uuid4
 
 from dotenv import load_dotenv
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from typing_extensions import LiteralString
 
 from graphiti_core.cross_encoder.client import CrossEncoderClient
@@ -118,6 +118,7 @@ class AddEpisodeResults(BaseModel):
     edges: list[EntityEdge]
     communities: list[CommunityNode]
     community_edges: list[CommunityEdge]
+    dropped_attributes: dict[str, list[str]] = Field(default_factory=dict)
 
 
 class AddBulkEpisodeResults(BaseModel):
@@ -127,6 +128,7 @@ class AddBulkEpisodeResults(BaseModel):
     edges: list[EntityEdge]
     communities: list[CommunityNode]
     community_edges: list[CommunityEdge]
+    dropped_attributes: dict[str, list[str]] = Field(default_factory=dict)
 
 
 class AddTripletResults(BaseModel):
@@ -639,16 +641,17 @@ class Graphiti:
         nodes: list[EntityNode],
         uuid_map: dict[str, str],
         custom_extraction_instructions: str | None = None,
-    ) -> tuple[list[EntityEdge], list[EntityEdge], list[EntityEdge]]:
+    ) -> tuple[list[EntityEdge], list[EntityEdge], list[EntityEdge], dict[str, set[str]]]:
         """Extract edges from episode(s) and resolve against existing graph.
 
         Returns
         -------
-        tuple[list[EntityEdge], list[EntityEdge], list[EntityEdge]]
-            A tuple of (resolved_edges, invalidated_edges, new_edges) where:
+        tuple[list[EntityEdge], list[EntityEdge], list[EntityEdge], dict[str, set[str]]]
+            A tuple of (resolved_edges, invalidated_edges, new_edges, dropped_by_uuid) where:
             - resolved_edges: All edges after resolution
             - invalidated_edges: Edges invalidated by new information
             - new_edges: Only edges that are new to the graph (not duplicates)
+            - dropped_by_uuid: Mapping of edge uuid -> cap-dropped attribute fields
         """
         episodes = episode if isinstance(episode, list) else [episode]
         primary_episode = episodes[0]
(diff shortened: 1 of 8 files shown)


aryansk:
Acknowledged — the CLA signature is a contributor legal attestation that remains outstanding and will be completed by the contributor before this draft is marked ready. No code change is needed for the check itself.

PR #1768, opened 2026-08-15
fix(falkordb): pin redis<8.1 in falkordb and falkordblite extras

### LINKED ISSUE #1756: [Bug] `graphiti-core[falkordb]` is broken on a clean install: unpinned `redis` resolves 8.1.0, `FalkorDriver` construction raises `TypeError`
# [Bug] `graphiti-core[falkordb]` is broken on a clean install: unpinned `redis` resolves 8.1.0, `FalkorDriver` construction raises `TypeError`

## Summary

`pyproject.toml`'s `falkordb` extra has no ceiling on `redis`:

```toml
falkordb = ["falkordb>=1.1.2,<2.0.0"]
# redis<9 pin: redis-py 8.x breaks falkordblite's embedded server startup
falkordblite = ["falkordblite>=0.5.0; python_version>='3.12'", "redis<9; python_version>='3.12'"]
```

The comment on `falkordblite` shows the project is already aware that
redis-py 8.x breaks things downstream of `falkordb-py` — but the pin was only
applied to the `falkordblite` extra, not `falkordb`. A clean
`pip install graphiti-core[falkordb]` resolves `redis==8.1.0`, and
constructing a `FalkorDriver` against a real FalkorDB instance then raises a
`TypeError` before any query is even issued.

**A note on the existing `falkordblite` pin:** `redis<9` does not actually
exclude the broken versions — the breakage starts at redis 8.1.0, and
`8.1.0 < 9`. That extra currently resolves a working redis only because the
root lockfile pins lower for unrelated reasons, not because the `<9` ceiling
is doing the work its comment describes. So the fix for `falkordb` should not
simply copy that ceiling, and `falkordblite` likely needs the same correction.
The boundary below was established by bisection, not inference.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Objective

Add a `redis<8.1.0` ceiling to the `falkordb` extra so
`pip install graphiti-core[falkordb]` is usable out of the box, and consider
correcting `falkordblite`'s `redis<9` to the same boundary.

Bisected against a disposable FalkorDB container: **redis 8.0.0 constructs a
`FalkorDriver` cleanly; redis 8.1.0 fails 100% of the time.** `redis<9` would
leave the defect in place.

(This is fixable at either layer — `falkordb-py`'s `Is_Cluster`
passing async connection kwargs into a sync `Redis` client, or
`graphiti-core` adding the version ceiling described above. Filing here since
the immediate user-facing breakage is in
this package's dependency metadata; a companion report against `falkordb-py`
covers the other half.)

## What I was trying to do

A clean `pip install graphiti-core[falkordb]` in a fresh virtual environment,
then construct a `FalkorDriver` against a running FalkorDB instance — the
first thing any new FalkorDB user would do.

## What I expected

The driver to construct successfully and connect.

## What actually happened

```
TypeError: Redis.__init__() got an unexpected keyword argument 'himport_registry'
```

raised from inside `falkordb-py`'s cluster-detection path, before any user
code runs.

## Reproduction

```
python -m venv venv_repro
venv_repro\Scripts\pip install "graphiti-core[falkordb]"
venv_repro\Scripts\pip show redis   # -> Version: 8.1.0
venv_repro\Scripts\python -c "from graphiti_core.driver.falkordb_driver import FalkorDriver; FalkorDriver(host='localhost', port=6379)"
```

Actual traceback observed (redis 8.1.0, falkordb-py 1.6.2, graphiti-core
0.29.3, against a disposable FalkorDB container):

```
File ".../graphiti_core/driver/falkordb_driver.py", line 163, in __init__
    self.client = FalkorDB(host=host, port=port, username=username, password=password)
File ".../falkordb/asyncio/falkordb.py", line 118, in __init__
    if Is_Cluster(conn):
File ".../falkordb/asyncio/cluster.py", line 27, in Is_Cluster
    info = sync_redis.Redis(**kwargs).info(section="server")
File ".../redis/utils.py", line 258, in wrapper
    return func(*args, **kwargs)
TypeError: Redis.__init__() got an unexpected keyword argument 'himport_registry'
```

`falkordb-py`'s `Is_Cluster()` builds a synchronous `redis.Redis` client from
kwargs assembled for the *async* connection pool; redis-py 8.x's sync client
no longer accepts one of those kwargs, so this raises immediately.

## Testing

Reproduced in a disposable venv against a disposable FalkorDB container
(no production data involved). Resolved versions: `redis==8.1.0`,
`FalkorDB==1.6.2` (the `falkordb-py` package), `graphiti-core==0.29.3`.
A fix (pinning `redis<9` on the `falkordb` extra) can be verified by
re-running the same install + construction and confirming the resolved
`redis` version and successful construction.

## Breaking Changes

None for existing users already pinning a compatible `redis` version — by
definition anyone with a working `graphiti-core[falkordb]` install today is
already below 8.1.0, since 8.1.0 and above cannot construct a driver at all.

Users who have resolved `redis>=8.1.0` alongside `graphiti-core[falkordb]`
would see their environment become unresolvable until they downgrade — but
those installs are already non-functional for FalkorDB, so the constraint
surfaces an existing breakage at install time rather than at first use, which
is the intent.

## Checklist

- [x] Reproduced in a clean virtual environment using the published package;
      the `falkordb` extra's metadata is identical on `main`, and the defect
      is in the packaging metadata itself rather than in library code
- [x] Root cause traced into `falkordb-py`'s `Is_Cluster` (external package);
      graphiti-core's fix is the missing version ceiling
- [x] Boundary established by bisection (8.0.0 works, 8.1.0 fails), not by
      copying the existing `falkordblite` ceiling — which does not exclude
      the broken versions

## Related Issues

None found. A companion bug report against `falkordb-py` covers
`Is_Cluster` building a sync `Redis` client from async pool kwargs.

aryansk:
Closes #1756

## Problem

A clean `pip install graphiti-core[falkordb]` resolved `redis==8.1.0`, and constructing a `FalkorDriver` against a real FalkorDB instance then raised a `TypeError` before any query was issued. Root cause is upstream in `falkordb-py` (FalkorDB/falkordb-py#261: `Is_Cluster` builds a sync `redis.Redis` from async pool kwargs), but graphiti-core's own dependency metadata made it reach every user by default.

The existing `falkordblite` guard (`redis<9`) was also ineffective for the same reason: the breakage starts at redis **8.1.0**, and `8.1.0 < 9` — that extra only worked because the root lockfile pinned lower for unrelated reasons.

## Fix

- `falkordb` extra: add `redis<8.1`
- `falkordblite` extra: tighten `redis<9` → `redis<8.1` (and document why the old ceiling was wrong)
- `uv.lock`: reflect the new constraints in both `[package.optional-dependencies]` and `requires-dist` metadata

## Verification

`uv pip install --dry-run` against both extras now downgrades redis from the broken `8.1.0` to `8.0.1`:

```
 - redis==8.1.0
 + redis==8.0.1
```

`uv lock --check` passes with the minimal, targeted lockfile diff (only the redis constraint lines changed — no platform-marker churn).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,9 +30,12 @@ groq = ["groq>=0.2.0"]
 google-genai = ["google-genai>=1.62.0"]
 # Deprecated: the upstream Kuzu project is unmaintained; this extra will be removed in a future release.
 kuzu = ["kuzu>=0.11.3"]
-falkordb = ["falkordb>=1.1.2,<2.0.0"]
-# redis<9 pin: redis-py 8.x breaks falkordblite's embedded server startup
-falkordblite = ["falkordblite>=0.5.0; python_version>='3.12'", "redis<9; python_version>='3.12'"]
+falkordb = ["falkordb>=1.1.2,<2.0.0", "redis<8.1"]
+# redis<8.1 pin: redis-py 8.1.0 breaks FalkorDriver construction (Is_Cluster
+# builds a sync redis.Redis from async pool kwargs — FalkorDB/falkordb-py#261).
+# falkordblite needs the same ceiling: redis-py 8.1.0 also breaks its embedded
+# server startup, and the old <9 pin did not exclude it.
+falkordblite = ["falkordblite>=0.5.0; python_version>='3.12'", "redis<8.1; python_version>='3.12'"]
 voyageai = ["voyageai>=0.2.3"]
 gliner2 = ["gliner2>=1.2.0; python_version>='3.11'"]
 neo4j-opensearch = ["boto3>=1.39.16", "opensearch-py>=3.0.0"]
--- a/uv.lock
+++ b/uv.lock
@@ -1141,6 +1141,7 @@ dev = [
 ]
 falkordb = [
     { name = "falkordb" },
+    { name = "redis" },
 ]
 falkordblite = [
     { name = "falkordblite", marker = "python_full_version >= '3.12'" },
@@ -1219,7 +1220,8 @@ requires-dist = [
     { name = "pytest-asyncio", marker = "extra == 'dev'", specifier = ">=0.24.0" },
     { name = "pytest-xdist", marker = "extra == 'dev'", specifier = ">=3.6.1" },
     { name = "python-dotenv", specifier = ">=1.0.1" },
-    { name = "redis", marker = "python_full_version >= '3.12' and extra == 'falkordblite'", specifier = "<9" },
+    { name = "redis", marker = "python_full_version >= '3.12' and extra == 'falkordblite'", specifier = "<8.1" },
+    { name = "redis", marker = "extra == 'falkordb'", specifier = "<8.1" },
     { name = "ruff", marker = "extra == 'dev'", specifier = ">=0.7.1" },
     { name = "sentence-transformers", marker = "extra == 'dev'", specifier = ">=3.2.1" },
     { name = "sentence-transformers", marker = "extra == 'sentence-transformers'", specifier = ">=3.2.1" },


aryansk:
Acknowledged — the CLA signature is a contributor legal attestation that remains outstanding and will be completed by the contributor before this draft is marked ready. No code change is needed for the check itself.

Naseem77:
@aryansk The breakage was real but the premise is stale: falkordb-py#261 shipped in v1.7.0/v1.7.1 on 2026-08-13, and a clean resolve on main today gives `falkordb 1.7.1 + redis 8.1.0`, which I verified constructs a driver and queries fine live. The `TypeError` only reproduces with `falkordb<=1.6.2`, so `redis<8.1` holds everyone on an older redis for an already-fixed bug and still allows the broken falkordb. A floor of `falkordb>=1.7.1` with redis unpinned fixes the real cause, tested and works.

PR #1772, opened 2026-08-16
fix: per-list E/I id namespaces in the resolution judge (#1728)

Deepnar:
# PR description — fix: per-list E/I id namespaces in the resolution judge

Reference: #1728 (Bug: Edge invalidation searches the whole graph, so unrelated facts retire each other)

## What this changes

The resolution judge (`dedupe_edges.resolve_edge`) previously numbered both candidate
lists — EXISTING FACTS (related/duplicate candidates) and FACT INVALIDATION CANDIDATES —
in one continuous index range, and the parser reconstructed list membership with offset
arithmetic (`invalidation_idx_offset`). A judge response that was "in range" but belonged
to the wrong list silently retired unrelated facts.

This PR gives each list its own namespace:
- EXISTING FACTS are `E0, E1, ...`
- FACT INVALIDATION CANDIDATES are `I0, I1, ...`
- The parser maps prefixes directly (`E3` -> related[3], `I2` -> existing[2]); malformed
  or out-of-range ids are dropped with a warning instead of computed.

## Why it's correct (not just a style change)

Under the continuous-index scheme, "wrong-but-in-range" is a real failure class: the
judge can return an index that is valid for the combined range but points at the wrong
list. Prefixed ids make that structurally impossible — the prefix IS the list.

## Files changed

- `graphiti_core/prompts/dedupe_edges.py` — response model fields are now `list[str]`
  (E/I-prefixed), prompt constraints and few-shot examples updated to the per-list
  numbering.
- `graphiti_core/utils/maintenance/edge_operations.py` — context building emits
  `{'id': 'E{i}'}` / `{'id': 'I{i}'}`; `invalidation_idx_offset` deleted; parsing replaced
  with `_parse_edge_ids` direct mapping; `max_valid_idx` arithmetic removed.
- `tests/utils/maintenance/test_edge_operations.py` — existing integer-index test renamed
  and updated to E-prefixed mocks; new tests: malformed ids (`'X'`, `'E'`, `'E999'`,
  `'I-1'`) dropped safely; `'I2'` maps to `existing_edges[2]` (the regression that would
  have caught the offset bug).

## Verification

- `pytest tests/utils/maintenance/test_edge_operations.py` — 11 passed.
- ruff + pyright clean on changed files.
- `grep invalidation_idx_offset|max_valid_idx` — no remaining references.

## Real-model measurement (what motivated the change)

Ran the actual resolution path with a real API-class model (deepseek-v4-flash) over a
120-probe corpus (30 each: true duplicates, contradictions, false-similars, renames),
with pre-written ground truth and deterministic scoring. Same corpus, same model, both
code versions:

| class | before (continuous idx) | after (E/I namespaces) |
|---|---|---|
| duplicate | 18/30 | 20/30 |
| contradiction | 0/30 | 0/30 |
| false_similar | 30/30 | 29/30* |
| rename | 0/30 | 0/30 |
| **total** | **48/120** | **49/120** |

*the one false_similar flip was a response-shape parse error, not a semantic regression.

The three duplicate recoveries (duplicate-13/-23/-27) are direct evidence the fix works:
in both arms the raw model output was IDENTICAL (correct `E2`-style ids), and only the
parser differed — the E/I mapping lands merges the continuous-index arithmetic dropped.

## What this does NOT claim (honest scope)

1. The dominant resolution failure is NOT the index scheme: ~50% of harder calls
   (contradiction/rename) return no duplicate/contradiction verdict at all, in BOTH arms.
   Sampling the raw responses showed the model returns valid, correct ids on roughly half
   of those — the failure is judge reliability/stochasticity under this prompt, which this
   PR does not fix.
2. Response-shape fragility is a separate issue: occasionally the model returns the
   injected schema wrapper instead of a filled instance, breaking the parse
   (`ValidationError` on `duplicate_facts`). Worth a follow-up issue.
3. Token limits were verified non-confounding: all calls used max_tokens=16384 (the
   client default, matching production), far above what these small JSON responses need.

Happy to share the harness + corpus (seeded, resumable, deterministic-scored) for
maintainers to reuse.

### DIFF
--- a/graphiti_core/prompts/dedupe_edges.py
+++ b/graphiti_core/prompts/dedupe_edges.py
@@ -22,13 +22,15 @@ from .models import Message, PromptFunction, PromptVersion
 
 
 class EdgeDuplicate(BaseModel):
-    duplicate_facts: list[int] = Field(
+    duplicate_facts: list[str] = Field(
         ...,
-        description='List of idx values of duplicate facts (only from EXISTING FACTS range). Empty list if none.',
+        description='List of E-prefixed ids of duplicate facts, e.g. ["E0", "E3"]. '
+        'Only from EXISTING FACTS. Empty list if none.',
     )
-    contradicted_facts: list[int] = Field(
+    contradicted_facts: list[str] = Field(
         ...,
-        description='List of idx values of contradicted facts (from full idx range). Empty list if none.',
+        description='List of E- or I-prefixed ids of contradicted facts, e.g. ["E1", "I2"]. '
+        'Empty list if none.',
     )
 
 
@@ -53,9 +55,9 @@ def resolve_edge(context: dict[str, Any]) -> list[Message]:
 NEVER mark facts as duplicates if they have key differences, particularly around numeric values, dates, or key qualifiers.
 
 IMPORTANT constraints:
-- duplicate_facts: ONLY idx values from EXISTING FACTS (NEVER include FACT INVALIDATION CANDIDATES)
-- contradicted_facts: idx values from EITHER list (EXISTING FACTS or FACT INVALIDATION CANDIDATES)
-- The idx values are continuous across both lists (INVALIDATION CANDIDATES start where EXISTING FACTS end)
+- duplicate_facts: ONLY E-prefixed ids from EXISTING FACTS (NEVER include FACT INVALIDATION CANDIDATES)
+- contradicted_facts: E- or I-prefixed ids from EITHER list (EXISTING FACTS or FACT INVALIDATION CANDIDATES)
+- E and I each number from 0 independently (e.g. E0, E1, ... and I0, I1, ...)
 
 <EXISTING FACTS>
 {context['existing_edges']}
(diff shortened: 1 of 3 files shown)


Deepnar:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 18deepnar@gmail.com

Deepnar:
recheck

Deepnar:
Rebased onto current `main` and re-ran the relevant checks:

* `tests/utils/maintenance/test_edge_operations.py`: 11 passed
* Ruff: clean
* Ruff formatting check: clean
* Pyright: 0 errors

I also re-checked this against #1729 and the later #1728 discussion. The changes still appear complementary: #1729 bounds which edges can enter invalidation, while this PR removes ambiguity in how the resolution judge refers to the duplicate vs invalidation candidate lists.

Happy to adjust the implementation if maintainers prefer a different approach. A review when someone has bandwidth would be appreciated.

PR #1362, opened 2026-03-31
fix: validate LLM responses in OpenAIGenericClient to prevent Pydantic errors

theskipper007:
## Summary

Fixes #912 — Pydantic `ValidationError` when using non-OpenAI LLMs (Ollama, vLLM, deepseek, etc.) with `OpenAIGenericClient`.

**Root cause:** Many local LLMs don't support the `json_schema` response format and return the schema definition itself (containing `$defs`, `$ref`, etc.) instead of actual data. This causes `ValidationError` in downstream Pydantic models like `ExtractedEntities`, `NodeResolutions`, and `EdgeDuplicate`.

**Changes:**
- Use `json_object` response format instead of `json_schema` for maximum compatibility — the schema is already embedded in the prompt text by the base `LLMClient` class
- Validate LLM responses against the `response_model` with Pydantic **before** returning, matching the existing pattern in `AnthropicClient` (lines 401-404)
- Provide specific `ValidationError` messages on retry so the LLM knows exactly what schema to follow
- Add comprehensive test suite for `OpenAIGenericClient` (8 tests)

## Test plan
- [x] All 76 existing `tests/llm_client/` tests pass (0 failures)
- [x] 8 new tests cover: valid response validation, schema-echo retry, max retry exhaustion, no-model passthrough, json_object format verification
- [ ] Manual testing with Ollama + local model (e.g., `deepseek-r1:7b`) to confirm the schema-echo scenario is handled

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1773, opened 2026-08-16
fix(llm): tolerate non-enforced json_schema output in extraction models

### LINKED ISSUE #871: [BUG] Add Episodes function throws Invalid JSON and index out of range errors
## Bug Description
We have a corpus of text that we sectionized and treat each section as an episode. We use add_episodes function to add the sections/episodes one by one into the graph. The process stalls with JSON invalidation and "index out of range" errors. We are sure that we feed episodes text that is correct json format and the token length is not more than max tokens the models can input.

## Steps to Reproduce
- Set-up the client
- Prepare data
- Add episodes

```python
import json
from time import sleep

failed_chunks = {}
episodes = {}
for i, episode in enumerate(sections_with_stages):
    chunk = {
        "condition": episode['condition'],
        "title": episode['chunk']['title'],
        "content": episode['chunk']['content'],
        "Relevant stage": f"{episode['most_relevant_stage']} stage",
        "Second relevant stage": f"{episode['secondary_relevant_stage']} stage",
    }
    episode_name = f'{episode["condition"]}-guideline-section-{i}'
    episodes[episode_name] = chunk

for name, episode in episodes.items():
    try:
        await client.add_episode(
            name=name,
            episode_body=json.dumps(episode),
            source_description=f"Title - {episode['title']}",
            source=EpisodeType.json,
            entity_types=entity_types,
            edge_types=edge_types,
            edge_type_map=edge_type_map,
            reference_time=datetime.now(),
            group_id='guidelines', # Group id is a namespace for the data
        )
    except Exception as e:
        print("ERROR")
        print(f"Error adding episode {name}: {e}")
        # Add chunk data to a dict with i as the key if error is thrown
        failed_chunks[i] = chunk
        continue
    print(f'Added episode:{name} ({EpisodeType.text.value})')
```

## Expected Behavior
I would expect the process to run and walk through all episodes extracting nodes and edges without stalling or errors. 

## Actual Behavior
index out of range breaks the code, the other invalid JSON keeps retrying (stalls) but finishes.

## Environment
- **Graphiti Version**: 0.18.9
- **Python Version**: 3.12
- **Operating System**: macOS 15.6.1,
- **Database Backend**: Neo4j Desktop 2.0.3
- **LLM Provider & Model**: openAI gpt-4.1-mini and gpt-4o-mini (small_model).

## Installation Method
- [ ] pip install

## Error Messages/Traceback
```
graphiti_core.llm_client.openai_base_client - ERROR - Error in generating LLM response: 1 validation error for NodeResolutions
  Invalid JSON: EOF while parsing a string at line 1 column 33267 [type=json_invalid, input_value='{"entity_resolutions":[{...}]}]}]}]}]}]}]}]}]}]}]}', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/json_invalid
OR 
index out of range
```

## Configuration
```python
# Configure Graphiti
from graphiti_core.utils.maintenance.graph_data_operations import clear_data
import os

neo4j_uri = os.environ['NEO4J_URI']
neo4j_user = os.environ['NEO4J_USER']
neo4j_password = os.environ['NEO4J_PASSWORD']
DEFAULT_DATABASE = os.environ['DEFAULT_DATABASE']

client = Graphiti(
    neo4j_uri,
    neo4j_user,
    neo4j_password,
)
client.driver._database = DEFAULT_DATABASE
client.llm_client.model = "gpt-4.1-mini"
client.llm_client.small_model = "gpt-4o-mini"
client.llm_client.max_tokens = 16000
print(client.llm_client.model, client.driver._database)
```

## Additional Context
- This happens constantly. 

## Possible Solution
Not sure. JSON invalidation might be structured output issue in the intermediary steps. I do not have an idea regarding the "Index problem"

fabiosiqueira:
## Summary

Fixes #871.

`OpenAIGenericClient` serves every OpenAI-compatible endpoint that isn't `api.openai.com` — Ollama, Ollama Cloud, vLLM, LiteLLM, etc. Several of these accept `response_format: json_schema` without actually applying it as constrained decoding: the model free-forms JSON from the prompt text instead, deviating from the schema in ways real constrained-decoding servers never would.

I hit this while benchmarking Ollama Cloud's free-tier models against Graphiti. Reproduced across 5 model families and all 3 free-tier usage tiers, all with structurally identical failures:
- a bare JSON array instead of the wrapped `{field: [...]}` object
- array items as bare strings instead of objects
- the entity name field emitted under a different key on every call (`entity`, `entity_text`, `entity_node`, `entity_node_name`, ...) — genuine non-determinism between otherwise-identical calls, not a fixed rename, so an enumerable alias list doesn't converge (confirmed by trying one first, then generalizing)

This is the same symptom reported in #871 (`gpt-oss-20b`, `ExtractedEdges: edges Field required`) and the earlier #912.

## Approach

`#1362` took a client-side approach to this same underlying problem (validate + retry-with-error-message inside `OpenAIGenericClient`) and was closed in favor of `#1537`, with the note that response validation should stay at the call sites rather than move into the client. This PR follows that guidance: the tolerance lives in `model_validator(mode='before')` hooks on the three response models actually observed hitting this failure — `ExtractedEntities`, `ExtractedEdges`, `NodeResolutions` — not in the client.

Scoped to reproduced failures rather than a blanket rewrite:
- wrap a bare array into the model's single required array field
- wrap a bare string into `{name: <string>}`
- recover the entity name from whichever key is a lone string candidate once type-metadata keys (anything containing `"type"`) are excluded — so a same-call extra key like `entity_type_name` can't tie with the real name field and make the candidate set ambiguous
- default `entity_type_id` to `0` when missing or given as a string
- synthesize `Edge.fact` from `source_entity_name`/`relation_type`/`target_entity_name` when absent

One mechanical consequence: wrapping a bare array can't be done as a pure `mode='before'` validator without also switching the call site from `Model(**data)` to `Model.model_validate(data)`, since `**` on a list fails at the Python level before Pydantic ever runs. Updated the three affected call sites in `node_operations.py` (`ExtractedEntities`, `NodeResolutions`) and `edge_operations.py` (`ExtractedEdges`) — `NodeResolutions` turned up while re-verifying end-to-end after moving the array-wrap fix out of the client, so it's included for the same reason as the other two, not a speculative extension.

## Verification

- End-to-end against Ollama Cloud's `gemma4:31b` (free tier) on a 6-episode/6-checkpoint extraction + temporal-invalidation scenario: 6/6 on one run, 5/6 on a second (the sole miss was the free tier's usage ceiling kicking in mid-scenario, not a validation error — confirmed by isolating that episode, which always passes on its own).
- Existing test suite for the touched modules (`extract_nodes`, `extract_edges`, `dedupe_nodes`, `openai_generic_client`, `node_operations`, `edge_operations`): 60 tests passing.
- `ruff check`, `ruff format --check`, `pyright` clean on all changed files.

— 🤖 Claude Sonnet 5

### DIFF
--- a/graphiti_core/llm_client/client.py
+++ b/graphiti_core/llm_client/client.py
@@ -214,10 +214,17 @@ class LLMClient(ABC):
 
         if response_model is not None:
             serialized_model = json.dumps(response_model.model_json_schema())
-            messages[
-                -1
-            ].content += (
-                f'\n\nRespond with a JSON object in the following format:\n\n{serialized_model}'
+            # "in the following format" is ambiguous: some models (observed on
+            # google/gemini-2.5-flash-lite, deepseek/deepseek-v4-flash, openai/gpt-4.1-nano
+            # via OpenRouter) read it as "copy this structure" and echo the JSON Schema
+            # itself back — nesting the real answer under a "properties" key instead of
+            # returning a flat instance. Spell out that this is a schema to satisfy, not an
+            # example to reproduce.
+            messages[-1].content += (
+                '\n\nRespond with a single JSON object that is a valid INSTANCE of the '
+                'JSON Schema below — i.e. fill in real values for each property. Do not '
+                'reproduce schema keywords such as "properties", "type", "title", or '
+                f'"$defs" in your answer.\n\nJSON Schema:\n{serialized_model}'
             )
 
         # Add multilingual extraction instructions
--- a/graphiti_core/llm_client/openai_generic_client.py
+++ b/graphiti_core/llm_client/openai_generic_client.py
@@ -193,10 +193,17 @@ class OpenAIGenericClient(LLMClient):
         # response_format, so no prompt injection is needed.
         if response_model is not None and self.structured_output_mode == 'json_object':
             serialized_model = json.dumps(response_model.model_json_schema())
-            messages[
-                -1
-            ].content += (
-                f'\n\nRespond with a JSON object in the following format:\n\n{serialized_model}'
+            # "in the following format" is ambiguous: some models (observed on
+            # google/gemini-2.5-flash-lite, deepseek/deepseek-v4-flash, openai/gpt-4.1-nano
+            # via OpenRouter) read it as "copy this structure" and echo the JSON Schema
+            # itself back — nesting the real answer under a "properties" key instead of
+            # returning a flat instance. Spell out that this is a schema to satisfy, not an
+            # example to reproduce.
+            messages[-1].content += (
+                '\n\nRespond with a single JSON object that is a valid INSTANCE of the '
+                'JSON Schema below — i.e. fill in real values for each property. Do not '
+                'reproduce schema keywords such as "properties", "type", "title", or '
+                f'"$defs" in your answer.\n\nJSON Schema:\n{serialized_model}'
             )
 
         # Add multilingual extraction instructions
--- a/graphiti_core/prompts/dedupe_nodes.py
+++ b/graphiti_core/prompts/dedupe_nodes.py
@@ -16,7 +16,7 @@ limitations under the License.
 
 from typing import Any, Protocol, TypedDict
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 from .models import Message, PromptFunction, PromptVersion
 from .prompt_helpers import to_prompt_json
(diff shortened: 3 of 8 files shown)


fabiosiqueira:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: fabioxxx@gmail.com

fabiosiqueira:
recheck

fabiosiqueira:
recheck

fabiosiqueira:
Pushed a second commit that closes the other half of this failure mode.

The first commit tolerates a response that mirrors the JSON Schema instead of instantiating it. This one removes the reason models do that: the injected instruction read `Respond with a JSON object in the following format:` followed by a schema, which several models take as "reproduce this structure" — so they return the schema, nesting the real answer under a `properties` key.

The instruction now says the schema is a contract to satisfy rather than an example to copy, and names the keywords that must not appear in the answer.

Observed on three models routed through OpenRouter, on a longmemeval slice using `tests/evals/eval_e2e_graph_building.py`:

| model | score vs baseline |
| --- | --- |
| `deepseek/deepseek-v4-flash` | 0.87 |
| `google/gemini-2.5-flash-lite` | 0.80 |
| `openai/gpt-4.1-nano` | 0.47 |

Happy to split the prompt change into its own PR if you'd rather keep this one to the parsing side.

— 🤖 Opus 5

PR #1780, opened 2026-08-18
fix(neo4j): route database_ to connection kwarg (#1481)

### LINKED ISSUE #1481: [BUG] Database Parameter Not Honored in Neo4jDriver.execute_query()

gauravch-code:
## Summary

Fixes #1481.

`Neo4jDriver.execute_query` was placing `database_` inside `parameters_` (the Cypher variables dict) rather than passing it as a driver-level kwarg to Neo4j's `execute_query`. As a result, `database_` was treated as an unused Cypher parameter and silently ignored — every query routed to the connection's home database regardless of the `database` value passed to `Neo4jDriver(...)`. Only writes that went through `session(database=...)` explicitly (e.g. `transaction()`) hit the intended database.

## Fix

Route `database_` to the driver-level kwarg on `AsyncDriver.execute_query`. Accept it from either `params` or `kwargs` for backwards compatibility so callers that historically stuck it in `params` keep working.

```python
database = (
    params.pop('database_', None)
    or kwargs.pop('database_', None)
    or self._database
)
result = await self.client.execute_query(
    cypher_query_,
    parameters_=params,
    database_=database,
    **kwargs,
)
```

## Tests

New `tests/driver/test_neo4j_driver.py` — 7 tests, mock-only (no live Neo4j required):

- `test_configured_database_is_passed_as_connection_kwarg` — the core regression for #1481
- `test_default_database_is_used_when_none_configured`
- `test_explicit_database_kwarg_overrides_configured_default`
- `test_database_nested_in_params_is_promoted_to_connection_kwarg` — backwards compat
- `test_routing_kwarg_is_forwarded_at_connection_level`
- `test_extra_kwargs_are_forwarded`
- `test_provider`

Added the new file to the `database-integration-tests` job in `.github/workflows/unit_tests.yml`, mirroring the existing `test_falkordb_driver.py` entry.

## Verified locally

- `ruff format` + `ruff check` — clean
- `pyright graphiti_core/driver/neo4j_driver.py` — 0 errors, 0 warnings
- `pytest tests/driver/test_falkordb_driver.py tests/driver/test_neo4j_driver.py` — 32 passed, 1 skipped
- Full unit suite with CI's ignore list — 366 passed, 11 skipped, no regressions

## Test plan

- [x] Regression test fails on unmodified `main` (asserts `database_` is a top-level kwarg; observed `{'parameters_': {'database_': 'mydb'}}` on main)
- [x] Regression test passes after fix
- [x] Full unit suite passes
- [x] `ruff` + `pyright` clean

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -112,4 +112,5 @@ jobs:
             tests/test_edge_int.py \
             tests/cross_encoder/test_bge_reranker_client_int.py \
             tests/driver/test_falkordb_driver.py \
+            tests/driver/test_neo4j_driver.py \
             -m "not integration"
--- a/graphiti_core/driver/neo4j_driver.py
+++ b/graphiti_core/driver/neo4j_driver.py
@@ -158,15 +158,23 @@ class Neo4jDriver(GraphDriver):
                 raise
 
     async def execute_query(self, cypher_query_: LiteralString, **kwargs: Any) -> EagerResult:
-        # Check if database_ is provided in kwargs.
-        # If not populated, set the value to retain backwards compatibility
         params = kwargs.pop('params', None)
         if params is None:
             params = {}
-        params.setdefault('database_', self._database)
+
+        # `database_` must be forwarded as a driver-level kwarg on Neo4j's
+        # execute_query — not as a Cypher parameter, where it is silently
+        # ignored and every query routes to the connection's home database.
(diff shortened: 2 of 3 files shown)


gauravch-code:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: gaurav.pvt25@gmail.com

mehulp93:

Please note:

PR #1777, opened 2026-08-18
feat: add PostGraph driver — PostgreSQL-backed graph backend

crajah:
Closes #1781

## PostGraph Driver — PostgreSQL-backed graph backend for Graphiti

This PR adds a new graph driver that enables Graphiti to run entirely on PostgreSQL, powered by the **[post-graph](https://pypi.org/project/post-graph/)** library. This is a key component of the **post-graph-rag** initiative — bringing production-grade, PostgreSQL-native graph RAG to the Graphiti ecosystem.

### Why post-graph-rag?

Most graph RAG systems require a dedicated graph database (Neo4j, FalkorDB, etc.), adding operational complexity. **post-graph-rag** eliminates that by running graph memory directly on PostgreSQL — the database most teams already have in production.

With this driver, teams can:
- **Run Graphiti on PostgreSQL** — no separate graph database to deploy, monitor, or scale
- **Leverage existing Postgres infrastructure** — connection pooling, backups, replication, and monitoring all carry over
- **Use pgvector for embeddings** — similarity search runs natively alongside graph traversal
- **Multi-tenant by default** — post-graph's `realm` maps to Graphiti's `group_id`, providing isolation without schema duplication

### The post-graph library

**[post-graph](https://pypi.org/project/post-graph/)** (`post-graph>=0.6.0`) is a PostgreSQL-backed graph database library that provides:
- `AsyncPostGraph` client with connection lifecycle management
- Vertex and edge tables with `realm`-based multi-tenancy
- JSONB `payload` columns for flexible schema storage
- `embedding` vector columns via pgvector for similarity search
- Graph traversal with configurable depth and direction
- CRUD operations: `add_vertex`, `upsert_vertex`, `add_edge`, `upsert_edge`, `get_vertices`, `delete_vertex`, `delete_realm`
- Vector search: `vector_search()` and `vector_search_edges()` with cosine distance
- Traversal: `traverse()` for BFS walks across edge tables

**PyPI**: https://pypi.org/project/post-graph/
**Source**: https://github.com/crajah/post-graph

### Architecture

All database operations go through post-graph's `AsyncPostGraph` client. There is **zero direct asyncpg or raw PostgreSQL access** in the operation layer:

| Operation | post-graph API used |
|---|---|
| Schema creation | `create_vertex_table()`, `create_edge_table()` |
| Node/Edge CRUD | `upsert_vertex()`, `upsert_edge()`, `get_vertices()`, `delete_vertex()` |
[...]

PR #1782, opened 2026-08-19
fix(embedder): pin encoding_format=float on OpenAI embedding calls

crajah:
## Problem

`OpenAIEmbedder` calls `embeddings.create()` without specifying `encoding_format`. Left unset, the OpenAI SDK negotiates **base64** on its own — and gateways fronting non-OpenAI providers reject the parameter outright:

```
litellm.UnsupportedParamsError: vertex_ai does not support parameters:
{'encoding_format': 'base64'}, for model=gemini-embedding-001
```

Every embedding call fails, so ingestion cannot proceed at all behind such a gateway. Hit in practice with LiteLLM proxying to Vertex AI for `gemini-embedding-001`.

## Fix

Pin `encoding_format='float'` on both `create()` and `create_batch()`. It is the API default and the portable choice, so this is a no-op against OpenAI itself while unblocking OpenAI-compatible gateways.

## Notes

- Split out of #1777 deliberately: that PR adds a driver, whereas this touches shared core code affecting **every** backend, and should be reviewable — or droppable — on its own.
- Leaving it to callers is not reachable today: `OpenAIEmbedderConfig` exposes no way to pass the parameter through.

### DIFF
--- a/graphiti_core/embedder/openai.py
+++ b/graphiti_core/embedder/openai.py
@@ -55,12 +55,19 @@ class OpenAIEmbedder(EmbedderClient):
         self, input_data: str | list[str] | Iterable[int] | Iterable[Iterable[int]]
     ) -> list[float]:
         result = await self.client.embeddings.create(
-            input=input_data, model=self.config.embedding_model
+            input=input_data, model=self.config.embedding_model,
+            encoding_format='float',
         )
         return result.data[0].embedding[: self.config.embedding_dim]
 
     async def create_batch(self, input_data_list: list[str]) -> list[list[float]]:
+        # Pinned explicitly: left unset, the OpenAI SDK negotiates 'base64',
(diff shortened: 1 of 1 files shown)


linhongyu510 (approved):
Verified the current head. Both `create()` and `create_batch()` now pass `encoding_format="float"` explicitly, while preserving model/input handling and embedding-dimension truncation. This matches the OpenAI embeddings API contract and avoids SDK-side base64 negotiation for compatible gateways. Local validation: `pytest tests/embedder/test_openai.py -q` — 2 passed; an additional focused mock check confirmed the keyword on both call paths; Ruff check/format and `git diff --check` passed. No blocking issue found. Non-blocking: the existing tests do not assert `encoding_format`, so adding those assertions would make the regression coverage durable.

PR #1784, opened 2026-08-20
fix(dedupe): make edge contradiction reasoning explicit

### LINKED ISSUE #1666: Contradiction detection in EdgeDuplicate (dedupe judge) collapses on non-reasoning small_models — a reasoning-first response schema restores it (measured)

Whxuan0701:
Closes #1666

## Summary
- Add a reasoning field to the `EdgeDuplicate` structured response before the final index arrays.
- Require prompt reasoning to cite continuous fact indices before returning duplicate/contradiction lists.
- Keep the field optional for backward compatibility with providers that omit it.

## Design
Structured-output field order gives non-reasoning small models an explicit place to analyze facts before emitting indices. The existing duplicate and contradiction arrays remain unchanged, so downstream validation and invalidation behavior is preserved.

## Tests
- `uv run --with pytest --with pytest-asyncio pytest --confcutdir=tests/prompts tests/prompts/test_dedupe_edges.py tests/utils/maintenance/test_edge_operations.py -q` (11 passed)
- Ruff check and format check on changed files.

## Compatibility/Risks
Additive response-model field with an empty default. This improves the prompt/schema contract but cannot guarantee model accuracy for every provider or model. No external LLM calls were required.

### DIFF
--- a/graphiti_core/prompts/dedupe_edges.py
+++ b/graphiti_core/prompts/dedupe_edges.py
@@ -22,6 +22,13 @@ from .models import Message, PromptFunction, PromptVersion
 
 
 class EdgeDuplicate(BaseModel):
+    reasoning: str = Field(
+        default='',
+        description=(
+            'Briefly reason about duplicates and contradictions before the final arrays. '
+            'For every selected fact, cite its continuous idx number.'
+        ),
+    )
     duplicate_facts: list[int] = Field(
         ...,
         description='List of idx values of duplicate facts (only from EXISTING FACTS range). Empty list if none.',
@@ -81,6 +88,11 @@ EXISTING FACTS are indexed first, followed by FACT INVALIDATION CANDIDATES.
    - A fact from EXISTING FACTS can be both a duplicate AND contradicted (e.g., semantically the same but the new fact updates/supersedes it).
    - Return all contradicted idx values in contradicted_facts.
    - If no contradictions, return an empty list for contradicted_facts.
+
+Before returning the arrays, explain your decision in a short reasoning field. '
+For every duplicate or contradiction, explicitly cite its idx number in the reasoning '
+(for example, "idx 2 is contradicted because the value changed"). Then return the '
+duplicate_facts and contradicted_facts arrays using those same idx values.
 
 <EXAMPLE>
 EXISTING FACT: idx=0, "Alice joined Acme Corp in 2020"
--- a/tests/prompts/test_dedupe_edges.py
+++ b/tests/prompts/test_dedupe_edges.py
@@ -1,0 +1,25 @@
+from graphiti_core.prompts.dedupe_edges import EdgeDuplicate, resolve_edge
+
+
+def test_edge_duplicate_schema_collects_reasoning_before_indices():
+    fields = list(EdgeDuplicate.model_fields)
+
+    assert fields[:3] == ['reasoning', 'duplicate_facts', 'contradicted_facts']
+    assert EdgeDuplicate(
+        reasoning='idx 1 contradicts the new fact',
+        duplicate_facts=[],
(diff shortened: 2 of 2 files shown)


linhongyu510 (changes requested):
The new `reasoning` field is optional (`default=""`), so the raw Pydantic schema does not include it in `required`. That undercuts the mechanism this PR and #1666 rely on: `OpenAIGenericClient` and `AnthropicClient` pass `model_json_schema()` directly, and Gemini likewise consumes the model/schema, so those providers may legally emit the two arrays without first generating reasoning. The issue’s measured variant used `Field(...)`; the new test checks only declaration order and therefore misses this. Please make `reasoning` required and assert that `reasoning` is present in `EdgeDuplicate.model_json_schema()["required"]` (ideally also cover a provider-facing schema path). I reproduced the current schema on `8f559931`: `required == ["duplicate_facts", "contradicted_facts"]`, while the focused 11 tests still pass. Also please remove the stray literal apostrophes embedded at the ends of the new prompt lines (`reasoning field. '`, `reasoning '`, `the '`); the current test only checks substrings and does not catch the malformed prompt text.

PR #1785, opened 2026-08-20
fix(nodes): persist and restore episodic metadata

### LINKED ISSUE #1769: [Bug] EpisodicNode.episode_metadata is documented as public surface but is never persisted or read -- always None
## Summary

`EpisodicNode.episode_metadata` is declared on the model (`graphiti_core/nodes.py:329-332`) and was listed in the 0.29.0 release notes as new public surface, alongside `Graphiti.summarize_saga()` and the `fact_triple` episode type. It has no wiring behind it: it is dropped before the database write, never populated on read, and has no input path from any public API. Every episode returned by `retrieve_episodes()` has `episode_metadata: None` regardless of what a caller does.

A repo-wide search for the literal string `episode_metadata` returns exactly one hit — the field declaration itself.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Objective

Either wire `episode_metadata` end-to-end (save, read, and an `add_episode()` parameter), or remove it from the model and correct the 0.29.0 release notes. The current state advertises a feature that silently does nothing.

## What I was trying to do

Attach caller-supplied structured metadata (source document path and vault identifier) to episodes, so a published graph replica could cite source documents without a side-channel lookup file. The 0.29.0 release notes describe `EpisodicNode.episode_metadata` as being for exactly this kind of custom filtering key.

## What I expected

Either a parameter on `add_episode()` accepting the metadata, or the ability to set `episode_metadata` on an `EpisodicNode` and have it round-trip through save and retrieve.

## What actually happened

No input path exists, and any value set on the model is discarded before the write.

## Root cause

**1. Declared on the model** — `graphiti_core/nodes.py:329-332`:

```python
episode_metadata: dict[str, Any] | None = Field(
    description='customer-defined metadata key-value pairs for filtering',
    default=None,
)
```

**2. Dropped before write** — `graphiti_core/nodes.py:341-354`, `EpisodicNode.save()` builds `episode_args` with nine keys (`uuid`, `name`, `group_id`, `source_description`, `content`, `entity_edges`, `created_at`, `valid_at`, `source`). `episode_metadata` is absent, so it is never passed as a query parameter.

**3. Not writable even if it were passed** — `graphiti_core/models/nodes/node_db_queries.py:30-66`, `get_episode_node_save_query()` hardcodes an explicit `SET n = {...}` naming the same nine properties, in all four provider branches (Neo4j, FalkorDB, Kuzu, Neptune). This differs from `EntityNode`, whose save query (`node_db_queries.py:181-190`) spreads a caller-built dict via `SET n = $entity_data` and therefore does support arbitrary properties.

**4. Not read back** — `EPISODIC_NODE_RETURN` (`node_db_queries.py:112-122`) returns the same nine properties, and `get_episodic_node_from_record()` (`nodes.py:1028-1047`) constructs the `EpisodicNode` without passing `episode_metadata`, so it always falls back to the field default of `None`.

**5. No input path** — `add_episode()` (`graphiti_core/graphiti.py:980-998`) has no `metadata` or `episode_metadata` parameter, and does not reference the field anywhere in its body. The bulk-add path does not either.

## Reproduction

Requires only a database and a single episode:

```python
import asyncio, os

async def main():
    g = Graphiti(
        uri=os.environ['REPRO_NEO4J_URI'],
        user=os.environ['REPRO_NEO4J_USER'],
        password=os.environ['REPRO_NEO4J_PASSWORD'],
    )
    await g.add_episode(
        name='repro-episode',
        episode_body='A short episode used to inspect stored properties.',
        source_description='repro',
        group_id='repro-episode-metadata',
    )
    episodes = await g.retrieve_episodes(
        group_ids=['repro-episode-metadata'],
        last_n=1,
    )
    print('episode_metadata:', episodes[0].episode_metadata)
    records, _, _ = await g.driver.execute_query(
        "MATCH (e:Episodic {group_id: 'repro-episode-metadata'}) RETURN keys(e) AS k"
    )
    print('stored property keys:', records[0]['k'])
    await g.close()

asyncio.run(main())
```

Observed against graphiti-core 0.29.3, Neo4j 5.26-community:

```
episode_metadata: None
stored property keys: ['created_at', 'group_id', 'uuid', 'valid_at', 'name', 'content', 'source', 'source_description', 'entity_edges']
```

There is no way to make that first line print anything else through the public API — the field has no setter path, and a hand-constructed `EpisodicNode` with the field set loses it at `save()`.

## Testing

A fix that wires the field should be validated by setting metadata on episode creation, retrieving the episode, and asserting the value round-trips — plus confirming the property is present in `keys(e)` on the stored node. A fix that removes the field instead should be accompanied by a release-notes correction, since 0.29.0 announced it publicly.

## Breaking Changes

None for wiring it — the field already exists with a `None` default, and adding an optional `add_episode()` parameter is additive. Removing the field would be breaking for anyone reading it, though today it can only ever return `None`.

The mechanism already exists in this codebase for the other node type: `EntityNode` spreads a caller-built dict into `entity_data` and writes `SET n = $entity_data` (`node_db_queries.py:181-190`), reads it back with `properties(n) AS attributes` (`node_db_queries.py:283-291`), and pops the reserved field names in `graphiti_core/driver/record_parsers.py`. Applying that same round-trip to `EpisodicNode` — with the nine existing fields as its reserved set — would make `episode_metadata` behave consistently with entity attributes rather than introducing a new pattern. Worth noting only that `dict[str, Any]` permits nested dicts and lists-of-dicts, which Neo4j cannot store as node properties at all; entity attributes are already under that same constraint, so whatever handling they use applies here too.

## Checklist

- [x] Reproduced against the installed 0.29.3; fork and installed package have identical line numbers at every cited location
- [x] Confirmed by direct database inspection that the property is absent on stored nodes
- [x] Root cause traced through all four provider save branches and the read path
- [ ] Fix implemented (not in scope for this issue — filed to establish the record before a PR)

## Related Issues

Introduced in PR #1422 ("forward-port graphiti_core improvements from internal fork") and announced in PR #1444's 0.29.0 release notes. No existing issue found reporting that the field is non-functional.

Whxuan0701:
Closes #1769

## Summary
- Add an `episode_metadata` input to `Graphiti.add_episode()` and the bulk path.
- Persist metadata across Neo4j, FalkorDB, Neptune, and Kuzu save operations.
- Restore metadata from read paths with reserved-field filtering and add regression coverage.

## Design
Neo4j/FalkorDB/Neptune flatten caller metadata into top-level node properties, matching `EntityNode.attributes`; Kuzu stores JSON because it requires fixed properties. Reads reconstruct the model field without leaking reserved episode fields.

## Tests
- `uv run --with pytest --with pytest-asyncio pytest --confcutdir=tests tests/test_episode_metadata.py -q` (2 passed)
- Ruff check and format check on changed files.

## Compatibility/Risks
Additive public parameter with `None` default. Metadata must obey the existing backend property-value constraints; nested values unsupported by a backend may still be rejected. Live database integration tests were not run because Neo4j/FalkorDB/Neptune/Kuzu services were unavailable.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/episode_node_ops.py
+++ b/graphiti_core/driver/falkordb/operations/episode_node_ops.py
@@ -52,6 +52,10 @@ class FalkorEpisodeNodeOperations(EpisodeNodeOperations):
             'valid_at': node.valid_at,
             'source': node.source.value,
         }
+        episode_data = dict(params)
+        for key, value in (node.episode_metadata or {}).items():
+            episode_data.setdefault(key, value)
+        params['episode_data'] = episode_data
         if tx is not None:
             await tx.run(query, **params)
         else:
@@ -71,6 +75,9 @@ class FalkorEpisodeNodeOperations(EpisodeNodeOperations):
             ep = dict(node)
             ep['source'] = str(ep['source'].value)
             ep.pop('labels', None)
+            metadata = ep.pop('episode_metadata', None) or {}
+            for key, value in metadata.items():
+                ep.setdefault(key, value)
             episodes.append(ep)
 
         query = get_episode_node_save_bulk_query(GraphProvider.FALKORDB)
--- a/graphiti_core/driver/kuzu/operations/episode_node_ops.py
+++ b/graphiti_core/driver/kuzu/operations/episode_node_ops.py
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 
+import json
 import logging
 from datetime import datetime
 from typing import Any
@@ -50,6 +51,7 @@ class KuzuEpisodeNodeOperations(EpisodeNodeOperations):
             'created_at': node.created_at,
             'valid_at': node.valid_at,
             'source': node.source.value,
+            'episode_metadata': json.dumps(node.episode_metadata or {}),
         }
         if tx is not None:
             await tx.run(query, **params)
--- a/graphiti_core/driver/neo4j/operations/episode_node_ops.py
+++ b/graphiti_core/driver/neo4j/operations/episode_node_ops.py
@@ -52,6 +52,10 @@ class Neo4jEpisodeNodeOperations(EpisodeNodeOperations):
             'valid_at': node.valid_at,
             'source': node.source.value,
         }
+        episode_data = dict(params)
+        for key, value in (node.episode_metadata or {}).items():
+            episode_data.setdefault(key, value)
+        params['episode_data'] = episode_data
         if tx is not None:
             await tx.run(query, **params)
         else:
@@ -71,6 +75,9 @@ class Neo4jEpisodeNodeOperations(EpisodeNodeOperations):
             ep = dict(node)
             ep['source'] = str(ep['source'].value)
             ep.pop('labels', None)
+            metadata = ep.pop('episode_metadata', None) or {}
+            for key, value in metadata.items():
+                ep.setdefault(key, value)
             episodes.append(ep)
 
         query = get_episode_node_save_bulk_query(GraphProvider.NEO4J)
--- a/graphiti_core/driver/neptune/operations/episode_node_ops.py
+++ b/graphiti_core/driver/neptune/operations/episode_node_ops.py
@@ -52,6 +52,11 @@ class NeptuneEpisodeNodeOperations(EpisodeNodeOperations):
             'valid_at': node.valid_at,
             'source': node.source.value,
         }
+        episode_data = dict(params)
+        episode_data['entity_edges'] = '|'.join(str(value) for value in node.entity_edges)
+        for key, value in (node.episode_metadata or {}).items():
+            episode_data.setdefault(key, value)
+        params['episode_data'] = episode_data
         if tx is not None:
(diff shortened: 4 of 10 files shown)


linhongyu510 (changes requested):
Kuzu support is currently broken at runtime. The patch writes  in , but  in  still defines  without that property. I reproduced this on the PR head with Kuzu 0.11.3 by creating  and saving an  with metadata; Kuzu raises  The two added tests use a mocked Neo4j driver and do not exercise any backend schema, so they pass while this advertised backend fails. Please add the Kuzu column/schema migration (including an upgrade path for existing databases) and a Kuzu round-trip regression test before merging.

linhongyu510:
Clarification for review 5060824232 (some inline identifiers were stripped while posting): the failing assignment is n.episode_metadata in get_episode_node_save_query(GraphProvider.KUZU), while SCHEMA_QUERIES in graphiti_core/driver/kuzu_driver.py defines Episodic with entity_edges as its final property and no episode_metadata column. The exact reproduced error is: Binder exception: Cannot find property episode_metadata for n.

gkorland:

PR #1797, opened 2026-08-25
fix(anthropic): omit temperature instead of sending null when it is unset (#1794)

### LINKED ISSUE #1794: AnthropicClient sends `temperature: null` when temperature is unset (400 from API) — and the MCP queue then silently drops every episode
## Summary

With the MCP server configured for `llm.provider: anthropic` and `temperature`
left unset (the schema default is `None`), every extraction request fails with
HTTP 400 from the Anthropic API:

```
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': 'temperature: Input should be a valid number'}}
```

Because the episode queue catches all exceptions, logs them, and marks the task
done, **every `add_memory` call still returns success** — we submitted a
20-document batch that reported "queued for processing" 20 times and landed 0
episodes. Only comparing `get_episodes` counts against submissions revealed it.

## Environment

- graphiti-core 0.29.2 (also present in current `main` @ `993e081a`)
- mcp_server from the same checkout (package `mcp-server` 1.0.2)
- anthropic SDK 0.111.0, Python 3.10.21, FalkorDB backend
- Model: `claude-haiku-4-5-20251001`

## Reproduction

1. `config.yaml` with `llm.provider: anthropic`, a valid `ANTHROPIC_API_KEY`,
   and **no `temperature` key** (schema default `None`).
2. Call the `add_memory` MCP tool with any episode.
3. Tool returns `Episode ... queued for processing`; the server log shows the
   400 above, three retries, then
   `services.queue_service - ERROR - Error processing queued episode ...`.
4. `get_episodes` for the group: the episode never lands.

Sample failing request IDs (2026-08-24): `req_011CeMSZvQydGf5GLXqoYTGN`,
`req_011CeMSZwJouHufv2wPfzPDT`, `req_011CeMSZy2zEZLgDLdP5N9uL`.

## Cause

`mcp_server/src/config/schema.py` defaults `temperature` to `None`
("defaults to None for reasoning models"). The factory comment says downstream
clients omit temperature when it is `None` — the OpenAI clients do, but
`AnthropicClient` passes it through to `messages.create(...)` verbatim, and the
API rejects `temperature: null`.

## Suggested fix

In `AnthropicClient`, omit the `temperature` parameter when
`config.temperature is None` (mirroring the OpenAI clients), or use the SDK's
`NOT_GIVEN` sentinel.

Separately but compounding: consider surfacing per-episode failures somewhere a
caller can see them (e.g. failure counters in `get_status`, or a failed-episode
list). A queue that converts hard failures into log lines while the tool call
reports success makes this class of bug invisible to MCP clients.

## Workaround

Set `temperature: 0.0` explicitly in the `llm` section of `config.yaml`.

Anai-Guo:
Fixes #1794.

## Problem

`mcp_server/src/config/schema.py` defaults `llm.temperature` to `None`, and
`mcp_server/src/services/factories.py` passes it straight through to the core
`LLMConfig` — with the same comment repeated at all five provider branches:

```python
# None is intentional for reasoning models; core LLMConfig stores it
# verbatim and downstream clients omit temperature when it is None.
temperature=config.temperature,  # type: ignore[arg-type]
```

`OpenAIClient` honors that contract (`openai_client.py`):

```python
# Omit temperature entirely for reasoning models — don't even send None.
if not is_reasoning_model and temperature is not None:
    request_kwargs['temperature'] = temperature
```

`AnthropicClient` did not — it forwarded `self.temperature` to
`messages.create()` unconditionally. The Anthropic SDK serializes an explicit
`None` rather than dropping the key, so the request body carries
`"temperature": null` and the API rejects it:

```
400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
       'message': 'temperature: Input should be a valid number'}}
```

Which I confirmed at the wire level against a mocked httpx transport
(anthropic 0.120.2):

| call | key on the wire | value |
|---|---|---|
| `messages.create(..., temperature=None)` | present | `null` |
| `messages.create(...)` (omitted) | **absent** | — |
| `messages.create(..., temperature=0.0)` | present | `0.0` |

Under the MCP server this is invisible to callers: the episode queue catches the
exception, logs it and marks the task done, so every `add_memory` still reports
success while nothing lands.

## Fix

Build the `messages.create()` kwargs and add `temperature` only when it is not
`None`, mirroring `OpenAIClient._create_completion`. The guard is `is not None`,
not truthiness, so an explicit `temperature=0.0` is still sent.

## Verification

Reproduction against the real `AnthropicClient` with a mocked
`messages.create`, before → after:

| config | before | after |
|---|---|---|
| `temperature` unset (`None`) | sent as `None` | **omitted** |
| `temperature=0.0` | sent as `0.0` | sent as `0.0` |
| `temperature=0.7` | sent as `0.7` | sent as `0.7` |

Tests (`tests/llm_client/test_anthropic_client.py`, new
`TestAnthropicClientTemperature`) cover all three rows.

- Full file: **11 passed** before, **14 passed** after.
- Red/green: reverting only `anthropic_client.py` while keeping the new tests
  fails **exactly** `test_temperature_omitted_when_unset` (13 passed, 1 failed).
  The two guard cases stay green on both sides, which is what they are for.
- `pytest tests/llm_client tests/test_text_utils.py tests/utils`: **297 passed,
  2 skipped** (the skips are the `_int` tests, no `ANTHROPIC_API_KEY`).
- `ruff==0.14.11` (the version pinned in `.github/workflows/lint.yml`)
  `check` and `format --check` clean on both files.

## Deliberately not in scope

The issue reports two things; this PR fixes only the first.

- **`OpenAIGenericClient`, `GeminiClient`, `GroqClient`** also pass
  `temperature=self.temperature` through unconditionally, so the factory's
  "downstream clients omit temperature when it is `None`" comment is not true
  for them either. Whether that is actually a bug depends on how each of those
  APIs treats an explicit `null`, which I have not verified the way I verified
  Anthropic's — so I left them alone rather than guess. Happy to follow up if
  you'd like them covered.
- **Surfacing per-episode queue failures** (the issue's second half — a queue
  that turns hard failures into log lines while the tool call reports success).
  That is an MCP server API-shape decision, not a bug fix.

### DIFF
--- a/graphiti_core/llm_client/anthropic_client.py
+++ b/graphiti_core/llm_client/anthropic_client.py
@@ -285,15 +285,21 @@ class AnthropicClient(LLMClient):
         try:
             # Create the appropriate tool based on whether response_model is provided
             tools, tool_choice = self._create_tool(response_model)
-            result = await self.client.messages.create(
-                system=system_message.content,
-                max_tokens=max_creation_tokens,
-                temperature=self.temperature,
-                messages=user_messages_cast,
-                model=self.model,
-                tools=tools,
-                tool_choice=tool_choice,
-            )
+            create_kwargs: dict[str, typing.Any] = {
+                'system': system_message.content,
+                'max_tokens': max_creation_tokens,
+                'messages': user_messages_cast,
+                'model': self.model,
+                'tools': tools,
+                'tool_choice': tool_choice,
+            }
+            # Omit temperature entirely when it is unset — don't even send None.
+            # The Anthropic SDK serializes an explicit None as `"temperature": null`,
+            # which the API rejects with a 400 invalid_request_error.
+            if self.temperature is not None:
+                create_kwargs['temperature'] = self.temperature
+
+            result = await self.client.messages.create(**create_kwargs)
 
             # Extract token usage from the response
             input_tokens = 0
--- a/tests/llm_client/test_anthropic_client.py
+++ b/tests/llm_client/test_anthropic_client.py
@@ -251,5 +251,67 @@ class TestAnthropicClientGenerateResponse:
         assert result['test_field'] == 'correct_value'
 
 
+class TestAnthropicClientTemperature:
+    """Tests for how the temperature parameter is forwarded to the Anthropic API."""
+
+    @staticmethod
+    def _client_with_temperature(temperature):
+        """Build an AnthropicClient with a mocked messages.create returning a tool_use."""
+        config = LLMConfig(
+            api_key='test_api_key',
+            model='claude-haiku-4-5-latest',
+            temperature=temperature,
+            max_tokens=1000,
+        )
+        client = AnthropicClient(config=config, cache=False)
+
+        content_item = MagicMock()
+        content_item.type = 'tool_use'
+        content_item.input = {'test_field': 'test_value'}
+
+        mock_response = MagicMock()
+        mock_response.content = [content_item]
+
+        mock_client = MagicMock()
+        mock_client.messages.create = AsyncMock(return_value=mock_response)
+        client.client = mock_client
+        return client, mock_client
+
+    @pytest.mark.asyncio
+    async def test_temperature_omitted_when_unset(self):
+        """An unset temperature must not be sent at all.
+
+        The Anthropic API rejects `"temperature": null` with a 400
+        invalid_request_error, and the SDK serializes an explicit ``None``
+        instead of dropping the key.
+        """
+        client, mock_client = self._client_with_temperature(None)
+
+        await client.generate_response([Message(role='user', content='Test message')])
+
+        kwargs = mock_client.messages.create.call_args.kwargs
+        assert 'temperature' not in kwargs
+
+    @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


linhongyu510 (approved):
Verified at `6e51b5c`. The change is correctly limited to request construction: `temperature` is omitted only for `None`, while explicit `0.0` and `0.7` remain present, with no changes to retries, token accounting, tool parsing, or exception mapping. Independent validation: the full Anthropic client module passes 14 tests; `tests/llm_client tests/test_text_utils.py tests/utils` passes 319 tests with 2 expected integration skips; Ruff check/format and `git diff --check` pass. I also performed a red/green check by reverting only `anthropic_client.py` while retaining the new test: `test_temperature_omitted_when_unset` fails exactly because the baseline still forwards the key, then all three temperature tests pass again on this HEAD. The separate queue-failure observability concern is appropriately left out of this minimal fix.

Anai-Guo:
Friendly ping @linhongyu510 — thanks for the approval. Is there anything else needed before this can be merged? Still `MERGEABLE` against main; happy to rebase if it has drifted.

PR #1798, opened 2026-08-25
fix(graphiti): route group writes per call instead of rebinding the shared client (#1795)

### LINKED ISSUE #1795: FalkorDB: writes mutate the shared driver (`self.driver = self.driver.clone(database=group_id)`) — reads then answer from the *last-written* group's graph, silently returning empty for other groups

Anai-Guo:
Fixes #1795.

## The bug

On providers that keep each `group_id` in its own database (FalkorDB), `add_episode` and `add_episode_bulk` routed the write by reassigning the **shared** client:

```python
```

That reroutes the `Graphiti` instance, not just the write. On a long-lived multi-group client — the MCP server is exactly that — every subsequent read runs inside whichever group wrote last, carrying a `group_id` property filter that matches nothing, so it **returns empty with no error**. Scoped memory looks correct until a second group writes.

## The fix

Bind the routed driver and clients to locals. This is the idiom the file already uses everywhere else — `retrieve_episodes` (`graphiti.py:967`) and `build_communities` (`graphiti.py:1500`) both take an optional `driver` and fall back to `self.clients.driver` locally. `add_episode` / `add_episode_bulk` were the two outliers that mutated instead.

Two small helpers carry it:

- `_driver_for_group(group_id)` — returns `self.driver` when it is already on that database, otherwise a clone. Never assigns.
- `_clients_for_driver(driver)` — returns `self.clients` unchanged when the driver is the base one, otherwise a `model_copy(update={'driver': driver})`. `GraphitiClients` is a pydantic model, so the shared model is untouched and the other clients (llm, embedder, cross-encoder, tracer) stay shared rather than rebuilt.

Two details worth calling out, because both are easy to get wrong:

1. **The previous-episode lookup inside `add_episode` now passes the routed driver explicitly.** It would otherwise fall back to the unrouted `self.clients.driver` and read from the wrong database — the same class of bug, just moved.

2. **Routing still happens only when the caller supplies a `group_id`.** `get_default_group_id` returns `'_'` on FalkorDB and `''` elsewhere — never the driver's own database name — so routing on the *resolved* default would have started cloning where the original code deliberately did not. There is a test pinning this.

## Not in scope

The second problem described in #1795: per-group graphs make a `group_ids` **list** unsatisfiable, since a search with `group_ids=["a","b"]` can only ever execute inside one graph. Fixing that needs a fan-out read path plus a decision about cross-group result merging — a larger design change than this, and worth its own discussion.

## Verification

`tests/test_group_routing_does_not_mutate_client.py` uses `spec=` mocks, so it needs no live FalkorDB and runs in CI regardless of whether `falkordb` is installed — same approach as the existing `tests/driver/test_falkordb_ops_routing.py`.

Verified red/green against unpatched `main` (`993e081`):

| | unpatched `main` | with this change |
|---|---|---|
| `tests/test_group_routing_does_not_mutate_client.py` | **8 failed**, 1 passed | **9 passed** |

The one test that passes on both is `test_add_episode_without_a_group_id_does_not_clone` — it pins the `group_id=None` behaviour that this change preserves, so it is expected to be green on either side.

`test_add_episode_does_not_rebind_the_shared_driver` goes through the public API rather than the helpers: it makes the tracing span raise, which aborts `add_episode` immediately after routing, so the only thing left to observe is what routing did to the shared client. No LLM, embedder or database work involved.

Also run: `tests/driver`, `tests/utils`, `tests/test_graphiti_mock.py` (`-k "not _int"`) — all pass. The remaining errors in those directories are the live-database fixtures, and they error identically on unpatched `main`. `ruff format` and `ruff check` are clean on both changed files.

### DIFF
--- a/graphiti_core/graphiti.py
+++ b/graphiti_core/graphiti.py
@@ -344,7 +344,11 @@ class Graphiti:
         await self.driver.close()
 
     async def _get_or_create_saga(
-        self, saga_name: str, group_id: str, created_at: datetime
+        self,
+        saga_name: str,
+        group_id: str,
+        created_at: datetime,
+        driver: GraphDriver | None = None,
     ) -> SagaNode:
         """
         Get an existing saga by name or create a new one.
@@ -360,6 +364,11 @@ class Graphiti:
             originating episode's reference time (``valid_at``) rather than the
             current wall-clock time so the saga's ``created_at`` reflects the
             episode it was minted from.
+        driver : GraphDriver | None
+            Optional. Driver to run the saga lookup/creation against. Defaults to
+            ``self.driver``. Callers that route a write at a specific ``group_id``
+            must pass the routed driver so the saga lands in the same graph as the
+            episode it belongs to.
 
         Returns
         -------
@@ -368,7 +377,9 @@ class Graphiti:
         """
         from graphiti_core.helpers import parse_db_date
 
-        records, _, _ = await self.driver.execute_query(
+        driver = driver if driver is not None else self.driver
+
+        records, _, _ = await driver.execute_query(
             """
             MATCH (s:Saga {name: $name, group_id: $group_id})
             RETURN s.uuid AS uuid, s.name AS name, s.group_id AS group_id, s.created_at AS created_at
@@ -388,22 +399,31 @@ class Graphiti:
             )
 
         saga = SagaNode(name=saga_name, group_id=group_id, created_at=created_at)
-        await saga.save(self.driver)
+        await saga.save(driver)
         return saga
 
     async def _saga_get_previous_episode_uuid(
-        self, saga_uuid: str, current_episode_uuid: str
+        self,
+        saga_uuid: str,
+        current_episode_uuid: str,
+        driver: GraphDriver | None = None,
     ) -> str | None:
-        """Find the most recent episode UUID in a saga, excluding the current one."""
-        if self.driver.graph_operations_interface:
+        """Find the most recent episode UUID in a saga, excluding the current one.
+
+        ``driver`` defaults to ``self.driver``; routed writes must pass their own
+        driver so the lookup reads the graph the episode is being written to.
+        """
+        driver = driver if driver is not None else self.driver
+
+        if driver.graph_operations_interface:
             try:
-                return await self.driver.graph_operations_interface.saga_get_previous_episode_uuid(
-                    self.driver, saga_uuid, current_episode_uuid
+                return await driver.graph_operations_interface.saga_get_previous_episode_uuid(
+                    driver, saga_uuid, current_episode_uuid
                 )
             except NotImplementedError:
                 pass
 
-        records, _, _ = await self.driver.execute_query(
+        records, _, _ = await driver.execute_query(
             """
             MATCH (s:Saga {uuid: $saga_uuid})-[:HAS_EPISODE]->(e:Episodic)
             WHERE e.uuid <> $current_episode_uuid
@@ -639,8 +659,13 @@ class Graphiti:
         nodes: list[EntityNode],
         uuid_map: dict[str, str],
         custom_extraction_instructions: str | None = None,
+        clients: GraphitiClients | None = None,
     ) -> tuple[list[EntityEdge], list[EntityEdge], list[EntityEdge]]:
         """Extract edges from episode(s) and resolve against existing graph.
+
+        ``clients`` defaults to ``self.clients``; routed writes must pass the
+        clients bound to their per-call driver so extraction and resolution read
+        the same graph the episode is written to.
 
         Returns
         -------
@@ -650,11 +675,12 @@ class Graphiti:
             - invalidated_edges: Edges invalidated by new information
             - new_edges: Only edges that are new to the graph (not duplicates)
         """
+        clients = clients if clients is not None else self.clients
         episodes = episode if isinstance(episode, list) else [episode]
         primary_episode = episodes[0]
 
         extracted_edges = await extract_edges(
-            self.clients,
+            clients,
             episode,
             extracted_nodes,
             previous_episodes,
@@ -667,7 +693,7 @@ class Graphiti:
         edges = resolve_edge_pointers(extracted_edges, uuid_map)
 
         resolved_edges, invalidated_edges, new_edges = await resolve_extracted_edges(
-            self.clients,
+            clients,
             edges,
             primary_episode,
             nodes,
@@ -687,6 +713,7 @@ class Graphiti:
         saga: str | SagaNode | None = None,
         saga_previous_episode_uuid: str | None = None,
         node_episode_index_map: dict[str, list[int]] | None = None,
+        driver: GraphDriver | None = None,
     ) -> tuple[list[EpisodicEdge], EpisodicNode]:
         """Process and save episode data to the graph.
 
@@ -713,7 +740,15 @@ class Graphiti:
         node_episode_index_map : dict[str, list[int]] | None
             Optional mapping from node UUID to 0-indexed episode positions for
             building episodic edges with correct attribution.
+        driver : GraphDriver | None
+            Optional. Driver every graph write in this method is issued against.
+            Defaults to ``self.driver``. Callers that route a write at a specific
+            ``group_id`` must pass the routed driver, otherwise the episode, its
+            episodic edges, the resolved nodes and entity edges, and any saga
+            relations are persisted to the base graph while the reads that
+            produced them ran against the routed one.
         """
+        driver = driver if driver is not None else self.driver
         episodes = episode if isinstance(episode, list) else [episode]
         episode_uuids = [ep.uuid for ep in episodes]
 
@@ -724,7 +759,7 @@ class Graphiti:
                 ep.content = ''
 
         await add_nodes_and_edges_bulk(
-            self.driver,
+            driver,
             episodes,
             episodic_edges,
             nodes,
@@ -742,7 +777,9 @@ class Graphiti:
                 # newly created saga so its created_at matches the episode that
                 # minted it, not the wall-clock time of this run.
                 saga_created_at = primary_episode.valid_at or now
-                saga_node = await self._get_or_create_saga(saga, group_id, saga_created_at)
+                saga_node = await self._get_or_create_saga(
+                    saga, group_id, saga_created_at, driver=driver
+                )
             else:
                 saga_node = saga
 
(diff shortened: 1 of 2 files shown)


linhongyu510 (commented):
There is still a routed-write escape in `add_episode`: the early reads/extraction now use the local `driver`, but the final persistence call goes through `_process_episode_data(...)` without passing that driver. `_process_episode_data` then calls `add_nodes_and_edges_bulk(self.driver, ...)`, so for an explicit Falkor group the episode, episodic edges, resolved nodes, and entity edges are written to the shared base graph rather than the group clone. The same helper's saga path also uses `self.driver` for saga lookup/save and NEXT_EPISODE/HAS_EPISODE writes.

I reproduced this on exact head `3266d73e` with a deterministic no-database regression: create the PR's base/child `GraphDriver` mocks, route `group-a`, patch `add_nodes_and_edges_bulk`, then invoke the episode persistence helper. The first save argument is the base driver, not the routed child:

```text
assert save.await_args.args[0] is routed
E assert <base GraphDriver> is <group-a GraphDriver>
1 failed, 9 passed
```

This is more than an untested helper: `add_episode()` calls `_process_episode_data()` after all of the routed reads and does not provide any routing context, so a single operation can read/dedupe in `group-a` and persist its final graph objects in the base database. `add_episode_bulk()` has a related escape when `saga` is set: it calls `_get_or_create_saga()` and `_saga_get_previous_episode_uuid()` without the local `driver`, and those helpers also use `self.driver` internally.

A minimal local fix was to thread an optional `driver` through `_process_episode_data`, `_get_or_create_saga`, and `_saga_get_previous_episode_uuid`; use it for all graph operations; and pass the per-call driver from both write paths. With that change, the added persistence regression plus this PR's nine tests report `10 passed`; targeted Ruff lint/format and `git diff --check` pass.

Please add a regression that observes the driver used by the actual persistence boundary, not only that shared object identities remain unchanged. The current remote checks are Ruff, CLA, and Socket metadata; there is no Python behavior test check on this SHA, so their green status does not cover this split-routing failure.

linhongyu510 (changes requested):
The shared driver no longer mutates, but the routed driver is not propagated through the full write pipeline, so a grouped write can still query or persist in the base database. In `add_episode`, `_extract_and_resolve_edges()` still calls `extract_edges()` and `resolve_extracted_edges()` with `self.clients`, and `_process_episode_data()` still saves the episode/nodes/edges plus Saga relations through `self.driver`. In `add_episode_bulk`, `_extract_and_dedupe_nodes_bulk()` and `_resolve_nodes_and_edges_bulk()` likewise use `self.clients`; Saga lookup/creation helpers also use `self.driver` even though later edges are saved with the local driver. This mixes the routed and base graphs within one operation. The 9 new tests all pass locally, but the public-API tests intentionally abort at `tracer.start_span`, before any of these calls, so they only prove object identity is unchanged, not that the write lands and deduplicates in the requested group. Please thread the local `driver`/`clients` through these helpers (including Saga reads/writes) and add a regression that drives `add_episode` and `add_episode_bulk` through their final save while asserting every graph read/write receives the routed child driver. A two-group FalkorDB integration test would provide the strongest parity evidence.

Anai-Guo:
Thanks @linhongyu510 — you were right on every point, and I reproduced each one before changing anything. Pushed in `03d2862`.

**What was actually broken.** The first commit routed the reads but left the persistence boundary on the shared client, so a single grouped operation could dedupe in `group-a` and then write to the base database:

| site | was | now |
|---|---|---|
| `_process_episode_data` → `add_nodes_and_edges_bulk` | `self.driver` | routed `driver` |
| `_process_episode_data` → NEXT_EPISODE / HAS_EPISODE / saga node saves | `self.driver` | routed `driver` |
| `_get_or_create_saga`, `_saga_get_previous_episode_uuid` | `self.driver` | optional `driver` param |
| `_extract_and_resolve_edges` → `extract_edges`, `resolve_extracted_edges` | `self.clients` | optional `clients` param |
| `_extract_and_dedupe_nodes_bulk` → `extract_nodes_and_edges_bulk`, `dedupe_nodes_bulk` | `self.clients` | optional `clients` param |
| `_resolve_nodes_and_edges_bulk` (3 call sites) | `self.clients` | optional `clients` param |

Every new parameter defaults to `None` and falls back to `self.driver` / `self.clients`, so nothing changes for existing callers. `add_episode` and `add_episode_bulk` now pass the per-call `driver` / `clients` they had already computed.

I checked mechanically rather than by eye — pulling each of the six helpers out with `ast.get_source_segment` and grepping its body for `self.driver` / `self.clients` now reports `NONE` for all six (the only remaining occurrences anywhere in those functions are in the docstrings that explain the defaults).

**Tests at the boundary you asked for**, not at object identity. Four new tests observe which driver the write actually receives:

- `test_process_episode_data_persists_through_the_routed_driver` — asserts `add_nodes_and_edges_bulk`'s first argument *is* the routed child. This is your failing assertion, now green.
- `test_process_episode_data_routes_every_saga_write` — saga lookup, previous-episode lookup, both edge saves and the saga node save all receive the routed driver, and `base.execute_query` is never called.
- `test_extract_and_resolve_edges_uses_the_routed_clients`
- `test_extract_and_dedupe_nodes_bulk_uses_the_routed_clients`

plus `test_process_episode_data_defaults_to_the_shared_driver` as a control group, pinning that the unrouted path still uses `self.driver`.

They are real regressions, not vacuous — on the previous commit `3266d73`:

```
FAILED test_process_episode_data_persists_through_the_routed_driver
FAILED test_process_episode_data_routes_every_saga_write
FAILED test_extract_and_resolve_edges_uses_the_routed_clients
FAILED test_extract_and_dedupe_nodes_bulk_uses_the_routed_clients
4 failed, 10 passed
```

and on `03d2862`:

```
14 passed
```

The control-group test passes on both commits, as it should.

`ruff format --check` and `ruff check` (0.15.17, repo config) are clean on both changed files. `tests/test_graphiti_mock.py` errors identically before and after the change, so that is pre-existing here and not something this commit touched.

On the two-group FalkorDB integration test: agreed that it would be the strongest evidence, but it needs a live FalkorDB and I can't stand one up here, so I've left it out rather than commit something I haven't actually run. Happy to add it if a maintainer wants it in this PR.

---

linhongyu510 (approved):
Re-reviewed exact head `03d2862da67f`. The follow-up now threads the per-call routed `driver`/`clients` through extraction, dedupe, resolution, Saga lookup/write, and final persistence paths while preserving the shared-client fallback for existing callers. The new tests observe the actual persistence/routing boundaries rather than only shared-object identity. Local verification: `tests/test_group_routing_does_not_mutate_client.py` 14 passed; targeted Ruff lint and format checks passed; `git diff --check origin/main...HEAD` passed. I did not run a live two-group FalkorDB integration, so this approval is based on deterministic boundary tests and code-path inspection.

Anai-Guo:
Closing this — it has been superseded upstream.

`main` now carries an equivalent (and broader) fix for #1676 via #1699
(`08834609`, merged 2026-09-08): `_resolve_request_scope()` returns a per-call
`(group_id, driver, clients)` triple instead of reassigning `self.driver` /
`self.clients.driver`, and both `add_episode` and `add_episode_bulk` thread that
driver and clients bundle through `_get_or_create_saga`,
`_saga_get_previous_episode_uuid`, `_process_episode_data`,
`_extract_and_resolve_edges`, `_extract_and_dedupe_nodes_bulk` and
`_resolve_nodes_and_edges_bulk` — the same call sites this PR touched, with the
same default-to-`self.*` behaviour for the single-database case.

Thanks @linhongyu510 for the review and for walking me through the earlier
round of feedback. Nothing here is worth rebasing on top of #1699, so I would
rather close it than leave a conflicting duplicate open.

PR #1799, opened 2026-08-25
fix(search): scope Neptune fulltext search to group_ids

yatharth-anand-2410:
## Summary
`node_fulltext_search`, `episode_fulltext_search` and `community_fulltext_search` each compute a `group_id` filter for their non-Neptune (Cypher fulltext procedure) branch, but the Neptune branch -- which resolves candidate uuids via AOSS and then re-MATCHes them by uuid -- never spliced that filter into its own Cypher. A caller passing `group_ids` got results across every group/tenant on Neptune, even though the same call correctly scoped to `group_ids` on every other backend.

`community_fulltext_search`'s Neptune `MATCH` binds the `comm` alias, not `c` (the alias the non-Neptune branch's `group_filter_query` is written for), so it needed its own alias-correct filter rather than reusing the existing variable.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
N/A (bug fix, not a new feature/perf change).

## Testing
- [x] All existing tests pass

`tests/utils/search/test_neptune_fulltext_group_scoping.py` uses a recording driver (no live database) to assert the `group_id` predicate is present in the emitted Cypher, with the correct alias, when `group_ids` is passed, and absent when it is not. Verified failing against the pre-fix code and passing after.

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
N/A -- no existing issue tracks this; found while auditing Neptune's fulltext search branches for group_id scoping.

### DIFF
--- a/graphiti_core/driver/neptune_driver.py
+++ b/graphiti_core/driver/neptune_driver.py
@@ -68,7 +68,7 @@ aoss_indices = [
                     'uuid': {'type': 'keyword'},
                     'name': {'type': 'text'},
                     'summary': {'type': 'text'},
-                    'group_id': {'type': 'text'},
+                    'group_id': {'type': 'text', 'fields': {'keyword': {'type': 'keyword'}}},
                 }
             }
         },
@@ -84,7 +84,7 @@ aoss_indices = [
                 'properties': {
                     'uuid': {'type': 'keyword'},
                     'name': {'type': 'text'},
-                    'group_id': {'type': 'text'},
+                    'group_id': {'type': 'text', 'fields': {'keyword': {'type': 'keyword'}}},
                 }
             }
         },
@@ -102,7 +102,7 @@ aoss_indices = [
                     'content': {'type': 'text'},
                     'source': {'type': 'text'},
                     'source_description': {'type': 'text'},
-                    'group_id': {'type': 'text'},
+                    'group_id': {'type': 'text', 'fields': {'keyword': {'type': 'keyword'}}},
                 }
             }
         },
@@ -124,7 +124,7 @@ aoss_indices = [
                     'uuid': {'type': 'keyword'},
                     'name': {'type': 'text'},
                     'fact': {'type': 'text'},
-                    'group_id': {'type': 'text'},
+                    'group_id': {'type': 'text', 'fields': {'keyword': {'type': 'keyword'}}},
                 }
             }
         },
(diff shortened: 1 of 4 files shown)


linhongyu510 (changes requested):
The Cypher post-filter prevents cross-group rows from being returned, but it does not make the full-text search correctly scoped. `run_aoss_query(..., limit=limit)` still selects the top `limit` hits globally before the new `group_id` predicate runs. If those slots belong to other groups, the Cypher stage drops them and returns fewer results (or none), even when matching documents in the requested group exist immediately below the AOSS cutoff. Please pass `group_ids` into the AOSS query and apply a bool/filter there, then retain the Cypher predicate as defense in depth. Add a regression where the global top-N belong to another group and a requested-group hit ranks below them; the current recording tests only inspect the second-stage query and cannot catch this recall failure.

There is also an existing typo in the episode Neptune branch that prevents this PR's episode-scoping claim from being exercised: `input_ids` contains `{'id': ...}`, but the Cypher matches `WHERE e.uuid=i.uuid`. The new test returns no rows unconditionally, so it passes despite that mismatch. Please use `i.id` and make the test validate a returned episode (or otherwise fail when the candidate key and Cypher key diverge).

Validation on `9ffbcdca`: the four added tests pass, Ruff check/format and `git diff --check` pass; the blocking issues are semantic coverage rather than lint/test failures.

yathanand:
Good catches, both confirmed and fixed in 04cd5c5:

1. **Recall bug**: `run_aoss_query` now takes `group_ids` and applies an exact-match filter (`terms` on `group_id.keyword`) at the OpenSearch level, before the top-`limit` cutoff, instead of relying solely on the downstream Cypher check. This required adding a `.keyword` sub-field to `group_id` in the AOSS index mappings, since the existing field is analyzed text and can't support exact filtering — note this means **existing deployed indices need to be recreated** (`delete_all_indexes` + `build_indices_and_constraints`) to pick up the new sub-field, since OpenSearch/AOSS can't add a sub-field to an already-indexed field via a mapping update. The Cypher-side `group_id IN $group_ids` check stays in place as defense-in-depth.
2. **Episode typo**: `input_ids` uses key `'id'` but the Cypher matched on `i.uuid`, so that branch always returned zero rows regardless of scoping — fixed the Cypher to use `i.id` (matching the node/community convention), and rewrote the regression test to assert an actual row comes back instead of just checking the query string.

Added a new test file for `NeptuneDriver.run_aoss_query`'s filter construction, and updated the existing scoping tests to assert on real results.

PR #1800, opened 2026-08-25
fix(edges): always persist source_node_uuid/target_node_uuid, never source_uuid/target_uuid

yatharth-anand-2410:
## Summary
  `EntityEdge.save`'s single-edge write path (`edges.py`, and each provider's `EntityEdgeOperations.save` -- the actual live path when `driver.graph_operations_interface` is set) built its MATCH-key payload as `source_uuid`/`target_uuid`. The bulk-edge write path instead used `edge.model_dump()`, whose real Pydantic field names are `source_node_uuid`/`target_node_uuid`. Both write paths do `SET e = <payload>` (or the FalkorDB/Neptune equivalent), so the raw relationship property name depended on which path created the edge: single-save and bulk-save produced two incompatible property shapes for the same relationship type in the same graph, and the single-save shape leaked into `EntityEdge.attributes` on read since the record parsers only ever stripped `source_node_uuid`/`target_node_uuid`.

  This standardizes every write path -- the shared query builder (`edge_db_queries.py`), the generic `EntityEdge.save` fallback (`edges.py`), and each provider's `EntityEdgeOperations.save` (Neo4j, FalkorDB, Neptune, Kuzu) -- on `source_node_uuid`/`target_node_uuid`, matching what bulk save, the return-query aliases, and the record parsers already expect. No read-side changes were needed: the existing `attributes.pop('source_node_uuid', ...)` / `pop('target_node_uuid', ...)` in both record parsers now correctly strips these keys regardless of which path wrote the edge.

  ## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Performance improvement
  - [ ] Documentation/Tests

  ## Objective
  N/A (bug fix, not a new feature/perf change).

  ## Testing
  - [x] All existing tests pass

  `tests/test_edge_save_property_naming.py` (no live database) asserts, for every provider: (a) the shared query builder never emits a bare `source_uuid`/`target_uuid` token, (b) `EntityEdge.save`'s fallback path, and (c) each provider's `EntityEdgeOperations.save`, all send `source_node_uuid`/`target_node_uuid` and never `source_uuid`/`target_uuid`. Verified all 12 assertions failing against the pre-fix code and passing after.

  ## Breaking Changes

  Edges already saved by the pre-fix single-save path carry a `source_uuid`/`target_uuid` property in addition to `source_node_uuid`/`target_node_uuid` once re-saved; this PR does not add a migration for already-written data, since the record parsers already ignore unrecognized properties. Happy to add a migration note or backfill script if maintainers want one.

  ## Checklist
  - [x] Self-review completed

  ## Related Issues
  N/A -- no existing issue tracks this; found while auditing edge-save consistency across single-save and bulk-save paths. Related to #1690, which fixes the read-side `.attributes` leak for FalkorDB specifically; this PR fixes the write-side root cause across all providers instead.
  EOF
  cat /private/tmp/claude-501/-Users-stokome-Desktop-Personal-graphiti/4240c0fa-c065-4695-827b-2469b4a02d62/scratchpad/pr2_body.md)
## Summary
`EntityEdge.save`'s single-edge write path (`edges.py`, and each provider's `EntityEdgeOperations.save` -- the actual live path when `driver.graph_operations_interface` is set) built its MATCH-key payload as `source_uuid`/`target_uuid`. The bulk-edge write path instead used `edge.model_dump()`, whose real Pydantic field names are `source_node_uuid`/`target_node_uuid`. Both write paths do `SET e = <payload>` (or the FalkorDB/Neptune equivalent), so the raw relationship property name depended on which path created the edge: single-save and bulk-save produced two incompatible property shapes for the same relationship type in the same graph, and the single-save shape leaked into `EntityEdge.attributes` on read since the record parsers only ever stripped `source_node_uuid`/`target_node_uuid`.

This standardizes every write path -- the shared query builder (`edge_db_queries.py`), the generic `EntityEdge.save` fallback (`edges.py`), and each provider's `EntityEdgeOperations.save` (Neo4j, FalkorDB, Neptune, Kuzu) -- on `source_node_uuid`/`target_node_uuid`, matching what bulk save, the return-query aliases, and the record parsers already expect. No read-side changes were needed: the existing `attributes.pop('source_node_uuid', ...)` / `pop('target_node_uuid', ...)` in both record parsers now correctly strips these keys regardless of which path wrote the edge.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective
N/A (bug fix, not a new feature/perf change).

## Testing
- [x] All existing tests pass

`tests/test_edge_save_property_naming.py` (no live database) asserts, for every provider: (a) the shared query builder never emits a bare `source_uuid`/`target_uuid` token, (b) `EntityEdge.save`'s fallback path, and (c) each provider's `EntityEdgeOperations.save`, all send `source_node_uuid`/`target_node_uuid` and never `source_uuid`/`target_uuid`. Verified all 12 assertions failing against the pre-fix code and passing after.

## Breaking Changes

Edges already saved by the pre-fix single-save path carry a `source_uuid`/`target_uuid` property in addition to `source_node_uuid`/`target_node_uuid` once re-saved; this PR does not add a migration for already-written data, since the record parsers already ignore unrecognized properties. Happy to add a migration note or backfill script if maintainers want one.

## Checklist
- [x] Self-review completed

## Related Issues
N/A -- no existing issue tracks this; found while auditing edge-save consistency across single-save and bulk-save paths. Related to #1690, which fixes the read-side `.attributes` leak for FalkorDB specifically; this PR fixes the write-side root cause across all providers instead.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/entity_edge_ops.py
+++ b/graphiti_core/driver/falkordb/operations/entity_edge_ops.py
@@ -41,8 +41,8 @@ class FalkorEntityEdgeOperations(EntityEdgeOperations):
     ) -> None:
         edge_data: dict[str, Any] = {
             'uuid': edge.uuid,
-            'source_uuid': edge.source_node_uuid,
-            'target_uuid': edge.target_node_uuid,
+            'source_node_uuid': edge.source_node_uuid,
+            'target_node_uuid': edge.target_node_uuid,
             'name': edge.name,
             'fact': edge.fact,
             'fact_embedding': edge.fact_embedding,
--- a/graphiti_core/driver/falkordb/operations/entity_edge_ops.py
+++ b/graphiti_core/driver/falkordb/operations/entity_edge_ops.py
@@ -39,21 +39,23 @@ class FalkorEntityEdgeOperations(EntityEdgeOperations):
         edge: EntityEdge,
         tx: Transaction | None = None,
     ) -> None:
-        edge_data: dict[str, Any] = {
-            'uuid': edge.uuid,
-            'source_node_uuid': edge.source_node_uuid,
-            'target_node_uuid': edge.target_node_uuid,
-            'name': edge.name,
-            'fact': edge.fact,
-            'fact_embedding': edge.fact_embedding,
-            'group_id': edge.group_id,
-            'episodes': edge.episodes,
-            'created_at': edge.created_at,
-            'expired_at': edge.expired_at,
-            'valid_at': edge.valid_at,
-            'invalid_at': edge.invalid_at,
-        }
-        edge_data.update(edge.attributes or {})
+        edge_data: dict[str, Any] = dict(edge.attributes or {})
+        edge_data.update(
+            {
+                'uuid': edge.uuid,
+                'source_node_uuid': edge.source_node_uuid,
+                'target_node_uuid': edge.target_node_uuid,
+                'name': edge.name,
+                'fact': edge.fact,
+                'fact_embedding': edge.fact_embedding,
+                'group_id': edge.group_id,
+                'episodes': edge.episodes,
+                'created_at': edge.created_at,
+                'expired_at': edge.expired_at,
+                'valid_at': edge.valid_at,
+                'invalid_at': edge.invalid_at,
+            }
+        )
 
         query = get_entity_edge_save_query(GraphProvider.FALKORDB)
         if tx is not None:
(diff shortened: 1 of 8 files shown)


linhongyu510 (changes requested):
The canonical write keys are correct for fresh edges, but the current merge and parser behavior does not guarantee the stated invariant and introduces an endpoint-integrity regression. Both record parsers remove only `source_node_uuid` and `target_node_uuid`; an edge written by an older release therefore loads with legacy `source_uuid` and `target_uuid` inside `EntityEdge.attributes`. Every non-Kuzu provider then builds canonical fields and calls `edge_data.update(edge.attributes)`, so re-saving that edge writes both old and new key pairs. My reproduced payload contained all four endpoint keys. More seriously, because the query now MATCHes on canonical payload keys, attributes named `source_node_uuid` or `target_node_uuid` overwrite the trusted model fields after they are set; a model edge with endpoints `src` and `dst` produced a payload matching `wrong-src` and `wrong-dst`. The 12 new tests all use empty attributes and miss both cases. Please strip legacy aliases and reserved canonical endpoint keys when parsing or merging attributes, or merge trusted model fields after attributes, and add single, bulk, and provider regressions for legacy-read-to-save and reserved-key collision. The Kuzu JSON attributes path should be checked for the same leakage.

gkorland:
@galshubeli — this PR touches FalkorDB edge operations (source_node_uuid/target_node_uuid persistence in `graphiti_core/driver/falkordb/operations/entity_edge_ops.py`) and hasn't had FalkorDB team review yet. Could you take a look?

yathanand:
Thanks for catching this — both were real bugs, confirmed and fixed in f53079f:

1. **Legacy alias leak**: `entity_edge_from_record` now also pops `source_uuid`/`target_uuid` (not just the canonical `source_node_uuid`/`target_node_uuid`), so a pre-fix edge loaded from the DB no longer carries the legacy keys into `attributes` and re-persists them on next save.
2. **Reserved-key collision**: in each of neo4j/falkordb/neptune's `save()`/`save_bulk()`, `edge_data` is now built by seeding from `edge.attributes` first and then overlaying the trusted `uuid`/`source_node_uuid`/`target_node_uuid`/etc. fields, so an attribute that happens to be named `source_node_uuid`/`target_node_uuid` can never overwrite the real endpoints before the query runs.

Added regression tests for both: a legacy-record-then-resave case and a reserved-key-collision case for `save` and `save_bulk` on all three providers. Kuzu didn't need changes — its attributes are stored as a separate JSON column, never merged into the query params.

gkorland:

PR #1810, opened 2026-08-30
feat(mcp): surface ranker, scores, and invalidated-fact counts in search results (#1645)

tonydzi:
This is the "observable half" of #1645, scoped exactly to what @brentkearney invited there:

> `suppressed_invalidated` is the piece I'd most want to exist — it delivers the "this expired" signal without returning history the caller didn't ask for, and it doesn't require the filter to default on at all. Agreed on emitting `ranker` beside `score`: `reranker` is deployment config, so an operator flipping `rrf` → `mmr` silently invalidates every client-side `score > 0.4` with no error anywhere.

## What this adds

MCP server only; no changes to `graphiti_core`.

**1. `ranker` in both search responses.** `search_nodes` and `search_memory_facts` now report which reranker ordered the results (`rrf` or `node_distance`, derived from the recipe actually passed to `search_`, not hardcoded). A client-side score cutoff calibrated on one ranker is meaningless on another, so the response names the ranker that produced the scores.

**2. `score` on each result.** Nodes and facts carry the reranker score from `SearchResults.node_reranker_scores` / `edge_reranker_scores`. To get those for facts, `search_memory_facts` now calls `client.search_()` with an explicit copy of the same recipe `Graphiti.search()` selects (`EDGE_HYBRID_SEARCH_RRF`, or `EDGE_HYBRID_SEARCH_NODE_DISTANCE` with a center node) — `client.search()` discards the scores. The recipe is `model_copy`'d before setting `limit`, so the shared module-level recipe is no longer mutated per request.

**3. `invalidated_count` + `invalidated_uuids` on fact responses.** How many of the returned facts have `invalid_at`/`expired_at` set, and which ones. This lets an agent distinguish "no fact was ever recorded" from "a fact existed and every version of it has been superseded" — the silent-zero failure discussed in the issue.

On naming: the issue called this `suppressed_invalidated`, but since no default filter ships (see below), nothing is actually suppressed — the invalidated facts are still in the results, per the bi-temporal design. `invalidated_count` describes what the field really counts. Happy to rename if you prefer the original.

## Deliberately not included

- **No default-on filtering of invalidated facts.** You called `exclude_invalidated: true` the wrong default — graphiti is bi-temporal by design and `invalid_at` is already on every result. Result sets are byte-for-byte the same facts as before; only metadata is added.
- **No reranker configurability.** Net-zero per your comment; #1637 covers the cross-encoder provider side.
- **No "superseding fact" resolution.** As you noted, `edge_operations` sets `invalid_at`/`expired_at` with no back-link to the successor edge, so returning the replacement isn't automatic — this PR only signals that supersession happened.
- **No score threshold / `reranker_min_score`.** MMR scores span negatives (#777), and our own measurements in the issue thread showed a cutoff converts loud wrong answers into silent ones. Observability first; cutoffs can be calibrated from real traffic later, by whoever wants them.

## Tests

New unit tests in `mcp_server/tests/test_search_observability.py` (mocked client, no database), covering: ranker reported for both tools on both recipes, scores attached per result, invalidated/superseded counting, empty-result responses still carrying `ranker`, and the shared-recipe-not-mutated guard.

```
$ uv run pytest tests/test_search_observability.py
tests/test_search_observability.py::test_facts_response_reports_ranker_and_scores PASSED  [ 12%]
tests/test_search_observability.py::test_facts_response_counts_invalidated_and_superseded PASSED [ 25%]
tests/test_search_observability.py::test_facts_empty_result_still_reports_ranker PASSED   [ 37%]
tests/test_search_observability.py::test_facts_center_node_selects_node_distance_ranker PASSED [ 50%]
tests/test_search_observability.py::test_facts_limit_set_on_copy_not_shared_recipe PASSED [ 62%]
tests/test_search_observability.py::test_nodes_response_reports_ranker_and_scores PASSED  [ 75%]
tests/test_search_observability.py::test_nodes_empty_result_still_reports_ranker PASSED   [ 87%]
tests/test_search_observability.py::test_nodes_center_node_selects_node_distance_ranker PASSED [100%]
============================== 8 passed in 5.10s ===============================
```

Also ran the neighboring unit suites (`test_configuration.py`, `test_core_parity.py`): 40 passed. `ruff format`/`ruff check` clean; `pyright` on the touched files: 0 errors.

---

Authored by Mycroft, the synthetic co-founder at Anton Dzyatkovsky's lab (autonomous mode; named responsible person: Anton Dziatkovskii). The test runs above were independently re-executed before submission.

### DIFF
--- a/mcp_server/src/graphiti_mcp_server.py
+++ b/mcp_server/src/graphiti_mcp_server.py
@@ -505,6 +505,9 @@ async def search_nodes(
         entity_types: Optional list of entity type names (node labels) to filter by
         center_node_uuid: Optional UUID of a node to center the search around. Results
             closer to this node in the graph are ranked higher.
+
+    Each node carries a 'score', and the response reports the 'ranker' that
+    produced it; scores are only comparable within the same ranker.
     """
     global graphiti_service
 
@@ -540,6 +543,10 @@ async def search_nodes(
         node_config = (
             NODE_HYBRID_SEARCH_NODE_DISTANCE if center_node_uuid else NODE_HYBRID_SEARCH_RRF
         )
+        # Report which reranker ordered the results: the reranker is deployment
+        # config, so a bare score is meaningless without it (score semantics
+        # differ between rankers).
+        ranker = node_config.node_config.reranker.name if node_config.node_config else 'unknown'
         results = await client.search_(
             query=query,
             config=node_config,
@@ -552,12 +559,16 @@ async def search_nodes(
         nodes = results.nodes[:max_nodes] if results.nodes else []
 
         if not nodes:
-            return NodeSearchResponse(message='No relevant nodes found', nodes=[])
+            return NodeSearchResponse(message='No relevant nodes found', nodes=[], ranker=ranker)
 
         # Format the results (embeddings stripped by to_node_result)
         node_results = [to_node_result(node) for node in nodes]
+        for node_result, score in zip(node_results, results.node_reranker_scores, strict=False):
+            node_result['score'] = score
 
-        return NodeSearchResponse(message='Nodes retrieved successfully', nodes=node_results)
+        return NodeSearchResponse(
+            message='Nodes retrieved successfully', nodes=node_results, ranker=ranker
+        )
     except Exception as e:
         error_msg = str(e)
         logger.error(f'Error searching nodes: {error_msg}')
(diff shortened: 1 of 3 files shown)


tonydzi:

tonydzi:
recheck

PR #1826, opened 2026-09-02
fix(falkordb): use yielded node directly in episode fulltext search

### LINKED ISSUE #1819: [BUG] FalkorDB: episode_fulltext_search re-matches yielded nodes by unindexed uuid equality — O(hits × episodes), hangs search
## Bug Description

`episode_fulltext_search` in `graphiti_core/driver/falkordb/operations/search_ops.py` (~line 465 at v0.29.3) re-matches every fulltext hit against all `Episodic` nodes by unindexed equality instead of using the node the procedure already yielded:

```cypher
CALL db.idx.fulltext.queryNodes('episode_content', $query) YIELD node AS episode, score
MATCH (e:Episodic)
WHERE e.uuid = episode.uuid
```

The `WHERE e.uuid = episode.uuid` predicate can't be answered from an index (the right-hand side is a per-row value), so each yielded hit triggers a scan over all `Episodic` nodes — O(hits × episodes) — and episode nodes carry large `content` properties, making each row expensive. This is the same class of problem as #1500 (which covers the `RELATES_TO` re-MATCH in edge search) but on the episode path, and #1500 as currently written does not fix it.

On our production FalkorDB graph (2,163 `Episodic` nodes, ~16k `RELATES_TO` edges), any search recipe that includes the episode scope (e.g. `COMBINED_HYBRID_SEARCH_RRF` / `COMBINED_HYBRID_SEARCH_CROSS_ENCODER`) fails to complete within a 30s timeout, with FalkorDB pinning a core. Worse, because FalkorDB keeps executing the abandoned query after the client times out, subsequent unrelated queries queue behind it — one hung search degrades the whole instance.

## Fix

The yielded node *is* the target — use it directly, exactly as the `node_fulltext_search` and `community_fulltext_search` variants in the same file already do:

```cypher
CALL db.idx.fulltext.queryNodes('episode_content', $query) YIELD node AS e, score
WITH e, score
RETURN ...
ORDER BY score DESC
LIMIT $limit
```

We're running this patch in production (v0.29.3 + this change + #1500's changes): combined-recipe searches went from never-completing to low single-digit seconds; the `search_memory_facts` MCP path is back under ~1s warm.

## Environment

- graphiti-core: 0.29.3 (PyPI)
- FalkorDB: falkordb/falkordb:latest (graph module ver 41808), falkordb-py 1.2.0
- Python 3.13

Related: #1500 (same O(matches × graph) class on the edge search paths).

maui314159:
Fixes #1819. Same O(matches × graph) class as #1500, on the episode path.

## Problem

`episode_fulltext_search` in `graphiti_core/driver/falkordb/operations/search_ops.py` re-matches every fulltext hit against all Episodic nodes:

```cypher
CALL db.idx.fulltext.queryNodes('episode_content', $query) YIELD node AS episode, score
MATCH (e:Episodic)
WHERE e.uuid = episode.uuid
```

The `WHERE e.uuid = episode.uuid` predicate can't be answered from an index (the right-hand side is a per-row value), so each yielded hit scans all Episodic nodes — O(hits × episodes) over content-heavy nodes. On our production FalkorDB graph (2,163 episodes, ~16k edges) any search recipe including the episode scope failed to complete within 30s, pinning a core; because FalkorDB keeps executing after the client gives up, unrelated queries then queue behind the abandoned scan, so one search degrades the whole instance.

## Fix

Use the node the procedure already yields — exactly as the `node_fulltext_search` and `community_fulltext_search` variants in the same file do:

```cypher
CALL db.idx.fulltext.queryNodes('episode_content', $query) YIELD node AS e, score
WITH e, score
RETURN ... ORDER BY score DESC LIMIT $limit
```

Behavior is otherwise unchanged: optional `group_ids` filtering still applies to the yielded node, `EPISODIC_NODE_RETURN` / score ordering / `$limit` are preserved, and an empty/all-stopword query still returns before executing Cypher.

We're running this patch in production (alongside #1500's edge fixes): combined-recipe searches went from never completing to low single-digit seconds.

## Tests

`tests/driver/test_falkordb_episode_fulltext.py` — query-shape regression tests with a mocked executor (no live FalkorDB needed, mirrors `test_falkordb_ops_routing.py` conventions), scoped to the FalkorDB operations class so other providers aren't coupled to this query-plan workaround. Asserts the re-match shape is gone, the yielded node is used, group filtering/ordering/limit are preserved, and the empty-query short-circuit still holds. `ruff format` / `ruff check` clean; full `tests/driver/` suite passes.

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

https://claude.ai/code/session_01Rw6VtYNo2cqrgzVdAAs3DU

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -395,17 +395,21 @@ class FalkorSearchOperations(SearchOperations):
         filter_params: dict[str, Any] = {}
         group_filter_query = ''
         if group_ids is not None:
-            group_filter_query += '\nAND e.group_id IN $group_ids'
+            group_filter_query += '\nWHERE e.group_id IN $group_ids'
             filter_params['group_ids'] = group_ids
 
+        # Use the node yielded by the fulltext procedure directly (as the node
+        # and community variants do). Re-matching it via
+        # `MATCH (e:Episodic) WHERE e.uuid = episode.uuid` cannot be answered
+        # from an index (the right-hand side is a per-row value), so every hit
+        # triggered a scan over all Episodic nodes — O(hits x episodes).
         cypher = (
             get_nodes_query(
                 'episode_content', '$query', limit=limit, provider=GraphProvider.FALKORDB
             )
             + """
-            YIELD node AS episode, score
-            MATCH (e:Episodic)
-            WHERE e.uuid = episode.uuid
+            YIELD node AS e, score
+            WITH e, score
             """
             + group_filter_query
             + """
--- a/tests/driver/test_falkordb_episode_fulltext.py
+++ b/tests/driver/test_falkordb_episode_fulltext.py
@@ -1,0 +1,106 @@
+"""Unit tests for the FalkorDB episode fulltext search query shape.
+
+Regression guard for https://github.com/getzep/graphiti/issues/1819: the
+fulltext procedure already yields the Episodic node, but the query used to
+re-match it via ``MATCH (e:Episodic) WHERE e.uuid = episode.uuid``. That
+predicate cannot be answered from an index (the right-hand side is a per-row
+value), so every fulltext hit triggered a scan over all Episodic nodes —
+O(hits x episodes) — which pins a FalkorDB core and wedges any search recipe
+that includes the episode scope once the graph is non-trivial.
+
+The tests assert on the generated Cypher with a mocked executor, so they need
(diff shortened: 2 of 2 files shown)


maui314159:
I have read the CLA Document and I hereby sign the CLA behalf of my company, e-mail: maui@growthcurvecapital.com

gkorland:

gkorland:

PR #1828, opened 2026-09-02
fix(mcp): make LLM and embedder client construction fatal

### LINKED ISSUE #1827: [BUG] GraphitiService silently falls back to unconfigured OpenAIClient() when LLM client construction fails
## Bug Description

`GraphitiService.initialize()` in `mcp_server/src/graphiti_mcp_server.py` wraps `LLMClientFactory.create()` in a bare `except Exception as e: logger.warning(f'Failed to create LLM client: {e}')`, leaving `llm_client = None` on any failure. `graphiti_core.Graphiti.__init__()` then runs `self.llm_client = llm_client if llm_client else OpenAIClient()`, silently substituting a completely unconfigured client. That client ignores the user's provider, model, and base_url settings entirely, reads `OPENAI_API_KEY`/`OPENAI_BASE_URL` straight from the process environment, and issues structured extraction calls via the Responses API using graphiti-core's hardcoded `DEFAULT_MODEL`/`DEFAULT_SMALL_MODEL`. The only trace of the original failure is a warning-level log line; nothing else in the server's behavior indicates that the configured LLM client never got built.

We hit this via a concrete trigger: `LLMClientFactory.create()`'s `openai` branch selects `OpenAIGenericClient` for non-OpenAI base URLs (via `is_non_openai_provider()`) and passes `structured_output_mode=config.structured_output_mode` (added in #1619). In our deployment, the installed `graphiti-core` build predated `OpenAIGenericClient.__init__()` accepting that keyword argument, so the call raised `TypeError: OpenAIGenericClient.__init__() got an unexpected keyword argument 'structured_output_mode'`. That exception was swallowed by the `except Exception` above and never surfaced beyond the warning log. For the process's entire uptime (multiple days), every structured extraction call went out through the fallback client's hardcoded default model over the Responses API, against a self-hosted OpenAI-compatible proxy (LiteLLM) whose credential for this workload was not scoped to that model, so every call failed. We only noticed once an unrelated proxy-side failure-rate alert fired, and only found the actual cause by shelling into the running container and reproducing `LLMClientFactory.create()` directly against the live config. We're separately tracking down why our own image ended up with an older graphiti-core than intended, since that part may well be a mistake in our own build pipeline rather than anything here, but the behavior we're reporting doesn't depend on that: whatever the reason `LLMClientFactory.create()` might raise (this keyword mismatch, a malformed API key, an unreachable base_url, anything), the current code treats it as a warning and quietly swaps in a different, unconfigured client rather than failing.

That's inconsistent with the cross-encoder construction two lines below the LLM client code in the same function, which already states the intended principle in a comment:

```python
# Create cross-encoder (reranker) client. Without this, Graphiti defaults to
# OpenAIRerankerClient, which needs an OpenAI API key even on non-OpenAI setups.
# Reranker setup errors must remain fatal rather than silently restoring that default.
cross_encoder_client = CrossEncoderFactory.create(self.config.llm, self.config.embedder)
```

The LLM client, and the embedder client a few lines above it (same `except Exception: logger.warning(...)` pattern), don't follow that principle.

## Steps to Reproduce

Minimal reproduction of the core-level design issue, no MCP server involved:

```python
import os

os.environ['OPENAI_API_KEY'] = '<a key scoped to a proxy that only allows model X>'
os.environ['OPENAI_BASE_URL'] = 'https://internal-proxy.example.com/v1'

# Caller intended to always pass an explicit llm_client; a bug upstream of
# this call (see the MCP server path below) leaves it as None instead.
g = Graphiti(uri='bolt://localhost:7687', user='neo4j', password='...', llm_client=None)

# No error raised. g.llm_client is a bare OpenAIClient() using graphiti-core's
# hardcoded DEFAULT_MODEL, completely ignoring whatever provider the caller's
# other configuration implied. Every subsequent add_episode() call sends
# requests to DEFAULT_MODEL / DEFAULT_SMALL_MODEL through OPENAI_BASE_URL.
```

And the concrete path that actually produces `llm_client=None` in the MCP server, `mcp_server/src/graphiti_mcp_server.py`, `GraphitiService.initialize()`:

```python
try:
    llm_client = LLMClientFactory.create(self.config.llm)
except Exception as e:
    logger.warning(f'Failed to create LLM client: {e}')
```

Any exception here, including one caused by a graphiti-core/mcp_server version mismatch on a factory kwarg, is caught, logged as a warning, and the server continues to initialize and accept requests.

## Expected Behavior

LLM client (and embedder client) construction failure should be fatal, or at minimum `Graphiti.__init__()` should not silently substitute an unconfigured default when the caller's own factory explicitly failed to build one. The failure should stop startup or otherwise be impossible to miss, not a warning-level log line while the server keeps serving requests against a completely different provider than configured.

## Actual Behavior

A warning is logged once at startup; the server proceeds normally and answers every subsequent extraction call using a bare `OpenAIClient()` against whatever `OPENAI_API_KEY`/`OPENAI_BASE_URL` happen to be set in the process environment, using graphiti-core's hardcoded default models, not the configured provider/model.

## Environment

- **Graphiti Version**: graphiti-core 0.29.1, mcp_server built from `mcp_server/docker/Dockerfile.standalone` at commit `1ba04a7`
- **Python Version**: 3.11 (the `Dockerfile.standalone` base image)
- **Operating System**: Linux (container), Kubernetes
- **Database Backend**: Neo4j
- **LLM Provider & Model**: OpenAI-compatible (self-hosted LiteLLM proxy), `llm.provider: openai` with a custom `api_url`

## Installation Method

- [x] Development installation (Docker image built from `mcp_server/docker/Dockerfile.standalone`)

## Error Messages/Traceback

```
WARNING - Failed to create LLM client: OpenAIGenericClient.__init__() got an unexpected keyword argument 'structured_output_mode'
```

This is the only trace of the failure. Nothing else in the logs indicates the configured LLM client never got built.

## Configuration

```yaml
llm:
  provider: "openai"
  model: "<configured model, never actually used>"
  structured_output_mode: "json_schema"
  providers:
    openai:
      api_url: "https://<self-hosted-proxy>/v1"
```

## Additional Context

- Reproduces consistently for the life of the process once the fallback client is installed, not intermittent.
- Component: MCP server (`GraphitiService.initialize()`) plus the underlying design in `graphiti_core.Graphiti.__init__()`.
- Related: #1746 documents different but adjacent OpenAI-compatible-endpoint failures (Responses-API-only extraction, `DEFAULT_SMALL_MODEL='gpt-4.1-nano'` rejected by non-OpenAI backends). This issue is specifically about the silent-fallback-on-construction-failure behavior, a distinct defect from those, though it compounds with them: the fallback client is exactly the code path #1746's `DEFAULT_SMALL_MODEL` problem lives in.

## Possible Solution

1. In `GraphitiService.initialize()`, treat LLM client (and embedder client) construction failure as fatal, consistent with the principle the adjacent cross-encoder comment already states.
2. In `graphiti_core.Graphiti.__init__()`, consider making the "use a bare default client when none is given" behavior something a caller opts into explicitly, rather than the implicit behavior of any falsy `llm_client`/`embedder`, so a caller-side factory bug can't silently degrade into "quietly use a totally different provider."

alekc:
## Summary
LLM and embedder client construction failures were caught and only logged as warnings; Graphiti then silently used an unconfigured default OpenAIClient()/OpenAIEmbedder(), ignoring the caller's actual provider. Makes both fatal, matching the cross-encoder construction two lines below, which already states and follows this principle.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Testing
- [x] All existing tests pass

## Breaking Changes
- [x] This PR contains breaking changes

A misconfigured LLM or embedder provider now fails server startup instead of silently degrading to an unconfigured OpenAI default. Migration: fix the provider config so factory construction succeeds.

## Checklist
- [x] Self-review completed

## Related Issues
Closes #1827

### DIFF
--- a/mcp_server/src/graphiti_mcp_server.py
+++ b/mcp_server/src/graphiti_mcp_server.py
@@ -202,21 +202,13 @@ class GraphitiService:
     async def initialize(self) -> None:
         """Initialize the Graphiti client with factory-created components."""
         try:
-            # Create clients using factories
-            llm_client = None
-            embedder_client = None
-
-            # Create LLM client based on configured provider
-            try:
-                llm_client = LLMClientFactory.create(self.config.llm)
-            except Exception as e:
-                logger.warning(f'Failed to create LLM client: {e}')
-
-            # Create embedder client based on configured provider
-            try:
-                embedder_client = EmbedderFactory.create(self.config.embedder)
-            except Exception as e:
-                logger.warning(f'Failed to create embedder client: {e}')
+            # Create clients using factories. LLM and embedder construction errors
+            # must remain fatal, same as the cross-encoder below: Graphiti.__init__()
+            # silently substitutes an unconfigured default (OpenAIClient() /
+            # OpenAIEmbedder()) for any client left as None, which would bypass this
+            # server's configured provider entirely rather than fail loudly.
+            llm_client = LLMClientFactory.create(self.config.llm)
+            embedder_client = EmbedderFactory.create(self.config.embedder)
 
             # Create cross-encoder (reranker) client. Without this, Graphiti defaults to
             # OpenAIRerankerClient, which needs an OpenAI API key even on non-OpenAI setups.
@@ -321,18 +313,10 @@ class GraphitiService:
 
             logger.info('Successfully initialized Graphiti client')
 
-            # Log configuration details
-            if llm_client:
-                logger.info(
-                    f'Using LLM provider: {self.config.llm.provider} / {self.config.llm.model}'
-                )
-            else:
-                logger.info('No LLM client configured - entity extraction will be limited')
-
-            if embedder_client:
-                logger.info(f'Using Embedder provider: {self.config.embedder.provider}')
-            else:
-                logger.info('No Embedder client configured - search will be limited')
+            # Log configuration details. llm_client and embedder_client are always
+            # set by this point: construction failures above are fatal, not swallowed.
+            logger.info(f'Using LLM provider: {self.config.llm.provider} / {self.config.llm.model}')
+            logger.info(f'Using Embedder provider: {self.config.embedder.provider}')
 
             if self.entity_types:
                 entity_type_names = list(self.entity_types.keys())
--- a/mcp_server/tests/test_cross_encoder_factory.py
+++ b/mcp_server/tests/test_cross_encoder_factory.py
@@ -5,7 +5,6 @@ import builtins
 import logging
 import sys
 from pathlib import Path
-from unittest.mock import AsyncMock, Mock
 
 import pytest
 
@@ -15,14 +14,11 @@ sys.path.insert(0, str(Path(__file__).parent.parent / 'src'))
 from graphiti_core.cross_encoder.gemini_reranker_client import GeminiRerankerClient
 from graphiti_core.cross_encoder.openai_reranker_client import OpenAIRerankerClient
 
-import graphiti_mcp_server
 from config.schema import (
     AnthropicProviderConfig,
-    DatabaseConfig,
     EmbedderConfig,
     EmbedderProvidersConfig,
     GeminiProviderConfig,
-    GraphitiConfig,
     LLMConfig,
     LLMProvidersConfig,
     OpenAIProviderConfig,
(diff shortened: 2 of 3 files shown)


alekc:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: [alexander@chernov.it](mailto:alexander@chernov.it)

alekc:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: alexander@chernov.it

alekc:
recheck

PR #1831, opened 2026-09-04
fix(falkordb): drop standalone fulltext punctuation

### LINKED ISSUE #1820: FalkorDB fulltext search: a standalone `_` token survives sanitization and makes RediSearch reject the whole query
### Summary

On the FalkorDB backend, `Graphiti.search()` raises `RediSearch: Syntax error` for any
query string in which an underscore ends up isolated between separator characters.
`sanitize_falkor_fulltext_query` maps FalkorDB's separators to whitespace but does not
treat `_` as one, so splitting a token like `<ISO-ts>_<uuid>` leaves a bare `_` in the
OR-chain that `build_falkor_fulltext_query` emits. RediSearch will not parse a lone `_`
as a term.

This is not a relevance problem — no results are returned and no partial match happens.
The whole search call fails.

Note the failure is specific to the *standalone* case. An underscore inside a word
(`a_b`) is fine; the bug only appears when the surrounding characters are themselves
separators, which is exactly what happens with identifier-shaped text such as file
paths, session ids, and any group id convention that uses `_` next to `-`, `<`, `>`,
`/` or `.`.

### Affected code

- `graphiti_core/driver/falkordb/fulltext.py` — `sanitize_falkor_fulltext_query()`,
  whose `_SEPARATOR_MAP` covers ``,.<>{}[]"':;!@#$%^&*()-+=~?|/\` `` but not `_`.
- `build_falkor_fulltext_query()` in the same module calls it and joins the surviving
  tokens with ` | `, so the bare `_` becomes a query term.
- Reached on every `Graphiti.search()` call against FalkorDB:
  `graphiti_core/search/search.py` →
  `search_utils.{node,edge,episode,community}_fulltext_search` →
  `search_utils.fulltext_query` → (`driver.provider == GraphProvider.FALKORDB`) →
  `FalkorDriver.build_fulltext_query` → `fulltext.build_falkor_fulltext_query`. The
  resulting string is passed straight to `db.idx.fulltext.queryNodes` as the `$query`
  parameter.
- Note that `FalkorDriver.build_fulltext_query` is now a one-line delegation to the
  module-level function and does **not** consult `self.sanitize` — `FalkorDriver.sanitize`
  is itself only a thin wrapper around `sanitize_falkor_fulltext_query` that nothing on
  the query path invokes. This matters for anyone working around the bug: overriding
  `sanitize` in a `FalkorDriver` subclass used to fix it and silently stopped doing
  anything in 0.29.3, while every name the override touches still exists.
- `graphiti_core/driver/falkordb/operations/search_ops.py` carries a
  `_build_falkor_fulltext_query` wrapper over the same function, but it is unreachable:
  `FalkorSearchOperations` is used only through `GraphDriver.search_interface`, which
  defaults to `None` and which `FalkorDriver` never assigns.

### Environment

| | |
| --- | --- |
| graphiti-core | 0.29.3 (extra: `falkordb`) |
| falkordb (client) | 1.7.1 |
| FalkorDB server | `falkordb/falkordb:v4.20.4` |
| Python | 3.14.7 |

### Reproduction

```bash
docker run -d --rm --name graphiti-repro -p 6379:6379 falkordb/falkordb:v4.20.4
```

```python
import asyncio

from graphiti_core.driver.falkordb.fulltext import (
    build_falkor_fulltext_query,
    sanitize_falkor_fulltext_query,
)
from graphiti_core.search.search_utils import fulltext_query, node_fulltext_search

QUERY = "sessions/<enc-cwd>/<ISO-ts>_<uuid>.jsonl"
GROUP_ID = "user_trent_workspace_ai_metrics"


async def main() -> None:
    driver = FalkorDriver(host="127.0.0.1", port=6379, database="repro")

    print("input       :", QUERY)
    print("sanitize    :", sanitize_falkor_fulltext_query(QUERY))
    print("built       :", build_falkor_fulltext_query(QUERY, [GROUP_ID]))
    print("fuzzy_query :", fulltext_query(QUERY, [GROUP_ID], driver))

    await driver.execute_query(
        "CALL db.idx.fulltext.createNodeIndex({label:'Entity'}, 'name', 'summary', 'group_id')"
    )
    await driver.execute_query(
        "CREATE (:Entity {uuid:'e1', name:'session jsonl', summary:'t', group_id:$g, "
        "created_at:'2026-01-01T00:00:00Z', labels:['Entity']})",
        g=GROUP_ID,
    )
    try:
        nodes = await node_fulltext_search(driver, QUERY, SearchFilters(), [GROUP_ID])
        print(f"RESULT: search returned {len(nodes)} node(s): {[n.uuid for n in nodes]}")
    except Exception as exc:
        print(f"RESULT: {type(exc).__module__}.{type(exc).__name__}: {exc}")
    await driver.close()


asyncio.run(main())
```

`node_fulltext_search` is the helper `graphiti_core/search/search.py` calls on every
`Graphiti.search()`, so this is the production path — but it does not depend on an LLM,
an embedder, or any ingested data beyond the one node created above.

### Actual output

```
input       : sessions/<enc-cwd>/<ISO-ts>_<uuid>.jsonl
sanitize    : sessions enc cwd ISO ts _ uuid jsonl
built       : (@group_id:"user\_trent\_workspace\_ai\_metrics") (sessions | enc | cwd | ISO | ts | _ | uuid | jsonl)
fuzzy_query : (@group_id:"user\_trent\_workspace\_ai\_metrics") (sessions | enc | cwd | ISO | ts | _ | uuid | jsonl)
RESULT: redis.exceptions.ResponseError: RediSearch: Syntax error at offset 87 near ts
```

(The driver also logs the failing Cypher and its `$query` parameter at ERROR before the
exception propagates. Elided above.)

### Expected

The bare `_` is dropped (or `_` is treated as a separator) and the search returns the
`e1` node. Confirmed by patching `fulltext.sanitize_falkor_fulltext_query` to pre-split
on `_` and rerunning the script otherwise unchanged:

```
fuzzy_query : (@group_id:"user\_trent\_workspace\_ai\_metrics") (sessions | enc | cwd | ISO | ts | uuid | jsonl)
RESULT: search returned 1 node(s): ['e1']
```

For contrast, an underscore *inside* a word is accepted by RediSearch as-is on the same
build — `session_jsonl name` and `a_b` both go through `node_fulltext_search` without
error. Only the standalone token is unparseable.

### Suggested fix

Either add `_` to `_SEPARATOR_MAP` in `graphiti_core/driver/falkordb/fulltext.py`, or
drop single-character non-alphanumeric tokens in the same comprehension of
`build_falkor_fulltext_query` that already removes stopwords. The second is narrower:
it leaves `a_b` intact, which RediSearch indexes and matches fine, and only removes the
tokens that cannot parse. Either lands in one place now that both the driver method and
the (unreachable) `search_ops` wrapper delegate to this module.

We currently carry the first shape as a pre-filter applied to the sanitizer. Worth
noting for anyone else working around this: the natural seam — overriding `sanitize` on
a `FalkorDriver` subclass — worked in 0.29.2 and no longer does, because
`build_fulltext_query` stopped consulting `self`. The override still installs and still
type-checks; it simply never runs, which turns a workaround into a silent regression on
upgrade. That is an argument for fixing it upstream rather than leaving downstream
users to patch a module-level function through another module's globals.

Happy to open a PR for whichever shape maintainers prefer.

mikemikimike:
## Summary

Fixes #1820.

FalkorDB fulltext queries can contain a standalone `_` token after sanitization. RediSearch rejects that token with a syntax error, so searches for identifier-shaped text such as `..._<uuid>...` fail instead of returning results.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Objective

Remove standalone non-alphanumeric tokens from the FalkorDB fulltext term list before joining terms with ` | `. This prevents invalid RediSearch queries while preserving embedded underscores such as `a_b`, which are valid search terms.

## Implementation

The existing stopword filter now also excludes one-character tokens that are not alphanumeric. The change is shared by the module-level builder and its existing driver/operations wrappers, so no public API or database schema changes are involved.

Added a parameterized regression covering the reported identifier-shaped query and the embedded-underscore compatibility case.

## Testing

- `.venv\\Scripts\\python.exe -m pytest -p no:cacheprovider tests/utils/search/test_search_security.py tests/utils/search/search_utils_test.py -q` — 22 passed
- `.venv\\Scripts\\ruff.exe check graphiti_core/driver/falkordb/fulltext.py tests/utils/search/test_search_security.py` — passed
- `.venv\\Scripts\\ruff.exe format --check graphiti_core/driver/falkordb/fulltext.py tests/utils/search/test_search_security.py` — passed
- `.venv\\Scripts\\pyright.exe graphiti_core/driver/falkordb/fulltext.py` — 0 errors
- `git diff --check` — passed

The full no-DB test command was attempted with FalkorDB, Kuzu, and Neptune disabled, but collection stopped at seven pre-existing missing optional provider dependencies (`sentence-transformers`, `google-genai`, `voyageai`, and `anthropic`). No database, LLM, API key, or network service is needed for the focused regression.

## Breaking Changes

## Checklist

- [x] Code follows project style guidelines for the touched files
- [x] Self-review completed
- [x] Documentation updated where necessary (no public API change)

## Related Issues

Closes #1820

## AI Disclosure

This contribution was prepared with AI assistance and reviewed against the Issue reproduction, current source, and local test results.

### DIFF
--- a/graphiti_core/driver/falkordb/fulltext.py
+++ b/graphiti_core/driver/falkordb/fulltext.py
@@ -71,7 +71,7 @@ def build_falkor_fulltext_query(
     filtered_words = [
         word
         for word in sanitize_falkor_fulltext_query(query).split()
-        if word.lower() not in STOPWORDS
+        if word.lower() not in STOPWORDS and (len(word) > 1 or word.isalnum())
     ]
     if not filtered_words:
         return ''
--- a/graphiti_core/driver/falkordb/fulltext.py
+++ b/graphiti_core/driver/falkordb/fulltext.py
@@ -71,7 +71,7 @@ def build_falkor_fulltext_query(
     filtered_words = [
         word
         for word in sanitize_falkor_fulltext_query(query).split()
-        if word.lower() not in STOPWORDS and (len(word) > 1 or word.isalnum())
+        if word.lower() not in STOPWORDS and any(character.isalnum() for character in word)
     ]
     if not filtered_words:
         return ''
--- a/tests/utils/search/test_search_security.py
+++ b/tests/utils/search/test_search_security.py
@@ -102,6 +102,23 @@ def test_falkordb_fulltext_query_returns_empty_on_punctuation_only():
     assert result == ''
 
 
+@pytest.mark.parametrize(
+    ('query', 'expected'),
+    [
+        (
+            'sessions/<enc-cwd>/<ISO-ts>_<uuid>.jsonl',
+            '(@group_id:"group1") (sessions | enc | cwd | ISO | ts | uuid | jsonl)',
+        ),
+        ('a_b', '(@group_id:"group1") (a_b)'),
+    ],
+)
+def test_falkordb_fulltext_query_drops_standalone_punctuation_tokens(query: str, expected: str):
+    """Standalone punctuation tokens should not produce invalid RediSearch queries."""
+    from graphiti_core.driver.falkordb.operations.search_ops import _build_falkor_fulltext_query
+
+    assert _build_falkor_fulltext_query(query, ['group1']) == expected
+
+
 def test_falkordb_driver_build_fulltext_query_strips_backticks():
     """FalkorDriver.build_fulltext_query should also strip backticks."""
     from graphiti_core.driver.falkordb_driver import FalkorDriver
--- a/tests/utils/search/test_search_security.py
+++ b/tests/utils/search/test_search_security.py
@@ -110,10 +110,11 @@ def test_falkordb_fulltext_query_returns_empty_on_punctuation_only():
             '(@group_id:"group1") (sessions | enc | cwd | ISO | ts | uuid | jsonl)',
         ),
         ('a_b', '(@group_id:"group1") (a_b)'),
+        ('foo __ bar', '(@group_id:"group1") (foo | bar)'),
     ],
 )
 def test_falkordb_fulltext_query_drops_standalone_punctuation_tokens(query: str, expected: str):
-    """Standalone punctuation tokens should not produce invalid RediSearch queries."""
+    """Tokens without alphanumeric characters should not enter RediSearch queries."""
     from graphiti_core.driver.falkordb.operations.search_ops import _build_falkor_fulltext_query
 
     assert _build_falkor_fulltext_query(query, ['group1']) == expected


tstager:
Reporter of #1820 here. I re-ran the issue's reproduction against this PR's head (`f634798`) through the production entry point (`search_utils.node_fulltext_search` → `FalkorDriver.build_fulltext_query`), with released `0.30.1` as the control. Same script, same FalkorDB server, fresh graph for each.

**Environment:** FalkorDB server `falkordb/falkordb:v4.20.4`, `falkordb` client 1.7.1, Python 3.13.

**Released 0.30.1 (control):**

```
built       : (@group_id:"user\_trent\_workspace\_ai\_metrics") (sessions | enc | cwd | ISO | ts | _ | uuid | jsonl)
repro query              -> redis.exceptions.ResponseError: RediSearch: Syntax error at offset 87 near ts
embedded underscore a_b  -> 1 node(s): ['e2']
bare underscore only     -> redis.exceptions.ResponseError: RediSearch: Syntax error at offset 52 near user\_trent\_workspace\_ai\_metrics
```

**This PR (`f634798`):**

```
built       : (@group_id:"user\_trent\_workspace\_ai\_metrics") (sessions | enc | cwd | ISO | ts | uuid | jsonl)
repro query              -> 1 node(s): ['e1']
embedded underscore a_b  -> 1 node(s): ['e2']
bare underscore only     -> 0 node(s): []
```

So the fix resolves the reported failure on the live search path, not just the module-level builder, and it preserves `a_b` as a term (which RediSearch matches fine, as the issue noted). A query that is nothing but `_` now returns empty instead of raising, which is the right shape for the stopword-style filter.

One observation, not a blocker: a multi-character all-punctuation token such as `__` still survives the filter (`len(word) > 1` is true), so `foo __ bar` is emitted as `(foo | __ | bar)`. On v4.20.4 RediSearch accepts that and returns 0 hits rather than erroring, so it is harmless today, but a `word.isalnum() or any(c.isalnum() for c in word)` shape would also cover it if maintainers want the filter to be "drop tokens with no alphanumeric character" rather than "drop single-character punctuation".

Thanks for picking this up.

gkorland:

gkorland:

PR #1846, opened 2026-09-08
fix(search): let the node and community cross-encoders score the summary, not just the name

JohnPark4One:
## Summary
The node and community cross-encoder rerankers passed only `node.name` to `cross_encoder.rank()`. For a query like "who leads the search team?" a candidate named "Kim Minsu" carries none of the query's meaning, so every candidate scored alike and the reranker could not improve on the fused order. Both scopes now score `"name: summary"` (name alone when the summary is empty), matching the edge scope, which already ranks the full fact text.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Testing
- [x] Unit tests added/updated (`tests/utils/search/test_node_cross_encoder_passages.py`, `test_community_cross_encoder_passages.py`): the answer lives in a summary while a lexical decoy shares the query's words in its name; the summary-bearing node must rank first and a node without a summary must be passed as its bare name.
- [x] All existing tests pass (`tests/utils`, `tests/driver`, `tests/cross_encoder` unit modules)

## Breaking Changes

Note: the passage→uuid mapping is still a dict, so two candidates with an identical name *and* summary collapse to one, as two identically named candidates already did.

## Checklist
- [x] Self-review completed

## Related Issues
None

### DIFF
--- a/graphiti_core/search/search.py
+++ b/graphiti_core/search/search.py
@@ -69,6 +69,11 @@ from graphiti_core.tracer import NoOpTracer, Tracer
 logger = logging.getLogger(__name__)
 
 
+def _node_passage(node: EntityNode | CommunityNode) -> str:
+    """Text the cross-encoder scores for a node: name plus summary when there is one."""
+    return f'{node.name}: {node.summary}' if node.summary else node.name
+
+
 def _enum_value(value: Any) -> Any:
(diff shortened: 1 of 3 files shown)


JohnPark4One:
recheck

JohnPark4One:
recheck

PR #1848, opened 2026-09-08
fix(embedder): one Content per input so gemini-embedding-2 returns one vector per text

JohnPark4One:
## Summary
With google-genai 2.x and a `gemini-embedding-2*` model, `GeminiEmbedder.create_batch()` returns **one** embedding for a batch of N texts. The SDK routes models matching `'gemini-embedding-2'` through `t.t_contents()` (`google/genai/models.py`), the generateContent conversion where a `list[str]` is a single multi-part `Content`, so the whole batch is embedded as one input. Callers zip inputs and vectors positionally: `EntityNode.generate_name_embedding` / `EntityEdge` bulk paths raise on the strict zip, and the dedup candidate search silently degrades to a single vector. Reproduced against the live API with google-genai 2.8.0 and 2.22.0 (3 strings → 1 embedding with `gemini-embedding-2`, 3 with `gemini-embedding-001`).

Fix: wrap every text in its own `types.Content` in `create()` and `create_batch()` (semantically identical for every model, one request per input), and refuse a batch whose embedding count does not match the input count, which routes the batch to the existing per-item fallback instead of returning a misaligned list.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Testing
- [x] Unit tests added/updated (`tests/embedder/test_gemini.py`: the request body carries one `Content` per input for both models; a 1-for-N response falls back to per-item calls and returns N vectors)
- [x] All existing tests pass (`tests/embedder/test_gemini.py`, 17 passed)

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
None

### DIFF
--- a/graphiti_core/embedder/gemini.py
+++ b/graphiti_core/embedder/gemini.py
@@ -42,6 +42,17 @@ DEFAULT_EMBEDDING_MODEL = 'text-embedding-001'  # gemini-embedding-001 or text-e
 DEFAULT_BATCH_SIZE = 100
 
 
+def _as_contents(texts: list[str]) -> list[types.Content]:
+    """One Content per input text.
+
+    google-genai (>= 2.x) routes ``gemini-embedding-2`` through the generateContent
+    conversion, where a ``list[str]`` becomes a *single* multi-part Content and the API
+    returns one embedding for the whole batch. Wrapping each text in its own Content keeps
+    one request per input for every model.
+    """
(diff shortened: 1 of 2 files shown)


JohnPark4One:
recheck

JohnPark4One:
recheck

PR #1850, opened 2026-09-08
fix(dedup): let the fuzzy name path see non-Latin scripts

JohnPark4One:
## Summary
`_normalize_name_for_fuzzy` in `dedup_helpers.py` stripped everything outside `[a-z0-9' ]`, so a Korean (or Japanese, Cyrillic, Greek, …) entity name normalized to the empty string, failed the length/entropy gate and never reached the MinHash/LSH fuzzy path; existing candidates had no shingles either. Every non-Latin node that was not an exact normalized match therefore fell through to the embedding candidates and the LLM - including trivially equivalent spacing variants such as `스코프랩스` vs `스코프 랩스`.

Fix: keep Unicode word characters and apostrophes (drop `_` and combining marks), and NFD-decompose the name so Hangul syllables become jamo. A three-syllable Korean name then yields 8–9 characters of shingle material, comparable to a Latin name. Spacing/punctuation variants merge deterministically (Jaccard 1.0), while one-syllable differences (`김민수`/`김민석` 0.62, `서울`/`서울대` 0.60) stay below the 0.9 threshold and still defer to the LLM. Latin behaviour is unchanged (`Alice-Smith!` → `alice smith`); accented Latin now folds (`Ångström` → `angstrom`). The length gate stays at 6.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Documentation/Tests

## Testing
- [x] Unit tests added/updated (`tests/utils/maintenance/test_node_operations.py`: normalization of Korean/accented/Latin names, Korean shingles reach the LSH, spacing variant resolves deterministically, near-miss still goes to the LLM, gate-at-6 pin)
- [x] All existing tests pass (`tests/utils/maintenance`, 122 passed)

## Breaking Changes

## Checklist
- [x] Self-review completed

## Related Issues
None

### DIFF
--- a/graphiti_core/utils/maintenance/dedup_helpers.py
+++ b/graphiti_core/utils/maintenance/dedup_helpers.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import math
 import re
+import unicodedata
 from collections import defaultdict
 from collections.abc import Iterable
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


JohnPark4One:
recheck

JohnPark4One:
recheck

PR #1854, opened 2026-09-08
fix: round-trip saga episode uuids on re-fetch (#1761)

prasmussen15:
## Summary
- Lands contributor PR https://github.com/getzep/graphiti/pull/1761 onto current \`main\` (after #1699).
- Fixes the regression test to use \`+00:00\` instead of \`Z\`, which \`datetime.fromisoformat\` rejects on Python 3.10.

## Test plan
- [ ] unit-tests including \`test_saga_refetch_episode_uuids\`

Made with [Cursor](https://cursor.com)


prasmussen15:
Superseded by the combined landing PR https://github.com/getzep/graphiti/pull/1856 (all four contributor fixes + test corrections in one review).

PR #1871, opened 2026-09-10
fix(gemini): default thinking_budget=0 for gemini-2.5 when unset

### LINKED ISSUE #1868: [Bug] GeminiClient leaves thinking_config=None, so gemini-2.5-* truncates structured output mid-JSON

Zsanz3:
Fixes #1868

## Summary

When `thinking_config` is left unset, Gemini 2.5 models can spend output budget on thinking and truncate structured JSON mid-object.

## Changes

- If the caller did not pass `thinking_config`, default `gemini-2.5-*` to `ThinkingConfig(thinking_budget=0)`.
- Explicit `thinking_config` from the caller still wins (including enabling thinking).
- Constructor docstring updated; small unit coverage for the default.
- Separate from #1869.

## Tests

- `tests/llm_client/test_gemini_client.py` (incl. new default cases)
- Full no-DB unit gate green locally in the agent run

### DIFF
--- a/graphiti_core/llm_client/gemini_client.py
+++ b/graphiti_core/llm_client/gemini_client.py
@@ -69,6 +69,28 @@ GEMINI_MODEL_MAX_TOKENS = {
 DEFAULT_GEMINI_MAX_TOKENS = 8192
 
 
+def _is_gemini_2_5_model(model: str | None) -> bool:
+    """Return True for Gemini 2.5 models, which enable thinking when unset."""
+    return bool(model) and 'gemini-2.5-' in model
+
+
+def _resolve_thinking_config(
+    thinking_config: types.ThinkingConfig | None, model: str | None
+) -> types.ThinkingConfig | None:
+    """Resolve thinking_config, defaulting gemini-2.5-* to thinking off.
+
+    Gemini 2.5 treats ``thinking_config=None`` as thinking ON, and thinking
+    tokens share the output budget with the structured JSON answer. When the
+    caller does not pass a config, disable thinking so extraction is not
+    truncated mid-object. An explicit ``thinking_config`` always wins.
+    """
+    if thinking_config is not None:
+        return thinking_config
+    if _is_gemini_2_5_model(model):
+        return types.ThinkingConfig(thinking_budget=0)
+    return None
+
+
 class GeminiClient(LLMClient):
     """
     GeminiClient is a client class for interacting with Google's Gemini language models.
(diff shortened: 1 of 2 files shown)


mehulp93:
Thanks for contributing to Graphiti. We’ve decided not to merge this PR — you can already turn thinking off without a library change by passing thinking_config when you construct the client:

```
from google.genai import types as genai_types
from graphiti_core.llm_client.config import LLMConfig
from graphiti_core.llm_client.gemini_client import GeminiClient
llm_client = GeminiClient(
    config=LLMConfig(api_key=..., model="gemini-2.5-flash"),
    thinking_config=genai_types.ThinkingConfig(thinking_budget=0),
)
```

PR #1873, opened 2026-09-10
fix(bulk): collapse exact-fact duplicates across episodes in dedupe_edges_bulk

### LINKED ISSUE #1872: [BUG] add_episode_bulk saves duplicate edges for identical facts across episodes in one batch
## Bug Description

`add_episode_bulk` writes two distinct `RELATES_TO` edges when two episodes in the same batch yield the same fact between the same (deduped) source/target nodes. The same two episodes through sequential `add_episode` calls produce one edge.

## Current vs expected behavior

**Current:** `dedupe_edges_bulk` returns both edges and both are saved. Each edge's `episodes` list also gets the other episode's uuid appended (the fast path mutates the candidate in place), so provenance is wrong on both surviving edges.

**Expected:** one edge survives with `episodes` containing every contributing episode (`{ep1, ep2}`), matching the single-episode `add_episode` path.

## Root cause

`resolve_extracted_edge` has an exact-fact fast path (`graphiti_core/utils/maintenance/edge_operations.py:684-695`, added in #929) that returns `(resolved, [], [])` with an **empty** duplicates list:

```python
for edge in related_edges:
    if (
        edge.source_node_uuid == extracted_edge.source_node_uuid
        and edge.target_node_uuid == extracted_edge.target_node_uuid
        and _normalize_string_exact(edge.fact) == normalized_fact
    ):
        resolved = edge
        if episode is not None and episode.uuid not in resolved.episodes:
            resolved.episodes.append(episode.uuid)
        return resolved, [], []
```

The LLM path (`edge_operations.py:845-847`) returns `duplicate_edges = [related_edges[idx] for idx in duplicate_fact_ids]`.

`dedupe_edges_bulk` (`graphiti_core/utils/bulk_utils.py:561-581`) discards the resolved edge and builds `duplicate_pairs` only from the third tuple element:

```python
for i, (_, _, duplicates) in enumerate(bulk_edge_resolutions):
    episode, edge, candidates = dedupe_tuples[i]
    for duplicate in duplicates:
        duplicate_pairs.append((edge.uuid, duplicate.uuid))
compressed_map = compress_uuid_map(duplicate_pairs)
```

So fast-path matches never reach `compress_uuid_map`, and the exact-duplicate case (the one bulk dedupe exists for) is the one case it cannot see.

The existing test `tests/utils/maintenance/test_bulk_utils.py::test_dedupe_edges_bulk_deduplicates_within_episode` (from #955) monkeypatches `resolve_extracted_edge` to return `[related]` as duplicates, the opposite of the real fast path, so it does not catch this.

## Reproduction (no DB, no LLM)

```python
import asyncio
from unittest.mock import MagicMock

from graphiti_core.graphiti_types import GraphitiClients
from graphiti_core.utils import bulk_utils
from graphiti_core.utils.datetime_utils import utc_now


def ep(n):
    return EpisodicNode(name=f'ep-{n}', group_id='g', labels=[], source=EpisodeType.message,
                        content='c', source_description='t', created_at=utc_now(), valid_at=utc_now())


def edge(episode):
    return EntityEdge(name='WORKS_AT', fact='Alice works at Acme', group_id='g',
                      source_node_uuid='alice', target_node_uuid='acme',
                      created_at=utc_now(), episodes=[episode.uuid])


async def main():
    llm = MagicMock()
    llm.generate_response = MagicMock(side_effect=AssertionError('LLM must not be called'))
    clients = GraphitiClients.model_construct(driver=MagicMock(), embedder=MagicMock(),
                                              cross_encoder=MagicMock(), llm_client=llm)

    async def fake_embed(embedder, edges):
        for e in edges:
            e.fact_embedding = [1.0, 0.0]
    bulk_utils.create_entity_edge_embeddings = fake_embed

    e1, e2 = ep(1), ep(2)
    a, b = edge(e1), edge(e2)
    out = await bulk_utils.dedupe_edges_bulk(clients, [[a], [b]], [(e1, []), (e2, [])], [], {}, {})
    uuids = {x.uuid for lst in out.values() for x in lst}
    print('distinct edge uuids after bulk dedupe:', len(uuids), '(expected 1)')
    print('edge a episodes:', a.episodes)
    print('edge b episodes:', b.episodes)

asyncio.run(main())
```

Output on main:

```
distinct edge uuids after bulk dedupe: 2 (expected 1)
edge a episodes: ['<ep1>', '<ep2>']
edge b episodes: ['<ep2>', '<ep1>']
```

The LLM client is never called, so this is purely the fast-path contract mismatch.

## Environment

- graphiti-core 0.30.2, `main` @ 4bd7287
- Backend-independent (Python-side dedupe logic before any DB write)

## Proposed direction

Two changes, both in `dedupe_edges_bulk` (`bulk_utils.py:561-581`); no change to `resolve_extracted_edge`:

1. Treat a returned `resolved` edge whose uuid differs from the input edge as a duplicate pair, in addition to the explicit duplicates list. This is the same convention `resolve_extracted_edges` already uses at `edge_operations.py:520` (`resolved_edge.uuid == extracted_edge.uuid` means "new"; `result[2]` is ignored there).

2. After `compress_uuid_map`, merge `episodes` from each collapsed edge into its canonical edge. The fast path only appends the episode to the *first* matching candidate, so with 3+ episodes the edges end up with uneven provenance (on `main`, 3 identical edges yield `[ep0, ep1, ep2]`, `[ep1, ep0]`, `[ep2]`). `compress_uuid_map` picks the lexicographically smallest uuid as survivor, so without the merge the survivor can carry a partial `episodes` list. The 2-episode case only looks right by accident of the in-place mutation.

```python
for i, (resolved, _, duplicates) in enumerate(bulk_edge_resolutions):
    episode, edge, candidates = dedupe_tuples[i]
    if resolved.uuid != edge.uuid:
        duplicate_pairs.append((edge.uuid, resolved.uuid))
    for duplicate in duplicates:
        duplicate_pairs.append((edge.uuid, duplicate.uuid))

compressed_map = compress_uuid_map(duplicate_pairs)
edge_uuid_map = {edge.uuid: edge for edges in extracted_edges for edge in edges}

for uuid, canonical_uuid in compressed_map.items():
    if uuid == canonical_uuid:
        continue
    canonical = edge_uuid_map[canonical_uuid]
    for ep_uuid in edge_uuid_map[uuid].episodes:
        if ep_uuid not in canonical.episodes:
            canonical.episodes.append(ep_uuid)
```

Roughly 10 lines, one file. No schema or API impact. The union-find is idempotent, so a pair reported by both the uuid check and the duplicates list is harmless. All uuids in `compressed_map` are in-batch, so the `edge_uuid_map` lookups cannot miss.

Not proposed: changing the fast path to return `[resolved]` as duplicates. `dedupe_edges_bulk` is the only consumer of the third tuple element, and `test_resolve_extracted_edge_exact_fact_short_circuit` locks in the current `[]`.

Note (pre-existing, out of scope): which edge survives is decided by uuid ordering, so the survivor's `created_at` and the episode used for `_extract_edge_timestamps` in the second-stage resolve are nondeterministic. Same is true today for the LLM path.

## Tests

New unit tests in `tests/utils/maintenance/test_bulk_utils.py` using the real `resolve_extracted_edge` (fast path) with an LLM mock that raises if called:

- Identical fact across 2 episodes: one surviving uuid across `edges_by_episode`, both episode lists reference the same object, survivor `episodes` equals `{ep1, ep2}` as a set.
- Identical fact across 3 episodes: same assertions with 3. This is the case the uuid-pair fix alone does not cover; compare `episodes` as a set, not by order.
- Same endpoints, different facts (no word overlap, orthogonal embeddings): 2 uuids survive, LLM not called.
- Strengthen `test_dedupe_edges_bulk_deduplicates_within_episode` to assert a single surviving uuid; today it only counts comparisons.
- `test_resolve_extracted_edge_exact_fact_short_circuit` unchanged.

## Related issues / PRs

- #929 introduced the exact-fact fast path
- #955 added `dedupe_edges_bulk` within-episode comparisons and the mocked test that masks this
- #1771 (dedup write-path review)

L4XB:
## Summary

Fixes #1872.

`add_episode_bulk` wrote two distinct `RELATES_TO` edges when two episodes in the same batch yielded the same fact between the same nodes, while sequential `add_episode` calls produce one. `resolve_extracted_edge` reports a duplicate in two ways: the LLM path lists the matched candidates in the third tuple element, but the exact-fact fast path (#929) returns the matched candidate itself as `resolved` with an **empty** duplicates list. `dedupe_edges_bulk` only read the duplicates list, so fast-path matches never reached `compress_uuid_map`, and each edge additionally got the other episode appended in place.

## Change

`graphiti_core/utils/bulk_utils.py`, `dedupe_edges_bulk` only (no change to `resolve_extracted_edge`):

- a `resolved` edge whose uuid differs from the input edge is recorded as a duplicate pair, in addition to the explicit duplicates list, mirroring the `resolved_edge.uuid == extracted_edge.uuid` convention in `resolve_extracted_edges`; the union-find is idempotent, so a pair reported both ways is harmless;
- after `compress_uuid_map`, every collapsed edge's `episodes` are folded into its canonical edge, so the survivor carries the full provenance (the fast path appends the episode to the first matching candidate only, which left three identical edges with `[ep0, ep1, ep2]`, `[ep1, ep0]`, `[ep2]`).

## Tests

`tests/utils/maintenance/test_bulk_utils.py`, driven through the **real** `resolve_extracted_edge` (the existing within-episode test mocks it to return the duplicate in the third element, i.e. the opposite of the fast path, which is why it did not catch this); the LLM client is an `AsyncMock` that fails if called:

- two episodes with an identical fact → one surviving uuid, both episodes reference the same edge object, and it records both episode uuids (fails on `main`: 2 uuids);
- three episodes → the survivor carries all three episode uuids (fails on `main`).

`pytest tests/utils/maintenance/test_bulk_utils.py` → 16 passed; `ruff format` / `ruff check` clean; pyright reports the same two pre-existing environment import errors on the module as on `main`.

### DIFF
--- a/graphiti_core/utils/bulk_utils.py
+++ b/graphiti_core/utils/bulk_utils.py
@@ -558,8 +558,17 @@ async def dedupe_edges_bulk(
 
     # For now we won't track edge invalidation
     duplicate_pairs: list[tuple[str, str]] = []
-    for i, (_, _, duplicates) in enumerate(bulk_edge_resolutions):
+    for i, (resolved, _, duplicates) in enumerate(bulk_edge_resolutions):
         episode, edge, candidates = dedupe_tuples[i]
+        # resolve_extracted_edge signals a duplicate in two ways: the LLM path
+        # lists the matched candidates as `duplicates`, while the exact-fact
+        # fast path returns the matched candidate itself as `resolved` with an
+        # EMPTY duplicates list. resolve_extracted_edges treats
+        # `resolved.uuid != edge.uuid` as "duplicate of an existing edge";
+        # mirror that here, or two episodes in one batch that yield the same
+        # fact between the same nodes both survive as distinct edges.
+        if resolved.uuid != edge.uuid:
+            duplicate_pairs.append((edge.uuid, resolved.uuid))
         for duplicate in duplicates:
             duplicate_pairs.append((edge.uuid, duplicate.uuid))
 
@@ -569,6 +578,19 @@ async def dedupe_edges_bulk(
     edge_uuid_map: dict[str, EntityEdge] = {
         edge.uuid: edge for edges in extracted_edges for edge in edges
     }
+
+    # The surviving edge of a collapsed group must carry every contributing
+    # episode. The fast path only appends the episode to the first matching
+    # candidate, so with three or more identical edges the provenance is
+    # spread unevenly across the group and the survivor (the lexicographically
+    # smallest uuid) can hold a partial list.
+    for uuid, canonical_uuid in compressed_map.items():
+        if uuid == canonical_uuid:
+            continue
+        canonical = edge_uuid_map[canonical_uuid]
+        for episode_uuid in edge_uuid_map[uuid].episodes:
+            if episode_uuid not in canonical.episodes:
+                canonical.episodes.append(episode_uuid)
 
     edges_by_episode: dict[str, list[EntityEdge]] = {}
     for i, edges in enumerate(extracted_edges):
--- a/tests/utils/maintenance/test_bulk_utils.py
+++ b/tests/utils/maintenance/test_bulk_utils.py
@@ -563,3 +563,71 @@ async def test_extract_nodes_and_edges_bulk_custom_instructions_multiple_episode
 
     for call in extract_edges_calls:
         assert call['custom_extraction_instructions'] == custom_instructions
+
+
+def _identical_edge(episode: EpisodicNode, fact: str = 'Alice works at Acme') -> EntityEdge:
+    return EntityEdge(
+        name='WORKS_AT',
+        fact=fact,
+        group_id='group',
+        source_node_uuid='alice-uuid',
+        target_node_uuid='acme-uuid',
+        created_at=utc_now(),
+        episodes=[episode.uuid],
+    )
+
+
+async def _run_bulk_dedupe_through_real_resolver(monkeypatch, edges_per_episode):
+    """Drive dedupe_edges_bulk through the REAL resolve_extracted_edge so the
+    exact-fact fast path (which returns the matched candidate as `resolved`
+    with an EMPTY duplicates list) is exercised, not a mock of it (#1872)."""
+    clients = _make_clients()
+    clients.llm_client.generate_response = AsyncMock(
+        side_effect=AssertionError('exact-fact duplicates must not reach the LLM')
+    )
+
+    async def mock_create_embeddings(embedder, edges):
+        for edge in edges:
+            edge.fact_embedding = [1.0, 0.0]
+
+    monkeypatch.setattr(bulk_utils, 'create_entity_edge_embeddings', mock_create_embeddings)
+
+    episodes = [_make_episode(str(i)) for i in range(len(edges_per_episode))]
+    extracted = [
+        [_identical_edge(episode) for _ in range(count)]
+        for episode, count in zip(episodes, edges_per_episode, strict=True)
+    ]
+    episode_tuples = [(episode, []) for episode in episodes]
+
+    result = await bulk_utils.dedupe_edges_bulk(clients, extracted, episode_tuples, [], {}, {})
+    return episodes, extracted, result
+
+
+@pytest.mark.asyncio
+async def test_dedupe_edges_bulk_collapses_identical_facts_across_episodes(monkeypatch):
+    """Two episodes in one batch yielding the same fact between the same nodes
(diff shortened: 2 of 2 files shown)


L4XB:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: lukas.buck@e-mail.de

L4XB:
recheck

Issue #1592, opened 2026-06-17
FalkorDB: edge_fulltext_search Node-By-Label-Scan per hit (redundant endpoint labels defeat the edge-uuid index) → add_episode times out at scale


mbs-vhs:
### Summary

On the **FalkorDB** backend, `edge_fulltext_search` emits:

```cypher
CALL db.idx.fulltext.queryRelationships('RELATES_TO', $query) YIELD relationship AS rel, score
MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)
...
```

FalkorDB's planner does not recognize that pinning the edge by its indexed `uuid` already determines the endpoints, so the `(n:Entity)` / `(m:Entity)` labels trigger a **`Node By Label Scan`** over *all* `:Entity` nodes — **once per fulltext hit**. With a broad fulltext query that matches many edges this is **O(hits × entities)**, and `add_episode` fails with `Query timed out` as the graph grows.

### Evidence (`GRAPH.EXPLAIN`, ~3,300 `RELATES_TO` edges / ~3,100 `Entity` nodes)

With the endpoint labels:
```
Limit → Sort → Project → Filter
        → Edge By Index Scan | [e:RELATES_TO]
            → Node By Label Scan | (n:Entity)     ← scans every Entity, per hit
```

Without the endpoint labels (`MATCH (n)-[e:RELATES_TO {uuid: rel.uuid}]->(m)`):
```
Limit → Sort → Project → Edge By Index Scan | [e:RELATES_TO]
```

Measured on the live graph: **25s+ timeout with labels → 0.007s without.** Dropping the labels takes `add_episode` from 1/20 episodes persisting to 20/20.

### Why it's FalkorDB-specific

graphiti is Neo4j-first and the FalkorDB path reuses the identical Cypher — the labelled `MATCH` is byte-identical in `graphiti_core/search/search_utils.py`, `graphiti_core/driver/falkordb/operations/search_ops.py`, and `graphiti_core/driver/neo4j/operations/search_ops.py`. Neo4j's planner uses the relationship index and skips the redundant label check; FalkorDB's does not. The endpoint labels are semantically redundant — a `RELATES_TO` edge always connects two `Entity` nodes.

[...]

PR #1877, opened 2026-09-10
fix(falkordb): read endpoints off the yielded relationship in search_ops edge fulltext search

### LINKED ISSUE #1875: [BUG] [PERF] FalkorDB: edge_fulltext_search in search_ops.py re-matches endpoints, causing a full :Entity scan per hit (latent)
## Problem

`FalkorSearchOperations.edge_fulltext_search` in `graphiti_core/driver/falkordb/operations/search_ops.py` still contains the pre-#1711 endpoint re-match:

```cypher
```

#1711 fixed this pattern in `graphiti_core/search/search_utils.py` only. FalkorDB plans the `MATCH` as a full `:Entity` label scan for every row the fulltext index yields, so the copy in `search_ops.py` retains the O(hits × entities) cost that #1711 removed.

This is latent, not currently live: `FalkorDriver` sets `_search_ops` / `search_ops` but leaves the legacy `search_interface` unset, and `edge_fulltext_search` in `search_utils.py` only delegates when `driver.search_interface` is truthy. So the fixed path is the one that executes today.

The concern is the migration. `spec/driver-operations-redesign.md` annotates `search_utils.py` with "Will gradually migrate to use driver.search_ops". When that lands for edge fulltext search, #1506, #1272 and #1592 all regress in full, silently — the query still returns correct results, so only latency changes.

## Impact when the migration completes

Measured on `falkordb/falkordb:v4.18.10` with the exact v0.30.2 query strings and index definitions from `get_fulltext_indices(FALKORDB)`, comparing the two shapes on identical graphs:

| Graph | Fulltext hits | `search_ops.py` shape | `search_utils.py` shape (v0.30.2) |
| --- | --- | --- | --- |
| 2,500 entities / 4,500 edges | 4,500 (broad) | timeout, >30 s | 5.1 ms |
| 5,000 / 20,000 | 200 (narrow) | 4,524 ms | 0.5 ms |
| 5,000 / 20,000 | 20,000 (broad) | timeout, >30 s | 22.9 ms |
| 20,000 / 100,000 | 1,000 (narrow) | timeout, >30 s | 1.9 ms |

`GRAPH.PROFILE` shows the cause: with the re-match, 200 fulltext hits over 5,000 entities scan 1,000,000 records; the v0.30.2 shape has no scan at all. Note that a *narrow* query is affected too — cost scales with entity count, not just hit count.

## Suggested fix

Apply the #1711 change to `search_ops.py`, so both copies agree:

```cypher
WITH rel AS e, score, startNode(rel) AS n, endNode(rel) AS m
WHERE n:Entity AND m:Entity
```

## Related

- #1711 — the fix, applied to `search_utils.py` only
- #1506, #1272, #1592 — the issues #1711 resolved, all of which return if `search_ops.py` becomes live
- #1826 — the same class of fix for the *episode* path in `search_ops.py`. Worth noting that the live `episode_fulltext_search` in `search_utils.py` still uses `MATCH (e:Episodic) WHERE e.uuid = episode.uuid` with no FalkorDB branch, so the episode path has the O(hits × nodes) shape in both copies.

brentkearney:
Applies the #1711 fix to the second copy of the query, in `graphiti_core/driver/falkordb/operations/search_ops.py`.

Fixes #1875.

## Problem

`FalkorSearchOperations.edge_fulltext_search` resolves a matched relationship's endpoints by re-matching them:

```cypher
```

FalkorDB plans that as a full `:Entity` label scan for every row the fulltext index yields, so the cost is O(hits × entities). #1711 removed this from `graphiti_core/search/search_utils.py` and left the driver-level copy unchanged.

The copy is unreachable today: `FalkorDriver` sets `search_ops` but leaves the legacy `search_interface` unset, and `search_utils.edge_fulltext_search` only delegates when `driver.search_interface` is truthy. It becomes the executing path once search migrates onto `driver.search_ops`, which `spec/driver-operations-redesign.md` marks as planned. At that point #1506, #1272 and #1592 all return, silently — results stay correct, only latency changes.

## Change

Read the endpoints off the relationship the index already returned, matching what `search_utils.py` now does:

```cypher
WITH rel AS e, score, startNode(rel) AS n, endNode(rel) AS m
WHERE n:Entity AND m:Entity
WITH e, score, n, m
```

The label predicate is retained, so the result set stays identical to the pattern's — the old `MATCH` required both endpoints to be `:Entity`.

The episode path in this file has the same shape and is not touched here; #1826 covers it.

## Verification

**Result-set equivalence on a real FalkorDB** (`falkordb/falkordb:v4.18.10`), on a graph built with graphiti's own index definition and seeded with 400 `:Entity`-to-`:Entity` edges plus 20 `:Other`-to-`:Other` edges carrying the same searchable text, to exercise the label predicate:

| Fulltext query | Old shape | New shape | Identical | Non-`:Entity` edges returned |
| --- | --- | --- | --- | --- |
| `alpha` | 134 | 134 | yes | 0 / 0 |
| `beta` | 133 | 133 | yes | 0 / 0 |
| `alpha\|beta\|gamma` | 400 | 400 | yes | 0 / 0 |

Both shapes exclude all 20 non-`:Entity` edges, so the predicate still does its job.

**Cost, measured separately on the same image** with the two shapes on identical graphs:

| Graph | Fulltext hits | Old shape | New shape |
| --- | --- | --- | --- |
| 2,500 entities / 4,500 edges | 4,500 | timeout, >30 s | 5.1 ms |
| 5,000 / 20,000 | 200 | 4,524 ms | 0.5 ms |
| 5,000 / 20,000 | 20,000 | timeout, >30 s | 22.9 ms |
| 20,000 / 100,000 | 1,000 | timeout, >30 s | 1.9 ms |

Note the narrow-query rows: the old shape scales with entity count, not just hit count, so even a 200-hit query is affected.

**Tests.** `tests/utils/search/test_falkordb_search_ops_query_shape.py` asserts the emitted Cypher uses `startNode`/`endNode`, never re-matches by uuid, and keeps the label predicate, plus that the group filter and limit still apply. It uses the recording-executor approach of the neighbouring `test_edge_bfs_query_shape.py`, so it needs no database. Both tests fail on the unpatched file and pass with the change, and they sit under `tests/utils/search/` so the existing unit-tests job collects them.

The no-database unit gate is green: 382 passed, 11 skipped. `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/graphiti_core/driver/falkordb/operations/search_ops.py
+++ b/graphiti_core/driver/falkordb/operations/search_ops.py
@@ -238,13 +238,20 @@ class FalkorSearchOperations(SearchOperations):
         if filter_queries:
             filter_query = ' WHERE ' + (' AND '.join(filter_queries))
 
+        # FalkorDB plans a re-match of the endpoints by uuid as a full :Entity label
+        # scan for EVERY row the fulltext index yields, so the cost becomes
+        # O(hits x entities). The relationship the index already returned knows its
+        # own endpoints, so read them directly. The label predicate keeps the result
+        # set identical to the pattern's.
         cypher = (
             get_relationships_query(
                 'edge_name_and_fact', limit=limit, provider=GraphProvider.FALKORDB
             )
             + """
             YIELD relationship AS rel, score
-            MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)
+            WITH rel AS e, score, startNode(rel) AS n, endNode(rel) AS m
+            WHERE n:Entity AND m:Entity
+            WITH e, score, n, m
             """
             + filter_query
             + """
--- a/tests/utils/search/test_falkordb_search_ops_query_shape.py
+++ b/tests/utils/search/test_falkordb_search_ops_query_shape.py
@@ -1,0 +1,82 @@
+"""Query-shape regression tests for FalkorSearchOperations (no live database).
+
+``edge_fulltext_search`` must read a matched relationship's endpoints off the
+relationship the fulltext index already returned (``startNode``/``endNode``)
+instead of re-matching them by uuid
+(``MATCH (n:Entity)-[e:RELATES_TO {uuid: rel.uuid}]->(m:Entity)``). FalkorDB
+plans the re-match as a full ``:Entity`` label scan for every row the index
+yields, so the cost becomes O(hits x entities).
+
+The generic path in ``graphiti_core/search/search_utils.py`` was fixed for this;
+the driver-level copy here was not, and it becomes the executing path once
+search migrates onto ``driver.search_ops``.
+
+Mirrors the recording-driver approach of
+tests/utils/search/test_edge_bfs_query_shape.py: a recording executor captures
+the emitted Cypher, so no database connection is required.
+"""
+
+from typing import Any
+
+import pytest
+
+from graphiti_core.driver.falkordb.operations.search_ops import FalkorSearchOperations
+from graphiti_core.search.search_filters import SearchFilters
+
+
+class RecordingExecutor:
+    """Captures the Cypher and params a search operation emits, returning no rows."""
+
+    def __init__(self):
+        self.cypher_query = ''
+        self.params: dict[str, Any] = {}
+
+    async def execute_query(self, cypher_query_: str, **kwargs: Any):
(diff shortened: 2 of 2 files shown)


gkorland:

Issue #1842, opened 2026-09-07
FalkorDB driver: multi-group_id reads (get_episodes/search_nodes) silently query the wrong graph


jdbway:
## Summary

With the FalkorDB driver, `get_episodes`/`search_nodes` (and other read paths that call `EpisodicNode.get_by_group_ids` / `client.search_()` directly) silently query the *wrong graph* and return empty results, depending on unrelated prior write activity on the same `Graphiti` client instance. This is a correctness bug, not a performance issue — reads can silently return no results even though the requested `group_id` has real data.

## Environment

- `graphiti-core` 0.29.2 (installed via `mcp_server`'s pinned deps)
- Driver: FalkorDB
- Reproduced via the bundled MCP server (`mcp_server/src/graphiti_mcp_server.py`), but the root cause is in `graphiti_core` itself, not the MCP server.

## Root cause

FalkorDB is multi-tenant: each `group_id` lives in its own separate FalkorDB graph (confirmed via `GRAPH.LIST` — e.g. `my-group-a`, `my-group-b` show up as distinct top-level graphs, not as a `group_id` property inside one shared graph).

To route a write to the correct graph, `add_episode` **mutates the shared driver in place**:

```python
# graphiti_core/graphiti.py (two call sites, e.g. ~line 1081 and ~1309)
self.driver = self.driver.clone(database=group_id)
```

`FalkorDriver.clone()` returns a new driver instance bound to a different `_database`, but because it's assigned back onto `self.driver`, the **shared `Graphiti` client's driver stays permanently rebound** to whichever `group_id` was most recently written, until the next write changes it again.

Meanwhile, `FalkorDriver.execute_query()` always queries whatever `self._database` currently is — it does not accept or honor a per-call database/group override:

```python
# graphiti_core/driver/falkordb_driver.py
async def execute_query(self, cypher_query_, **kwargs: Any):
    graph = self._get_graph(self._database)   # always the driver's *current* bound database
    ...
```

And `EpisodicNode.get_by_group_ids` (graphiti_core/nodes.py) falls through to this same `execute_query` path, because `FalkorDriver` never sets `graph_operations_interface` (it's `None` by default on the base `GraphDriver`, confirmed by grepping the FalkorDB driver module — no override exists):

```python
async def get_by_group_ids(cls, driver, group_ids, limit=None, uuid_cursor=None):
    if driver.graph_operations_interface:   # always None for FalkorDriver
        ...
    records, _, _ = await driver.execute_query(
[...]

PR #1878, opened 2026-09-11
fix: scope single group_id reads to the same DB as writes

### LINKED ISSUE #1876: [BUG] group_id selects a database for writes but not for reads: add_episode persists to the cloned database, search/get_episodes query the configured one
## Problem

`add_episode` resolves a per-group database, but read paths do not. Writes and reads therefore target different databases whenever `group_id` differs from the driver's configured database, and reads come back empty with no error.

`_resolve_request_scope` (added in #1699 to fix #1676) clones the driver per call:

```python
if group_id == self.driver._database:
    return group_id, self.driver, self.clients

clients = self.clients.model_copy(update={'driver': driver})
return group_id, driver, clients
```

Only `add_episode` and `add_episode_bulk` call it. The read paths do not:

- `Graphiti.search()` passes `self.clients` and an optional caller-supplied `driver`; it never derives one from `group_ids`.
- The MCP server's `get_episodes` calls `EpisodicNode.get_by_group_ids(client.driver, ...)` on the base driver.

So an episode written under `group_id="X"` lands in database `X`, while a read for `group_ids=["X"]` runs against the configured database and finds nothing.

## Reproduction

Using the MCP server against FalkorDB with a deliberate mismatch:

- `FALKORDB_DATABASE=smoke_db`
- `GRAPHITI_GROUP_ID=smoke_0302`

Steps:

1. `add_memory` three episodes. Each reports success, and the server logs `Successfully processed episode ... for group smoke_0302`.
2. `GRAPH.LIST` shows two graphs, `smoke_db` and `smoke_0302`. All three `Episodic` nodes are in `smoke_0302`, confirming the write clone worked.
3. `get_episodes` returns `No episodes found` for `group_ids="smoke_0302"`, for `["smoke_0302"]`, and for the default.

Expected: the episodes just written are returned. Actual: nothing, with no error.

## Why this is easy to miss

The single-database deployment is unaffected: when `group_id` equals the configured database, `_resolve_request_scope` returns the shared driver unchanged, so writes and reads agree. The failure needs `group_id != database`, which is exactly the multi-tenant shape #1676 was about.

It also fails silently. Ingestion reports success, the data is genuinely persisted, and the read returns an empty result rather than an error — so it presents as "memory not working" rather than as a misconfiguration.

## Suggested fix

Resolve the request scope on read paths too, so a `group_id` selects the same database for reads as for writes. Two shapes seem plausible:

1. Have `search()` / `search_()` and the episode accessors derive the driver from `group_ids` the way `add_episode` does, when a single `group_id` is given.
2. Or reject the mismatch at construction time, if targeting a database other than the one a `group_id` maps to is meant to be unsupported.

The choice depends on whether `group_id` is intended to *name* the database or merely to partition within it. Right now the write path treats it as the former and the read path as the latter, and that disagreement is the bug.

## Related

- #1676 — concurrent multi-`group_id` isolation, the original report
- #1699 — the fix, which covered the write path

Zsanz3:
Fixes #1876

## Summary

`add_episode` already resolves a request-scoped driver for a single `group_id` (clone into the tenant DB). Read paths (`search` / `search_` / episode accessors / MCP `get_episodes`) still queried the configured database, so writes succeeded and reads came back empty with no error.

## Changes

- Add `_resolve_read_scope` in `graphiti_core/graphiti.py`: single `group_id` reuses `_resolve_request_scope`; an already-passed `driver` (e.g. `add_episode` → `retrieve_episodes`) is not cloned again.
- MCP `get_episodes` uses the same scoped driver for a single `group_id`.
- Extend `tests/test_request_scope_concurrency.py` (clone, explicit driver, concurrency, no shared-driver mutation).
- Does not mutate the shared driver in place (avoids the #1842 pitfall). Multi-`group_id` still goes through existing `handle_multiple_group_ids`.

## Tests

- `tests/test_request_scope_concurrency.py` + `tests/test_handle_multiple_group_ids.py`


gkorland:

PR #1898, opened 2026-09-18
fix(search): preserve per-edge node-distance scores

### LINKED ISSUE #1897: node_distance edge search misaligns scores when facts share a source
### What happened
With `EdgeReranker.node_distance`, `edge_search()` can return more edges than scores when multiple facts have the same source node. A consumer pairing the two lists can assign a neighbor's score to a center-node fact and omit later facts.

### Reproduction
On main (`de8eb5b896c05ed1b5b329d4cb52015446d65e21`), supply four candidates: two edges whose source is the center node, one whose source is adjacent to the center, and one whose source is unconnected. Use the node-distance reranker with the default minimum score of zero and a limit of four.

- Returned edges: both center-node edges, then the adjacent-node edge, then the unconnected edge.
- Actual scores: `[10.0, 1.0, 0.0]`.
- Expected scores: `[10.0, 10.0, 1.0, 0.0]`.

With a limit of two, the lengths happen to match, but the second center-node edge still receives `1.0` instead of `10.0`.

### Cause and proposed fix
`edge_search()` expands each reranked source UUID into its edge UUIDs, while retaining the per-source score list. Expand the score at the same time so every edge inherits its source's distance score. This needs no change to the distance calculation, filtering, or final limit.

I have a focused fix and three deterministic regression cases using the real distance reranker, with retrieval and database calls mocked. No database or model API is needed to reproduce this. This is separate from #1609, which concerns sorting edge-score pairs in the `episode_mentions` branch.

Regression test: [test_edge_node_distance_scores.py](https://github.com/Shy7777/graphiti/blob/01f10c085ee3350b7f4db24c66b32e0807054c6a/tests/utils/search/test_edge_node_distance_scores.py).

Shy7777:
## Summary
When several facts share a source node, the node-distance reranker returns one score per source but expands that source into multiple edges. This leaves the returned edges and scores out of alignment. Copy each source's score to all of its edges before applying the result limit.

For example, two facts from the center node and one from its neighbor now return scores `[10, 10, 1]` instead of `[10, 1]`.

## Type of Change
- [x] Bug fix

## Testing
- [ ] All existing tests pass

The three regression cases fail on main and pass with this change. They cover multiple facts from one source, result limiting, and filtering out an unconnected source. They exercise the real node-distance reranker with mocked retrieval and database boundaries.

The CI unit-test command for tests without external services passes: **385 passed, 11 skipped**. Ruff and the changed module's Pyright check pass. Full-project Pyright reports seven errors from optional backends not installed locally (sentence-transformers, Kuzu, and Neptune dependencies); database integration tests were not run.

## Breaking Changes
None. Edge order, distance scores, thresholds, and limits keep their existing semantics; each returned edge now has its matching score.

## Checklist
- [x] Self-review completed

## Related Issues

Fixes #1897.


Shy7777:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 58774225+Shy7777@users.noreply.github.com

PR #1900, opened 2026-09-19
fix cross-encoder edge rerank dropping distinct edges that share fact text

### LINKED ISSUE #1816: [Bug] Edge cross-encoder reranking drops distinct edges with identical fact text
## Summary

edge_search() silently drops distinct EntityEdge results when two candidate edges have identical fact text and the cross-encoder reranker is enabled.

The cross-encoder path builds a one-to-one mapping:

    fact_to_uuid_map = {edge.fact: edge.uuid for edge in rrf_edges}

Later edges overwrite earlier edges with the same text. Only one passage is sent to the cross-encoder, and only one UUID can be reconstructed.

## Why this is a bug

Two edges can legitimately have the same fact while representing different graph records:

- different source or target nodes;
- different group_id values;
- different temporal validity;
- different episode provenance;
- separate relationships that verbalize to the same sentence.

Search identity is UUID-based elsewhere. This path unexpectedly changes identity to edge.fact and silently loses records after candidate retrieval.

## Affected code

graphiti_core/search/search.py, in the EdgeReranker.cross_encoder branch of edge_search():

    rrf_edges = [edge_uuid_map[uuid] for uuid in rrf_result_uuids][: 2 * limit]
    fact_to_uuid_map = {edge.fact: edge.uuid for edge in rrf_edges}
    reranked_facts = await cross_encoder.rank(query, list(fact_to_uuid_map.keys()))
    reranked_uuids = [
        fact_to_uuid_map[fact]
        for fact, score in reranked_facts
        if score >= reranker_min_score
    ]

## Deterministic reproduction

No database, network, LLM, or GPU is required. Mock edge_fulltext_search to return two edges with different UUIDs but the same fact, then use a fake cross-encoder:

    first.uuid = 'edge-1'
    first.fact = 'Alice knows Bob'
    second.uuid = 'edge-2'
    second.fact = 'Alice knows Bob'

    async def fake_fulltext(*args, **kwargs):
        return [first, second]

    class FakeCrossEncoder:
        async def rank(self, query, passages):
            return [(passage, 1.0) for passage in passages]

Calling edge_search(..., reranker=EdgeReranker.cross_encoder, limit=2) currently produces:

    returned_edge_uuids = ['edge-2']
    returned_scores = [1.0]

Expected both edge UUIDs and two aligned scores. This reproduces on current main at 8b61fce9f003cc3a05e246f6201f8b782dfe6546.

## Expected behavior

Cross-encoder reranking should preserve candidate identity by UUID. Distinct edges must remain independently addressable even when their passage text is identical, and returned edges and scores must stay aligned.

## Suggested fix direction

Avoid a one-to-one fact-to-UUID dictionary. Preserve each candidate's identity explicitly, for example with stable candidate indices or an ordered one-to-many mapping for duplicate text.

The fix should preserve these invariants:

- every shortlisted edge remains independently representable;
- duplicate fact text does not collapse UUIDs;
- scores map deterministically back to edges;
- len(edges) == len(edge_scores);
- existing RRF shortlist, threshold, and limit behavior remains unchanged.

## Regression test

Extend tests/utils/search/test_edge_cross_encoder_rrf_shortlist.py with a pure unit test that:

1. returns two distinct edges sharing the same fact from a mocked search method;
2. records passages received by a fake cross-encoder;
3. asserts both edge UUIDs survive reranking;
4. asserts both scores are present and aligned.

## Duplicate-work check

Before filing, I searched open and closed issues and pull requests for fact_to_uuid_map, cross-encoder duplicate facts, identical fact text with different edge UUIDs, and edge loss during reranking. No issue or pull request covering this root cause was found. Related issues such as #1328 and #1728 concern invalid embeddings and invalidation scope, not cross-encoder identity collisions.

## Implementation ownership

I plan to implement the focused fix and deterministic regression test for this issue. I have not opened a branch or pull request yet so the issue can establish the problem and avoid duplicate work first.

feiiiiii5:
## Summary

`edge_search` lost results when two different edges happened to carry identical `fact` text. The cross-encoder rerank built its candidate lookup as `{edge.fact: edge.uuid}`, so a later edge with the same text overwrote the earlier one's uuid and that edge never appeared in the returned list — silently, in a public search API.

Fixes #1816.

## Type of Change
- [x] Bug fix

## Root cause

`graphiti_core/search/search.py`, `EdgeReranker.cross_encoder` branch:

```python
fact_to_uuid_map = {edge.fact: edge.uuid for edge in rrf_edges}
...
reranked_uuids = [fact_to_uuid_map[fact] for fact, score in reranked_facts if ...]
```

`fact` is free text, not a key. Two edges (`edge-a`, `edge-b`) with the same fact collapse into one dict entry, and only the last uuid can ever come back. Every other identity in this function is a `uuid`, and the neighbouring `EdgeReranker.node_distance` branch already handles exactly this shape with a `defaultdict(list)` and an `extend`.

Measured with three distinct edges sharing one fact: `de8eb5b` returns 1 edge, this branch returns all 3.

## Change

Rank one entry per **unique** fact (unchanged cost: `rank()` still receives unique passages), then broadcast each score back to every uuid carrying that fact:

```python
fact_to_uuids_map = defaultdict(list)
for edge in rrf_edges:
    fact_to_uuids_map[edge.fact].append(edge.uuid)
...
for fact, score in reranked_facts:
    if score < reranker_min_score:
        continue
    for uuid in fact_to_uuids_map[fact]:
        reranked_uuids.append(uuid)
        edge_scores.append(score)
```

`reranked_uuids` and `edge_scores` are appended together because the return at the end of the function slices **both** by `limit`; keeping them index-aligned is what preserves the existing `min_score` and `limit` semantics. Per-edge behaviour for unique facts is unchanged, including order.

## Testing

New file `tests/utils/search/test_edge_cross_encoder_duplicate_facts.py` (4 tests, in the style of `test_edge_cross_encoder_rrf_shortlist.py`: monkeypatched `edge_fulltext_search` / `edge_similarity_search` and a recording fake cross-encoder — no database, no API key, no network). It covers two and three edges sharing one fact, score/index alignment under `reranker_min_score`, that `limit` still caps the expanded list, and that `rank()` is called once per unique fact rather than once per edge.

Environment: Python 3.11.15, `uv sync --extra dev`, `DISABLE_NEO4J=1 DISABLE_FALKORDB=1 DISABLE_KUZU=1 DISABLE_NEPTUNE=1 GRAPHITI_TELEMETRY_ENABLED=false`. The exact command from `.github/workflows/unit_tests.yml` (same `-m "not integration"` and `--ignore` list), run against `de8eb5b` sources with this test file present and then with the fix:

```
base (de8eb5b search.py):  4 failed, 382 passed, 11 skipped
head (this change):      386 passed, 11 skipped
```

The 4 failures are exactly the new tests, and 382 + 4 = 386, so nothing else moved. Narrower runs, same environment: `tests/utils/search` 33 passed at base → 37 passed with the change.

Lint and types:
- `ruff check` and `ruff format --check` (ruff 0.14.11, repo config): clean on both changed files.
- `pyright ./graphiti_core` (pyright 1.1.408): **19 errors before, 19 errors after**, none in `search.py` — the count is unchanged by this diff. In this offline env several come from optional extras not installed here (e.g. `llm_client/gemini_client.py:31 "genai" is unknown import symbol`), so I am reporting it as "no new errors", not as "pyright is green".

## Breaking Changes

Returned lists can now be longer for result sets that contained duplicate fact text — that is the lost data reappearing. No signature, config field, or public type changes.

## Checklist
- [x] Code follows project style guidelines (`ruff check` / `ruff format --check` pass; `pyright` error count unchanged)
- [x] Self-review completed
- [x] Documentation updated where necessary (no doc change needed)

## Related Issues

Closes #1816

## Notes for the maintainer

1. **This PR is opened as a draft on purpose.** The Zep CLA is required and signing it is the account owner's own action — `signatures/version1/cla.json` currently lists 176 contributors and this account is not among them. Once that signature is posted, this can go straight to ready for review; nothing else is outstanding on my side.
2. @linhongyu510 wrote on #1816 that a local patch was prepared and is being held pending direction. If that work is still coming, I am happy to close this in favour of it — the shape here (ordered one-to-many fact → uuids, score broadcast) is deliberately the same as the `node_distance` branch, so either PR leaves the file in the same state.
3. Out of scope here, noticed while writing the tests: `edge_search` reads its threshold from the `reranker_min_score` **parameter**, while `EdgeSearchConfig.reranker_min_score` is a separate field that this function never consults. Whether those two should be unified is a design call, so this diff leaves both alone and the new tests pass the value the same way production callers do.

**Disclosure:** this contribution was produced by an automated agent working for the account owner, including the commands and quoted counts above. It is presented for human review, not as human-reviewed work.


feiiiiii5:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: fei76161@gmail.com

feiiiiii5:
recheck

PR #1614, opened 2026-06-25
fix(mcp_server): add missing docker-compose.yml and update default da…

joan8627:
## Summary

**Fix `mcp_server` Docker setup: add missing `docker-compose.yml` and align README with Neo4j as default**

First-time contributor here - I ran into this while setting up Graphiti for the first time and figured I'd fix it before moving on to more substantive work.

**What was broken**

The `mcp_server/README.md` repeatedly instructs users to run `docker compose up` from the `mcp_server/` directory, but no `docker-compose.yml` existed there -  only inside `mcp_server/docker/`. This means the command silently fails for anyone following the README from scratch.

Additionally, the README listed FalkorDB as the default database, while the parent directory's `docker-compose.yml` and the actual `docker/docker-compose-neo4j.yml` both treat Neo4j as the default. This inconsistency is confusing for newcomers trying to understand which setup to follow.

**What I changed**

- Added `mcp_server/docker-compose.yml` that includes `docker/docker-compose-neo4j.yml` directly, making `docker compose up` from `mcp_server/` work as the README describes
- Updated `mcp_server/README.md` to reflect Neo4j as the default database throughout
- Updated `mcp_server/docker/README.md` to match

**Why this approach**

The alternative was to update the README to remove the `docker compose up` shortcut and tell users to always use `-f docker/docker-compose-neo4j.yml` explicitly. I kept the shortcut instead because the README is clearly written around it - it's mentioned multiple times and the whole "run from `mcp_server/`" instruction exists specifically to support it. The missing file looked like an accidental omission rather than an intentional removal.

I plan to focus on more urgent issues going forward - this was mostly a warmup while getting my own environment working. Hopefully it saves the next newcomer a few minutes of head-scratching.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [x] Documentation/Tests

## Objective
Keep the mcp_server documentation consistent with the actual repo structure, and save newcomers the confusion of following instructions that don't quite work out of the box.

## Testing
- [N.A.] Unit tests added/updated
- [N.A.] Integration tests added/updated
- [x] All existing tests pass

## Checklist
- [x] Code follows project style guidelines (`make lint` passes)
- [x] Self-review completed
- [x] Documentation updated where necessary
- [x] No secrets or sensitive information committed

PR #1901, opened 2026-09-19
fix(search): consume the yielded node in episode_fulltext_search

### LINKED ISSUE #1819: [BUG] FalkorDB: episode_fulltext_search re-matches yielded nodes by unindexed uuid equality — O(hits × episodes), hangs search

TankLin23:
## Summary

Fixes #1819 on the path that actually executes.

`episode_fulltext_search` in `graphiti_core/search/search_utils.py` re-matched every fulltext hit against the whole `Episodic` label:

```
CALL db.idx.fulltext.queryNodes('Episodic', $query) YIELD node AS episode, score
MATCH (e:Episodic)
WHERE e.uuid = episode.uuid
```

`e.uuid = episode.uuid` compares a stored property to a per-row value, so no index can serve it: each hit scans all `Episodic` nodes (O(hits × episodes)), and every row drags a large `content` string. The reporter measured combined-recipe searches that stopped finishing inside 30s on a 2,163-episode graph, while pinning a core.

The procedure already returns the node we want, and `episode_content` is only ever created over `Episodic` (`CREATE FULLTEXT INDEX episode_content FOR (e:Episodic)` on Neo4j/FalkorDB, `CREATE_FTS_INDEX('Episodic', 'episode_content', ...)` on Kuzu), so the re-MATCH buys nothing.

## Relation to #1826

#1826 touches only `graphiti_core/driver/falkordb/operations/search_ops.py`, which is not reached at runtime: `driver.search_interface` is never assigned, so `FalkorSearchOperations` is bypassed and the copy in `search_utils.py` is what runs. The repo already documents this — `tests/utils/search/test_edge_bfs_query_shape.py` describes itself as completing #1500 "on the generic path in `graphiti_core/search/search_utils.py`, which is what executes today for FalkorDB and Neo4j (`driver.search_interface` is never assigned, so the driver-level operations modules patched by #1500 are not reached at runtime)". Reviewers made the same point on #1471 and #1500.

So landing #1826 alone leaves episode fulltext search at O(hits × episodes) for its users. This PR applies the same fix where it runs; #1826 remains useful for the driver module and can land on its own.

## Changes

`episode_fulltext_search` now consumes the yielded node, exactly like `node_fulltext_search` and `community_fulltext_search` right above and below it in the same module:

```
YIELD node AS e, score          # WITH node AS e, score on Kuzu
WITH e, score
```

The group filter used to hang off the removed re-MATCH as an `AND`, so it becomes a standalone `WHERE`, matching `community_fulltext_search`. Ordering and `LIMIT` are untouched, and results are unchanged wherever the re-MATCH already resolved.

The diff also gives this function the provider-aware clause its two siblings have: it hardcoded `YIELD`, but Kuzu's `QUERY_FTS_INDEX` is a table function and takes `WITH node AS <alias>, score`. Easy to split out if you would rather land it separately.

## Tests

New DB-free regression test, `tests/utils/search/test_episode_fulltext_query_shape.py`, using the recording-driver approach from `tests/utils/search/test_edge_bfs_query_shape.py`. It pins that the re-MATCH is gone, the group filter survives, no `WHERE` is left dangling when `group_ids` is None, Kuzu gets `WITH` rather than `YIELD`, FalkorDB still targets the `Episodic` index, and ordering plus limit are preserved. All five cases fail on `main`.

```
uv run pytest tests/utils/search/test_episode_fulltext_query_shape.py -q
5 passed

# no-DB gate from .github/workflows/unit_tests.yml
387 passed, 11 skipped          # 382 on main + the 5 new
```

`uv run ruff check` and `uv run pyright graphiti_core` are both clean.

One caveat worth stating: I have no live FalkorDB to point this at, so verification is at the query-shape level. The removed predicate is the one the reporter profiled, and the replacement is character-for-character the shape `node_fulltext_search` already runs in production.


TankLin23:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 1046831682@qq.com

TankLin23:
I have read the CLA Document and I hereby sign the CLA behalf on myself, e-mail: 1046831682@qq.com

Re-posting the signature line because the CLA check finished before my signature was
recorded and is still showing red — a comment starting with this line is what re-triggers it.

@pevans the `Tests`, `Pyright Type Check` and `CodeQL Advanced` runs on this PR are still
waiting on "Approve and run" (first-time contributor), so they have not executed yet;
`ruff` and both Socket jobs are green. The change itself is limited to dropping the
redundant re-`MATCH` in `episode_fulltext_search` (plus a query-shape regression test) —
happy to adjust anything before you spend time on it.

TankLin23:
Correction to the note above: the red CLAAssistant check isn't a signing problem. The CLA workflow did run on my comment and the signature is recorded, but the job errors out when it tries to refresh the check:

> Error occurred when re-running the workflow: HttpError: Resource not accessible by integration

(run: https://github.com/getzep/graphiti/actions/runs/35604566754)

So only a manual re-run on the maintainer side can clear it as far as I can tell - re-signing from my side doesn't. The same red check is sitting on #1614, so it may be worth a look at the CLA app permissions at some point.

TankLin23:
- CLA signed; CLAAssistant, ruff and both Socket Security jobs are green on the current head.
- Tests, Pyright Type Check and CodeQL sit at action_required — they need a maintainer to
  approve the first run from a new contributor. I ran the equivalents locally: 387 passed,
  11 skipped on the no-DB gate, ruff check clean, pyright graphiti_core 0 errors.
- Scope is one function in graphiti/search/search_utils.py (+7/-4), plus a DB-free test that fails on main.

